PR #19, opened 2025-09-25
Add initial support for devcontainer

### LINKED ISSUE #18: Add support for devcontainer
Running agent-shell's ACP against a devcontainer isn't currently possible.

Thanks to @fritzgrabo for investigating and finding what's needed. Pasting screenshot he's shared with me. With this in mind, we can come up with a solution. Will need changes in `agent-shell` and [acp.el](https://github.com/xenodium/acp.el/blob/main/acp.el).

Image

fritzgrabo:
Introduces a new customization option `agent-shell-map-path-function` to allow for mapping paths in ACP requests.

By default, maps paths from a devcontainer to the local filesystem, or vice versa, when there is a `.devcontainer/devcontainer.json` file in your project root / working directory, which specifies a `workspaceFolder`. If such a file is detected, we assume devcontainers are used and paths should be mapped.

If you prefer not to use a project's devcontainer setup at all, just set `agent-shell-map-path-function` to `nil` locally (for example, in dir-local variables).

Note that the new customization option is not not, in fact, devcontainer specific -- just its default setting is. You could set `agent-shell-map-path-function` to a function that maps paths between your local filesystem and, say, a remote SSH box via TRAMP even.

Note that this PR does not deal with starting the agent in a devcontainer -- you'll still need to use a handcrafted `client-maker` for that. For example, mine looks like this:

```lisp
(defun agent-shell-start-claude-code-agent ()
  "Start an interactive Claude Code agent shell."
  (interactive)
  (agent-shell--start
   ;; ...
   :client-maker (lambda ()
                   (acp-make-client
                    :command "devcontainer"
                    :command-params '("exec" "--workspace-folder" "." "claude-code-acp")))))

```

Resolves #18

### DIFF
--- a/README.org
+++ b/README.org
@@ -182,6 +182,31 @@ Start an agent shell session:
 - =M-x agent-shell-google-start-gemini= - Start a Gemini agent session
 - =M-x agent-shell-goose-start-agent= - Start a Goose agent session
 
+** Running agents in Devcontainers / Docker containers
+
+=agent-shell= provides rudimentary support for running agents in containers.
+
+Adapt the command that starts the agent so it is executed inside the container; for example:
+
+#+begin_src emacs-lisp
+(setq agent-shell-anthropic-claude-command '("devcontainer" "exec" "--workspace-folder" "." "claude-code-acp"))
+#+end_src
+
+Note that any =:environment-variables= you may have passed to =acp-make-client= will not apply to the agent process running inside the container.
+It's expected to inject environment variables by means of your devcontainer configuration / Dockerfile.
+
+Next, set an =agent-shell-path-resolver-function= that resolves container paths in the local working directory, and vice versa.
+Agent shell provides the =agent-shell--resolve-devcontainer-path= function for use with devcontainers:
+
+#+begin_src emacs-lisp
+(setq agent-shell-path-resolver-function #'agent-shell--resolve-devcontainer-path)
+#+end_src
+
+Note that this allows the agent to access files on your local file-system.
+While care has been taken to restrict access to files in the local working directory, it's probably possible for a malicious agent to circumvent this restriction.
+
+All of the above settings can be applied on a per-project basis using [[https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html][directory-local variables]].
+
 ** Keybindings
 
 - =C-c C-c= - Interrupt current agent operation
--- a/README.org
+++ b/README.org
@@ -205,7 +205,7 @@ Agent shell provides the =agent-shell--resolve-devcontainer-path= function for u
 Note that this allows the agent to access files on your local file-system.
 While care has been taken to restrict access to files in the local working directory, it's probably possible for a malicious agent to circumvent this restriction.
 
-Optional: to prevent access of your local file-system altogether and to have the agent read/modify files inside the container directly, in addition to setting the resolver function, disable the "read/write text file" client capabilities:
+Optional: to prevent the agent running inside the container to access your local file-system altogether and to have it read/modify files inside the container directly, in addition to setting the resolver function, disable the "read/write text file" client capabilities:
 
 #+begin_src emacs-lisp
 (setq agent-shell-text-file-capabilities nil)
--- a/README.org
+++ b/README.org
@@ -182,7 +182,7 @@ Start an agent shell session:
 - =M-x agent-shell-google-start-gemini= - Start a Gemini agent session
 - =M-x agent-shell-goose-start-agent= - Start a Goose agent session
 
-** Running agents in Devcontainers / Docker containers
+** Running agents in Devcontainers / Docker containers (Experimental)
 
 =agent-shell= provides rudimentary support for running agents in containers.
 
--- a/README.org
+++ b/README.org
@@ -205,6 +205,12 @@ Agent shell provides the =agent-shell--resolve-devcontainer-path= function for u
 Note that this allows the agent to access files on your local file-system.
 While care has been taken to restrict access to files in the local working directory, it's probably possible for a malicious agent to circumvent this restriction.
 
+Optional: to prevent access of your local file-system altogether and to have the agent read/modify files inside the container directly, in addition to setting the resolver function, disable the "read/write text file" client capabilities:
+
+#+begin_src emacs-lisp
+(setq agent-shell-text-file-capabilities nil)
+#+end_src
+
 All of the above settings can be applied on a per-project basis using [[https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html][directory-local variables]].
 
 ** Keybindings
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -575,11 +575,11 @@ For example:
     (if (string-prefix-p cwd path)
         (string-replace cwd devcontainer-path path)
       (if agent-shell-text-file-capabilities
-          (if (string-prefix-p devcontainer-path path)
-              (let ((local-path (expand-file-name (string-replace devcontainer-path cwd path))))
-                (or
-                 (and (file-in-directory-p local-path cwd) local-path)
-                 (error "Resolves to path outside of working directory: %s" path)))
+          (if-let* ((is-dev-container (string-prefix-p devcontainer-path path))
+                    (local-path (expand-file-name (string-replace devcontainer-path cwd path))))
+              (or
+               (and (file-in-directory-p local-path cwd) local-path)
+               (error "Resolves to path outside of working directory: %s" path))
             (error "Unexpected path outside of workspace folder: %s" path))
         (error "Refuse to resolve to local filesystem with text file capabilities disabled: %s" path)))))
 
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -70,7 +70,9 @@ returns the resolved path. Set to nil to disable mapping."
   :group 'agent-shell)
 
 (defcustom agent-shell-text-file-capabilities t
-  "Whether agents are initialized with read/write text file capabilities."
+  "Whether agents are initialized with read/write text file capabilities.
+
+See `acp-make-initialize-request' for details."
   :type 'boolean
   :group 'agent-shell)
 
@@ -549,7 +551,9 @@ LINE defaults to 1, LIMIT defaults to nil (read to end)."
   (funcall (or agent-shell-path-resolver-function #'identity) path))
 
 (defun agent-shell--get-devcontainer-workspace-path (cwd)
-  "Return devcontainer workspaceFolder for CWD; signal error if none found."
+  "Return devcontainer workspaceFolder for CWD; signal error if none found.
+
+See https://containers.dev for more information on devcontainers."
   (let ((devcontainer-config-file-name (expand-file-name ".devcontainer/devcontainer.json" cwd)))
     (condition-case _err
         (or
@@ -560,7 +564,12 @@ LINE defaults to 1, LIMIT defaults to nil (read to end)."
       (json-string-format (error "No valid JSON: %s" devcontainer-config-file-name)))))
 
 (defun agent-shell--resolve-devcontainer-path (path)
-  "Resolve PATH from a devcontainer in the local filesystem, and vice versa."
+  "Resolve PATH from a devcontainer in the local filesystem, and vice versa.
+
+For example:
+
+- /workspace/README.md => /home/xenodium/projects/kitchen-sink/README.md
+- /home/xenodium/projects/kitchen-sink/README.md => /workspace/README.md"
   (let* ((cwd (agent-shell-cwd))
          (devcontainer-path (agent-shell--get-devcontainer-workspace-path cwd)))
     (if (string-prefix-p cwd path)
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -69,6 +69,11 @@ returns the resolved path. Set to nil to disable mapping."
   :type 'function
   :group 'agent-shell)
 
+(defcustom agent-shell-text-file-capabilities t
+  "Whether agents are initialized with read/write text file capabilities."
+  :type 'boolean
+  :group 'agent-shell)
+
 (cl-defun agent-shell--make-state (&key buffer client-maker needs-authentication authenticate-request-maker)
   "Construct shell agent state with BUFFER.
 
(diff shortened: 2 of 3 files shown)


xenodium:
Thanks for this! Generally LGTM. One thing I'm wondering is if we should try getting this patch into acp.el. The benefit being that any other Emacs pacakge built on top of acp.el would get the feature for free. WDYT?

fritzgrabo:
> getting this patch into acp.el

Good call, I agree.

This will be more involved I believe, so would love to run my thoughts by you before I do anything to make sure I'm not missing something obvious that would allow for a simpler approach.

I'd give the ACP client alist two new attributes:

* `local-cwd`: can be set by passing a keyword arg of the same name to `acp-make-client`, defaults to a copy of `(agent-shell-cwd)` otherwise (would copy or move that function from `agent-shell` to `acp.el`).

* `remote-cwd`: can be set by passing a keyword arg to `acp-make-client`. If none given, would funcall a new `defcustom acp-determine-remote-cwd-function`, which defaults to a function that looks for a dev-container setup in `local-cwd` and reads from there. Folks can override this in dir-locals if they want to ignore dev-container files for a project, or do something else entirely.

I'd also rename the client's `{notification, request, error}-handlers` attributes to `incoming-{...}-handlers`, and pre-populate `incoming-request-handlers` with a handler that swaps out the local/remote cwd in well-known request types (for now, `fs/read` and `fs/write` requests), if the client's `remote-cwd` is non-nil.

~~Note that this requires a change of the handler signatures: besides the request, they'd also receive the client alist so they can access the local and remote cwd settings. This means changes in `agent-shell` as well, unfortunately, which is a bummer.~~ Note true: can bake the paths into a dynamically created handler function from within `acp-make-client`, where I still have access to them.

I'd add a new `outgoing-request-handlers` attribute to the client alist and pre-populate it with a handler that swaps out cwds if needed (same as above). For now, that's the case in `session/new` only.

What do you think? Is there a simpler way to do this? Thanks!

xenodium:
Sorry for the delay. Was a little tied up after the release with troubleshooting and fixing bugs. Btw, using Claude subscriptions should now be possible now.

> This will be more involved I believe, so would love to run my thoughts by you before I do anything to make sure I'm not missing something obvious that would allow for a simpler approach.

Thank you. I really appreciate pre-emptive discussion.

> with a handler that swaps out the local/remote cwd in well-known request types (for now, fs/read and fs/write requests), if the client's remote-cwd is non-nil

This is making me think that your original approach (to implement in agent-shell) is simpler and we should reconsider it. Also why I'm grateful of further discussion (enables changing our minds with new info without feeling too attached to an implementation). At present, acp.el is a thin transport layer, agnostic of request/response/notification-specific logic. Moving the container logic to acp changes that and isn't something I had considered. I was hoping only the acp-make-session-new-request would apply the required logic on the way out, but I missed the logic on the way in...

There may be another option in acp by allowing clients to inject "decorators" which are invoked on the way out (ie. send-request) as well as on the way in (ie. before invoking on-success). Having said all of that, I'm leaning towards your initial approach (implement in agent-shell) and see how we get on with it before getting more creative.

Going back to the original approach, some minor details... I would consider renaming the feature overall as "resolving paths".

- agent-shell-map-path-function -> agent-shell-path-resolver-function
- agent-shell-maybe-map-devcontainer-path -> agent-shell--resolve-devcontainer-path (-- for private).
- agent-shell-map-path -> agent-shell--resolve-path
- alist-get -> map-elt (no real reason other than trying to stay consistent and use map.el - I may have missed some myself).

Heads-up, https://github.com/xenodium/acp.el/commit/1ebce93920f478737f7f1516a7248e4a425aec48 is likely going to give you trouble. We can revert this change. I've made the same change in agent-shell. We just need to be mindful we don't regress https://github.com/xenodium/agent-shell/issues/29

Hope all that made sense. Happy to add more details if needed.

xenodium:
Thanks for the changes. Looks great! I did a smoke test without container as I'm not set up to do containers yet. If you're happy with the changes as they are, I'll merge.

ps. There were a couple of byte-compile and M-x checkdoc warnings. Happy to follow-up and fix after merging.

fritzgrabo:
I had a few minutes to address your feedback, but would still like to polish this MR up a little in case you decide to roll with devcontainer support in `agent-shell` (vs. `acp.el` or other places, see below):

- make the error handling in `agent-shell--get-devcontainer-workspace-path` more verbose. Failing silently and defaulting to running the client on the local host (outside of the devcontainer) might not be what the user expects or wants. After all, they might have chosen the devcontainer setup for security reasons.

- either find a way to run the server inside the devcontainer automatically when needed, or leave that to the user and make the use of devcontainer non-automatic / opt-in (i.e. defaulting `agent-shell-path-resolver-function` to `nil`). I think it's all-or-nothing, since using devcontainers needs both a custom way of launching the server and the path resolving. Automagically enabling one but not the other seems unexpected from a user's POV.

- add a note to README.md that explains how to use the feature (especially important when it's opt-in).

Re. decorators:

> There may be another option in acp by allowing clients to inject "decorators" which are invoked on the way out (ie. send-request) as well as on the way in (ie. before invoking on-success).

This idea is appealing, it feels like it has the potential for a proper separation of concerns.

I identified these blocks of functionality that could be moved around between `acp.el` and clients to support  devcontainers and similar setups (vanilla Docker, TRAMP, etc.):

(a) detect the situation
(b) allow for opting-out per project (dir-local variables)
(c) know which requests need patching where
(d) provide for detection/configuration of path resolving logic (ie. parse devcontainer config, defcustoms, etc.) 
(e) handle server launch config

(a) and (b) can be skipped if devcontainer logic is opt-in and needs to be manually enabled by the user.

Injectable decorators for request handling would help with (c) only imho, and still require the client package to implement the actual request patching logic. `acp.el` would just help with applying it in the right spots. (a), (b), (d), (e) would have to stay in the client package.

We could try and move all of (a)-(e) into `acp.el`. Doable, but the package would no longer be just about the protocol then. Also, it'd have to track all individual types of requests that ACP defines (still under development), no matter if client packages use actually that request type or not.

Moreover, transparently reworking server launch configs to use devcontainer seems brittle: you'd generally not use ENV vars when launching the servers, but declare them in the devcontainer config files, for example. 

tl;dr Given both packages are still being actively developed, I suggest making devcontainer usage opt-in, provide `agent-shell-path-resolver-function` (default to `nil`) and `agent-shell--resolve-devcontainer-path` (fail loud if no devcontainer setup found), skip any logic around launch configuration for now, and provide an example setup for the interested user in README.md to adapt to their own needs.

Sorry for the wall of text; please let me know what you think! Thanks!

xenodium:
> tl;dr Given both packages are still being actively developed, I suggest making devcontainer usage opt-in, provide agent-shell-path-resolver-function (default to nil) and agent-shell--resolve-devcontainer-path (fail loud if no devcontainer setup found), skip any logic around launch configuration for now, and provide an example setup for the interested user in README.md to adapt to their own needs.

This sounds reasonable to me. Opt-in for now. This gives us the chance to use it for a period of time and potentially learn of other edge cases we haven't thought of. We can start by including the feature in agent-shell and later decide whether we're happy to keep here or move to acp. We're still learning lots about ACP and dynamics around it, so we may as well stick with current approach and rethink later if needed.

fritzgrabo:
All done; ready for review 🙌

agent-shell.el
```diff
@@ -61,6 +61,19 @@
...
+(defcustom agent-shell-path-resolver-function nil
+  "Function for resolving remote paths on the local file-system, and vice versa.
+
+Expects a function that takes the path as its single argument, and
+returns the resolved path. Set to nil to disable mapping."
+  :type 'function
+  :group 'agent-shell)
+
+(defcustom agent-shell-text-file-capabilities t
+  "Whether agents are initialized with read/write text file capabilities."
```
xenodium: Maybe add:

See `acp-make-initialize-request' for details.

agent-shell.el
```diff
@@ -531,6 +544,36 @@ LINE defaults to 1, LIMIT defaults to nil (read to end)."
                            :code -32603
                            :message (error-message-string err))))))))
 
+(defun agent-shell--resolve-path (path)
+  "Resolve PATH using `agent-shell-path-resolver-function'."
+  (funcall (or agent-shell-path-resolver-function #'identity) path))
+
+(defun agent-shell--get-devcontainer-workspace-path (cwd)
+  "Return devcontainer workspaceFolder for CWD; signal error if none found."
```
xenodium: Maybe a URL/link to what a devcontainer is would be handy for newcomers (including myself).

agent-shell.el
```diff
@@ -531,6 +544,36 @@ LINE defaults to 1, LIMIT defaults to nil (read to end)."
...
+    (condition-case _err
+        (or
+         (map-elt (json-read-file devcontainer-config-file-name) 'workspaceFolder)
+         (error "No workspace folder defined in %s" devcontainer-config-file-name))
+      (file-missing (error "Not found: %s" devcontainer-config-file-name))
+      (permission-denied (error "Not readable: %s" devcontainer-config-file-name))
+      (json-string-format (error "No valid JSON: %s" devcontainer-config-file-name)))))
+
+(defun agent-shell--resolve-devcontainer-path (path)
+  "Resolve PATH from a devcontainer in the local filesystem, and vice versa."
```
xenodium: Would be nice to include an example usage so anyone looking to maintain this function gets a mental/visual picture of external behaviour.

"For example:

  (agent-shell--resolve-devcontainer-path "something")
    => "something else" "

agent-shell.el
```diff
@@ -531,6 +544,36 @@ LINE defaults to 1, LIMIT defaults to nil (read to end)."
...
+  (let* ((cwd (agent-shell-cwd))
+         (devcontainer-path (agent-shell--get-devcontainer-workspace-path cwd)))
+    (if (string-prefix-p cwd path)
+        (string-replace cwd devcontainer-path path)
+      (if agent-shell-text-file-capabilities
+          (if (string-prefix-p devcontainer-path path)
```
xenodium: Optionally, if you want to flatten a little, I think we can do something like:

```emacs-lisp
(if-let* ((is-dev-container (string-prefix-p devcontainer-path path))
          (local-path (expand-file-name (string-replace devcontainer-path cwd path))))
    (or
     (and (file-in-directory-p local-path cwd) local-path)
     (error "Resolves to path outside of working directory: %s" path))
  (error "Unexpected path outside of workspace folder: %s" path))
```

README.org
```diff
@@ -182,6 +182,37 @@ Start an agent shell session:
 - =M-x agent-shell-google-start-gemini= - Start a Gemini agent session
 - =M-x agent-shell-goose-start-agent= - Start a Goose agent session
 
+** Running agents in Devcontainers / Docker containers
```
xenodium: We may want to mark this as "(Experimental)"

README.org
```diff
@@ -182,6 +182,37 @@ Start an agent shell session:
...
+
+=agent-shell= provides rudimentary support for running agents in containers.
+
+Adapt the command that starts the agent so it is executed inside the container; for example:
+
+#+begin_src emacs-lisp
+(setq agent-shell-anthropic-claude-command '("devcontainer" "exec" "--workspace-folder" "." "claude-code-acp"))
+#+end_src
+
+Note that any =:environment-variables= you may have passed to =acp-make-client= will not apply to the agent process running inside the container.
```
xenodium, line 195: FYI: Tangentially related, you can now pass environment variables to claude code https://github.com/xenodium/agent-shell/commit/5d90cba6b314150e3b57ac7a3f1aaab1836d1f9c#diff-aa539d101a2aa2b9a6a9e61e52e2a0db39588c73b16857518c234afee7843af6

I've not documented this on README yet. Waiting for some confirmation: https://github.com/xenodium/agent-shell/issues/41

README.org
```diff
@@ -182,6 +182,37 @@ Start an agent shell session:
...
+(setq agent-shell-path-resolver-function #'agent-shell--resolve-devcontainer-path)
+#+end_src
+
+Note that this allows the agent to access files on your local file-system.
+While care has been taken to restrict access to files in the local working directory, it's probably possible for a malicious agent to circumvent this restriction.
+
+Optional: to prevent access of your local file-system altogether and to have the agent read/modify files inside the container directly, in addition to setting the resolver function, disable the "read/write text file" client capabilities:
+
+#+begin_src emacs-lisp
+(setq agent-shell-text-file-capabilities nil)
```
xenodium, line 211: Heads-up, `agent-shell-text-file-capabilities nil` won't restrict file access when running agent outside of the sandbox. Instead of asking Emacs to make file changes, it will go and access the files directly from the filesystem.

xenodium:
Thanks a lot for the changes. Looks great. Left some minor comments.

Optionally (I don't want to shift the goal post), I started adding tests (better late than never), in case you want to add some too https://github.com/xenodium/agent-shell/commit/5d90cba6b314150e3b57ac7a3f1aaab1836d1f9c#diff-9d9aea74742c1dbce474188857dd9d7444aa76861b0f5c2f1268413c1a1e1449

agent-shell.el
fritzgrabo: Done 👍

agent-shell.el
fritzgrabo: Done 👍

agent-shell.el
fritzgrabo: Added an example for both directions (container to local, and local to container) 👍

agent-shell.el
fritzgrabo: Ah, that's a good call -- thanks. Done 👍

README.org
fritzgrabo: Done 👍

README.org
fritzgrabo: I believe these env vars would still apply to the process environment for the locally-run `devcontainer` executable only. Specifically, they would not be visible to the `claude-code-acp` process run inside the container.

README.org
fritzgrabo: That's a good call-out, thanks. I tried to be more verbose about the effect of setting `agent-shell-text-file-capabilities` to `nil` in the `README.md` now.

fritzgrabo:
> Optionally (I don't want to shift the goal post), I started adding tests (better late than never), in case you want to add some too

I gave it a try. Let me know what you think! Thanks!

README.org
xenodium: Thank you. LGTM.

README.org
xenodium: Yup yup. I went on a tangent.

xenodium (approved):

xenodium:
Thanks a lot for iterating on this, your thoughtfulness, and adding tests! Much appreciated. Merging!

PR #33, opened 2025-09-27
Fix customization group for agent-shell-anthropic-authentication

rynffoll:
Change the customization group for the Anthropic API key setting from `acp` to `agent-shell` in `agent-shell-anthropic.el`


xenodium:
Thanks!

PR #47, opened 2025-10-02
Only call agent-shell-goose-key once

### LINKED ISSUE #46: agent-shell-goose-key
This function should only be called in `agent-shell-goose-make-client` and `agent-shell-goose-start-agent` doesn't need it because it calls `agent-shell-goose-make-client`.

hexmode:
fixes #46


xenodium:
Thanks for the PR!

PR #48, opened 2025-10-03
Enable context in diffs

djr7C4:
This turned out to be as simple as changing -U0 to -U3


xenodium (approved):

xenodium:
>  This turned out to be as simple as changing -U0 to -U3

Sweet :)

Merged

PR #73, opened 2025-10-17
Embedded context support

bontaq:
I've been having a lot of fun using this at work, wonderful quality library.  The traffic viewer is sick.

This adds support for embedding content & content refs based on https://agentclientprotocol.com/protocol/schema#contentblock 

I've been testing it with Claude Code

In regular operation (embedding supported)
image

And when it falls back:
image

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1561,7 +1561,7 @@ Returns list of alists with :start, :end, and :path for each mention."
               (text . ,(substring-no-properties prompt pos)))
             content-blocks))
 
-    content-blocks))
+    (nreverse content-blocks)))
 
 (cl-defun agent-shell--collect-attached-files (content-blocks)
   "Collect attached resource uris from CONTENT-BLOCKS."
@@ -1589,7 +1589,10 @@ Returns list of alists with :start, :end, and :path for each mention."
 
 (cl-defun agent-shell--send-command (&key prompt shell)
   "Send PROMPT to agent using SHELL."
-  (let* ((content-blocks (agent-shell--build-content-blocks prompt))
+  (let* ((content-blocks (condition-case nil
+                             (agent-shell--build-content-blocks prompt)
+                           (error `[((type . "text")
+                                     (text . ,(substring-no-properties prompt)))])))
          (attached-files (agent-shell--collect-attached-files content-blocks)))
 
     (when attached-files (agent-shell--display-attached-files attached-files))
@@ -1598,8 +1601,7 @@ Returns list of alists with :start, :end, and :path for each mention."
      :client (map-elt agent-shell--state :client)
      :request (acp-make-session-prompt-request
                :session-id (map-nested-elt agent-shell--state '(:session :id))
-               :prompt `[((type . "text")
-                          (text . ,(substring-no-properties prompt)))])
+               :prompt content-blocks)
      :buffer (current-buffer)
      :on-success (lambda (response)
                    ;; Tool call details are no longer needed after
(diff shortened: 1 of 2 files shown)


agent-shell.el
```diff
@@ -1439,29 +1452,117 @@ Must provide ON-SESSION-INIT (lambda ())."
    :on-request (lambda (request)
                  (agent-shell--on-request :state state :request request))))
 
+(defun agent-shell--parse-file-mentions (prompt)
+  "Parse @ file mentions from PROMPT string.
+Returns list of plists with :start, :end, and :path for each mention."
```
xenodium: Mind using alists + map.el please? Can still use same keys (ie. :start...). Sorry, I know it's a matter of taste, but the proposed combo is now the prevalent approach in the package. Staying consistent will make it a bit easier for me to maintain longer term.

agent-shell.el
```diff
@@ -1439,29 +1452,117 @@ Must provide ON-SESSION-INIT (lambda ())."
...
-                 (funcall (agent-shell--make-error-handler :state agent-shell--state :shell shell)
-                          error raw-message))))
+  (let* ((supports-embedded-context (map-elt agent-shell--state :agent-supports-embedded-context))
+         (cwd (agent-shell-cwd))
+         (mentions (agent-shell--parse-file-mentions prompt))
+         (content-blocks '())
+         (attached-files '())
+         (pos 0))
+
+    (dolist (mention mentions)
```
xenodium, line 1508: Can we move the looping out of `agent-shell--send-command` please? I'm thinking we can have two functions: one to return content blocks and another for attached files (makes them testable on their own). More importantly, `agent-shell--send-command` merely vconcats their returned values.

agent-shell.el
```diff
@@ -98,6 +98,13 @@ See `display-buffer' for the format of display actions."
   :type 'boolean
   :group 'agent-shell)
 
+(defcustom agent-shell-embed-file-size-limit 102400
```
xenodium, line 102: My agent usage only really started once I was able to use ACP, so not super familiar with what's common in CLI agents themselves... Is this common behavior (picking according to threshold)?

agent-shell.el
```diff
@@ -1439,29 +1452,117 @@ Must provide ON-SESSION-INIT (lambda ())."
...
+                (if (and supports-embedded-context
+                         file-size
+                         (< file-size agent-shell-embed-file-size-limit))
+                    (let ((content (with-temp-buffer
+                                    (insert-file-contents resolved-path)
+                                    (buffer-string))))
+                      (push `((type . "resource")
+                             (resource . ((uri . ,(concat "file://" resolved-path))
+                                         (text . ,content)
+                                         (mimeType . "text/plain"))))
```
xenodium: Can we add a TODO here please? Maybe something like:

TODO: Determine mime type instead of hardcoding.

agent-shell.el
```diff
@@ -1439,29 +1452,117 @@ Must provide ON-SESSION-INIT (lambda ())."
...
+                            content-blocks))
+                  ;; File too large or agent doesn't support embeddedContext - use resource link
+                  ;; so agent can fetch via fs/read_text_file RPC if needed
+                  (push `((type . "resource_link")
+                         (uri . ,(concat "file://" resolved-path))
+                         (name . ,path)
+                         (mimeType . "text/plain")
```
xenodium: Same here. TODO for mime type.

agent-shell.el
bontaq: I'm not too sure either and didn't see guidance in the docs, since it can choose to still read the file I was thinking of it as a sort of safety

agent-shell.el
bontaq: Moved it out, but I'm not sure it's split up as you were thinking 🤔 

Edit: I think I can do at least a little better and move the resource block building into its own fn too

xenodium: Thank you. I'll accept it as is. If you reckon you can improve it, I'd totally accept the follow-up too.

xenodium:
Thanks for the PR. Merged now!

Issue #72, opened 2025-10-16
Add support for Zed's Codex ACP


kvaneesh:
I am setting up the codex-acp (https://github.com/zed-industries/codex-acp ) using

```
(setq agent-shell-openai-authentication
      (agent-shell-openai-make-authentication
      :api-key "abc"))
```

But codex> interface gives the error
   ╭─

     ⚠ Error (-32000) ⚠

     Authentication required

details shows 
```
((:object (jsonrpc . "2.0") (id . 2)
	  (error (code . -32000) (message . "Authentication required")))
 (:json
  . "{\"jsonrpc\":\"2.0\",\"id\":2,\"error\":{\"code\":-32000,\"message\":\"Authentication required\"}}"))
```

xenodium:
> I am setting up the codex-acp (https://github.com/zed-industries/codex-acp ) using

I haven't tried Zed's Codex ACP.

So far, I've integrated this https://github.com/xenodium/agent-shell?tab=readme-ov-file#codex

Can you try that for the time being?

I'll rename this issue as a feature request to integrate Zed's client instead.

0x6362:
So I actually dug into this a bit tonight and it appears that Zed have actually taken the client portion of `openai/codex` into their ACP server so that they can piggyback off their login server code, [see here](https://github.com/zed-industries/codex-acp/blob/333e3f9ee2c7bd16f5d976e7a6b71795a74d23ad/src/codex_agent.rs#L148).

I have a branch that _roughly works_, if you're willing to boot up a `codex` client to handle the initial grab of an OAuth token en lieu of implementing the entire OAuth flow. 

@xenodium wasn't sure what approach you may want to take here and didn't want to presume - the hackiest is probably another Rust binary that wraps and serves up the necessary local server for the callback, least reverse engineering it and doing it within elisp (there must be a package that does at least part of this?). Thoughts?

kvaneesh:
That didn't work with :api-key auth mehtod. I get the error 
```
((:object (jsonrpc . "2.0") (id . 2)
	  (error (code . -32602) (message . "Invalid params")
		 (data . "unsupported authentication method")))
 (:json
  . "{\"jsonrpc\":\"2.0\",\"id\":2,\"error\":{\"code\":-32602,\"message\":\"Invalid params\",\"data\":\"unsupported authentication method\"}}"))
```

xenodium:
> That didn't work with :api-key auth mehtod. I get the error

[...]

PR #75, opened 2025-10-17
#72: Add OpenAPI Codex support for ChatGPT Subscription Authentication

0x6362:
Quoting myself from the discussion on #72:

> it appears that Zed have actually taken the client portion of openai/codex into their ACP server so that they can piggyback off their login server code, [see here](https://github.com/zed-industries/codex-acp/blob/333e3f9ee2c7bd16f5d976e7a6b71795a74d23ad/src/codex_agent.rs#L148).
> 
> I have a branch that roughly works, if you're willing to boot up a codex client to handle the initial grab of an OAuth token en lieu of implementing the entire OAuth flow.
> 
> @xenodium wasn't sure what approach you may want to take here and didn't want to presume - the hackiest is probably another Rust binary that wraps and serves up the necessary local server for the callback, least reverse engineering it and doing it within elisp (there must be a package that does at least part of this?). Thoughts?

[Looking more closely at cola-io/codex-acp](https://github.com/cola-io/codex-acp/blob/main/src/agent.rs#L568), thinking about it further, there's no reason any auth should be necessary within emacs so long as a user has run `codex login`. `codex-rs` should handle the reauth when its token expires if the refresh token is still valid. I've yet to dabble with Rust, so entirely possible I'm misunderstanding what's going on here.

### DIFF
--- a/agent-shell-openai.el
+++ b/agent-shell-openai.el
@@ -95,7 +95,7 @@ Returns an agent configuration alist using `agent-shell-make-agent-config'."
    :needs-authentication t
    :authenticate-request-maker (lambda ()
                                  (cond ((map-elt agent-shell-openai-authentication :api-key)
-                                        (agent-shell-openai-key))
+                                        (acp-make-authenticate-request :method-id "OpenAiApiKey"))
                                        (t
                                         (acp-make-authenticate-request :method-id "ChatGpt"))))
    :client-maker (lambda (buffer)


README.org
```diff
@@ -46,7 +46,7 @@ For Google's [[https://github.com/google-gemini/gemini-cli][Gemini CLI]], be sur
 
 *** Codex
 
-For OpenAI's Codex, install [[https://github.com/cola-io/codex-acp][codex-acp]] and ensure the `codex-acp` executable is in PATH.
+For OpenAI's Codex, install [[https://github.com/cola-io/codex-acp][cola-io/codex-acp]] or [[https://github.com/zed-industries/codex-acp][zed/codex-acp]] and ensure the `codex-acp` executable is in PATH.
```
xenodium, line 49: How about we just pick one? Whichever one works best and/or more likely to be maintained. My guess is Zed's but I'll leave it up to you to decide, since you're likely using Codex?

xenodium (approved):
Thank you for the PR!

xenodium:
> thinking about it further, there's no reason any auth should be necessary within emacs so long as a user has run codex login.

Yup. I agree. I think :login t is enough. Merging!

PR #79, opened 2025-10-19
Add ACP session mode cycling and modeline display

ElleNajt:
Implements support for the ACP session modes feature, which allows viewing and switching between different agent permission modes (e.g., "Always Ask", "Accept Edits", "Plan Mode", "Bypass Permissions").

## Changes
- Add `agent-shell-cycle-mode` interactive command to cycle through modes
- Display current mode in modeline as `[mode-name]` in `mode-line-misc-info`
- Store mode state (availableModes, currentModeId) from session/new response
- Mode display only appears if agent supports modes (optional ACP feature)
- Use `:eval` for automatic modeline updates when state changes

## Benefits
The current mode is always visible in the modeline when available, making it easy to see which permission mode is active and quickly switch between them while working with the agent.

## Screenshot/Demo
The modeline shows `[Accept Edits]` or similar, and users can call `M-x agent-shell-cycle-mode` to switch between available modes.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -184,7 +184,9 @@ and AUTHENTICATE-REQUEST-MAKER."
         (cons :chunked-group-count 0)
         (cons :request-count 0)
         (cons :tool-calls nil)
-        (cons :available-commands nil)))
+        (cons :available-commands nil)
+        (cons :available-session-modes nil)
+        (cons :current-session-mode-id nil)))
 
 (defvar-local agent-shell--state
     (agent-shell--make-state))
@@ -1427,18 +1429,17 @@ Must provide ON-SESSION-INIT (lambda ())."
   (acp-send-request
    :client (map-elt (agent-shell--state) :client)
    :request (acp-make-session-new-request :cwd (agent-shell--resolve-path (agent-shell-cwd)))
+   :buffer (current-buffer)
    :on-success (lambda (response)
                  (map-put! agent-shell--state
                            :session-id (map-elt response 'sessionId))
-                 ;; Store available modes and current mode
+                 ;; Store available session modes and current mode
                  (when-let ((modes-data (map-elt response 'modes)))
-                   (setf (map-elt agent-shell--state :available-modes)
-                         (map-elt modes-data 'availableModes))
-                   (setf (map-elt agent-shell--state :current-mode-id)
-                         (map-elt modes-data 'currentModeId))
-                   ;; Update modeline to show mode
-                   (with-current-buffer (map-elt agent-shell--state :buffer)
-                     (force-mode-line-update)))
+                   (map-put! agent-shell--state :available-session-modes
+                             (map-elt modes-data 'availableModes))
+                   (map-put! agent-shell--state :current-session-mode-id
+                             (map-elt modes-data 'currentModeId))
+                   (force-mode-line-update))
                  (agent-shell--update-dialog-block
                   :state agent-shell--state
                   :block-id "starting"
@@ -1474,6 +1475,7 @@ Must provide ON-SESSION-INIT (lambda ())."
              :session-id (map-elt agent-shell--state :session-id)
              :prompt `[((type . "text")
                         (text . ,(substring-no-properties prompt)))])
+   :buffer (current-buffer)
    :on-success (lambda (response)
                  ;; Tool call details are no longer needed after
                  ;; a session prompt request is finished.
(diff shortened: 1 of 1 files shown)


agent-shell.el
```diff
@@ -1429,6 +1430,15 @@ Must provide ON-SESSION-INIT (lambda ())."
    :on-success (lambda (response)
                  (map-put! agent-shell--state
                            :session-id (map-elt response 'sessionId))
+                 ;; Store available modes and current mode
+                 (when-let ((modes-data (map-elt response 'modes)))
+                   (setf (map-elt agent-shell--state :available-modes)
```
xenodium: Could we please add to agent-shell--make-state, so we get an overall picture of all values available in state?

agent-shell.el
```diff
@@ -1429,6 +1430,15 @@ Must provide ON-SESSION-INIT (lambda ())."
...
+                         (map-elt modes-data 'availableModes))
+                   (setf (map-elt agent-shell--state :current-mode-id)
```
xenodium: Same for :current-permission-mode-id please (include in agent-shell--make-state).

agent-shell.el
```diff
@@ -1429,6 +1430,15 @@ Must provide ON-SESSION-INIT (lambda ())."
...
+                         (map-elt modes-data 'currentModeId))
+                   ;; Update modeline to show mode
+                   (with-current-buffer (map-elt agent-shell--state :buffer)
```
xenodium: with-current-buffer should't be needed anymore after https://github.com/xenodium/agent-shell/commit/3936be586dbaf9772f9104551f172f7b948a98cb Did you run into issues?

acp should now invoke callbacks with the intended context buffer

agent-shell.el
xenodium: Does (map-put! (agent-shell--state) :available-modes ...) work here? Same for others.

agent-shell.el
```diff
@@ -2029,6 +2039,71 @@ When DEACTIVATE is non-nil, deactivate region/selection."
               "\n"
               "```"))))
 
+;;; Mode cycling
+
+(defun agent-shell--get-mode-name (mode-id available-modes)
```
xenodium: Nit. Maybe name gent-shell--resolve-permission-mode-name?

agent-shell.el
```diff
@@ -2029,6 +2039,71 @@ When DEACTIVATE is non-nil, deactivate region/selection."
...
+  "Get the name of the mode with MODE-ID from AVAILABLE-MODES.
+AVAILABLE-MODES is the list of mode objects from the ACP session/new response.
```
xenodium: A link to the corresponding model would be helpful here https://agentclientprotocol.com/protocol/schema

agent-shell.el
```diff
@@ -2029,6 +2039,71 @@ When DEACTIVATE is non-nil, deactivate region/selection."
...
+Each mode has an 'id' and 'name' field. We look up the mode by ID to get its
+display name."
+  (when-let ((mode (cl-find mode-id available-modes
+                            :key (lambda (m) (map-elt m 'id))
+                            :test #'string=)))
+    (map-elt mode 'name)))
+
+(defun agent-shell--mode-line-format ()
+  "Return the mode-line format for displaying the current permission mode."
```
xenodium: Including a sample modeline in the docstring would be helpful for anyone reading the code.

agent-shell.el
```diff
@@ -2029,6 +2039,71 @@ When DEACTIVATE is non-nil, deactivate region/selection."
...
+  (when (derived-mode-p 'agent-shell-mode)
+    (let* ((state (agent-shell--state))
```
xenodium: Can't we combine let* with when-let as when-let*?

agent-shell.el
```diff
@@ -2029,6 +2039,71 @@ When DEACTIVATE is non-nil, deactivate region/selection."
...
+           (current-mode-id (map-elt state :current-mode-id))
+           (available-modes (map-elt state :available-modes)))
+      (when-let ((mode-name (and current-mode-id available-modes
+                                 (agent-shell--get-mode-name current-mode-id available-modes))))
+        (propertize (format " [%s]" mode-name)
+                    'face 'font-lock-constant-face
+                    'help-echo (format "ACP Permission Mode: %s" mode-name))))))
```
xenodium: Nit: Let's omit ACP here (it's more of an implementation detail).

agent-shell.el
xenodium: With the term "mode" being so overriden in Emacs, how about we qualify and name something like `:permission-modes`.

agent-shell.el
```diff
@@ -2029,6 +2039,71 @@ When DEACTIVATE is non-nil, deactivate region/selection."
...
+
+(defun agent-shell--setup-modeline ()
+  "Set up the modeline to display permission mode.
+Uses :eval so the mode updates automatically when state changes."
+  (setq-local mode-line-misc-info
+              (append mode-line-misc-info
+                      '((:eval (agent-shell--mode-line-format))))))
+
+(defun agent-shell-cycle-mode ()
```
xenodium: Ditto: agent-shell-cycle-*permission*-mode

agent-shell.el
```diff
@@ -2029,6 +2039,71 @@ When DEACTIVATE is non-nil, deactivate region/selection."
...
+                               :key (lambda (mode) (map-elt mode 'id))
+                               :test #'string=)))
+      ;; Send request to change mode
+      (acp-send-request
+       :client (map-elt state :client)
+       :request (acp-make-session-set-mode-request
+                 :session-id session-id
+                 :mode-id next-mode-id)
+       :on-success (lambda (_response)
+                     (setf (map-elt agent-shell--state :current-mode-id) next-mode-id)
```
xenodium: map-put! if possible please.

agent-shell.el
```diff
@@ -2029,6 +2039,71 @@ When DEACTIVATE is non-nil, deactivate region/selection."
...
+                     (with-current-buffer (map-elt state :buffer)
```
xenodium: We can omit with-current-buffer here, right? Lemme know if otherwise as it would be a bug.

agent-shell.el
```diff
@@ -2029,6 +2039,71 @@ When DEACTIVATE is non-nil, deactivate region/selection."
...
+                       (force-mode-line-update)))
+       :on-failure (lambda (error)
```
xenodium: Check out other usages.

```emacs-lisp
:on-failure (lambda (error raw-message)
...
```

xenodium (commented):
Thanks a lot for the PR!

Left some minor comments. Hope that's ok.

xenodium:
ps. I'm guessing there's also an acp.el PR coming for `acp-make-session-set-mode-request`?

ElleNajt:
Oops sorry Claude eagerly force pushed before I got a chance to review the changes it made in response to the feedback, reading it over now.

ElleNajt:
Yes -- thanks for pointing that out! Here it is: https://github.com/xenodium/acp.el/pull/4

Though, let me sync them both to have "permission-mode" instead of "mode", I agree that that term is overloaded. 

EDIT: session-mode instead of permission-mode, to be consistent with acp https://agentclientprotocol.com/protocol/session-modes

xenodium:
ps. Two more things (please ignore if you already did):

1. Run M-x checkdoc
2. Run M-x byte-compile-file

and fix any warnings please ;)

agent-shell.el
ElleNajt: Got rid of it, thank you! Using :buffer everywhere now and it's working.

agent-shell.el
ElleNajt: Yes, it works. Fixed!

agent-shell.el
ElleNajt: Using ` agent-shell--resolve-session-mode-name ` for consistency with acp protocol

agent-shell.el
ElleNajt: Linking to https://agentclientprotocol.com/protocol/session-modes

agent-shell.el
ElleNajt: 100% ty

agent-shell.el
ElleNajt: yes, we can.

agent-shell.el
ElleNajt: Addressed, thanks. Just sticking with a simple minibuffer message, unlike most of the other examples that use agent-shell--make-error-handler, bc I think there's no reason to have the failure of the cycling be in the agent window.

xenodium (approved):
Thank you for the follow-up changes.

xenodium:
Merged!

xenodium:
Modes are now listed on handshake too

modes

xenodium:
Screenshot 2025-10-20 at 13 32 15

Issue #74, opened 2025-10-17
feature request: support shell/bash mode


armindarvish:
Support typing `!` to enter shell/bash mode. In Claude code, one can use `!` to send a command directly to the shell; for example, `!ls -a` would send `ls -a` to the shell without going through Claude. I believe Gemini CLI has a similar mode, but it may need typing `!` and `RET`. Here is a screenshot:

xenodium:
Thanks for filing. The video is also helpful.

On the surface, this is relatively straightforward to implement as long as we keep it a basic way to execute a command and include its output as part of the prompt history. The UX would have to be a little different. We may need to send the shell command output as an FYI prompt to the agent. The implementation may start getting a bit more involved once we start considering edge cases.

Let's see how popular the feature request becomes. I may have a play with a prototype if I have some time available.

xenodium:
Ok, I have something going...

![Image](https://github.com/user-attachments/assets/a6879ddd-3e06-4367-9c9d-97cdf851bfe1)

It enables inserting command output into the prompt before sending. It gives you an opportunity to ask something specific about the command output before sending to agent.

It's not tied to "!" binding, so you'll need `M-x agent-shell-insert-shell-command-output` or your own custom binding. I think this is the best I can do for the time being. Hope that helps.

xenodium:
This is now pushed in https://github.com/xenodium/agent-shell/commit/3f89346658310345c4d15506914af3c08e281017

Can you please give it a try?

armindarvish:
@xenodium I just tested it and it works for me. As far as I am concerned this can be closed now.  

I think a binding in `agent-shell-mode-map` for `agent-shell-insert-shell-command-output` could be useful, but I am not sure what would be a good general approach. Perhaps, something like `C-c !`? and may be in the future more commands can be bound under `C-c` leader key.

xenodium:
> @xenodium I just tested it and it works for me. As far as I am concerned this can be closed now.

Awesome. Thanks for checking.

> a think a binding in agent-shell-mode-map for agent-shell-insert-shell-command-output could be useful

Yep. Good idea. I just don't know which binding yet. At some point I'll consider a transient menu.

armindarvish:
[...]

PR #84, opened 2025-10-20
Add bash command support with ! prefix

ElleNajt:
Add bash command support with ! prefix

This PR adds support for executing bash commands locally using the `!` prefix, matching Claude Code's behavior.

## Feature

Users can now run bash commands directly in agent-shell:
```
Claude Code> !ls
file1.txt  file2.txt

Claude Code> !pwd
/Users/elle/code/agent-shell
```

Commands are queued and bundled with the next message sent to the agent, formatted as `<bash-input>`, `<bash-stdout>`, `<bash-stderr>` tags that Claude Code expects.

## How it works

1. User types `!command` in agent-shell
2. Command executes locally and output is displayed immediately
3. Result is queued in `:bash-command-queue`
4. When user sends next message to agent, all queued bash outputs are prepended
5. Queue is cleared after sending

Multiple bash commands can be run before sending a message - they're all bundled together.

## Implementation Details

- **stderr capture**: Uses temporary file (required by `call-process` API)
- **Error display**: Non-zero exit codes show stderr and exit code to user
- **Working directory**: Commands execute in project working directory (`agent-shell-cwd`)

## Execution Context

Commands run in the **CLIENT environment** (your local machine), not the agent's environment. 

When the agent runs in a container (e.g., using devcontainer):
```
!hostname        → Shows your local machine name
agent Bash tool  → Executes in container, shows container ID
```

### Rationale for local-first approach

Ideally, we would support both execution contexts:
- **Local execution**: useful for local development and simple setups
- **Container execution**: matches agent's environment when containerized

However, container execution is more complex to implement. It would require:
- Refactoring configuration to separate container command from ACP command (e.g., `agent-shell-anthropic-container-command`)
- Storing container command in session state
- Handling working directory resolution inside containers
- The ACP protocol doesn't expose container or session identification needed for routing (as far as I could tell)

This implementation provides local execution first for immediate utility. A future enhancement can add container execution as an optional mode.

### Future UI suggestion

A nice UI for both contexts could be:
```
!command   → Execute locally (current implementation)
!!command  → Execute in agent's container (future enhancement)
```

## Example Configuration

For users running in containers:
```elisp
(setq agent-shell-anthropic-claude-command
      '("container-exec" "claude-code-acp"))
```

Currently, `!` commands run locally. Future enhancement would route `!!` commands through the container.

## Testing

- ✅ Byte-compile passes
- ✅ Checkdoc passes
- Tested with various commands including error cases

## Example Output

**Success:**
```
Claude Code> !echo "test"
test
```

**Error:**
```
Claude Code> !ls /nonexistent

ls: /nonexistent: No such file or directory

[exit code: 1]
```

**Multiple commands queued:**
```
Claude Code> !pwd
/Users/elle/code/agent-shell

Claude Code> !git status
On branch bash-command-support
nothing to commit, working tree clean

Claude Code> Now check these paths
[Agent receives both bash outputs bundled with the message]
```


xenodium:
Thanks for the PR. Please have a look at https://github.com/xenodium/agent-shell/issues/74 as we already have this feature (with exception of ! binding).

xenodium:
ps. Bummer. I hope you didn't spend much time on this one. In the future, please reach out by filing a feature request before implementing a feature.

Heads-up. I also have to be mindful of all the features I accept since ultimately I inherit all the code to maintain it. In some instances, I may suggest the feature lives in a separate package. For example: https://github.com/cmacrae/agent-shell-sidebar I'm also happy to link to all such projects from the README.

ElleNajt:
Oops, ty. No, not that much time.

Do you have takes regarding where the command executes? I think it useful be better to have the option to run it in the same environment as where the claude agent lives, to support the containerized or remote claude use case. I can open a feature request about that.

xenodium:
Yep. Please open a feature request so we can hash out what the feature would look like.

One request I would have is to keep the current behaviour of immediately inserting into buffer (short lived command) or using a separate buffer if the command takes longer to execute. Also no command queuing support yet as this opens up a bunch of edge cases that may need some infrastructure work.

I've yet to use the container support myself, but please loop in @fritzgrabo in the feature request. He contributed the container implementation we have today https://github.com/xenodium/agent-shell/pull/19

PR #87, opened 2025-10-21
Make qwen authentication optional

vindvaki:
When the right environment variables are provided, qwen-code can use a local server. In that case, it needs no authentication.

### DIFF
--- a/agent-shell-qwen.el
+++ b/agent-shell-qwen.el
@@ -35,11 +35,20 @@
 (declare-function agent-shell-make-agent-config "agent-shell")
 (declare-function agent-shell-start "agent-shell")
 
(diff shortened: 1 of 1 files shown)


xenodium:
Thank you for the PR!

Mind following the same pattern we introduced for the same purpose [here](https://github.com/xenodium/agent-shell/commit/6680858a1c858bdb38c7f889e10c47b76f5bf443)?

Client code would look like:

```emacs-lisp
  (setq agent-shell-qwen-make-authentication
        (agent-shell-qwen-make-authentication :none t))
```

vindvaki:
@xenodium thanks! I've made the change 🏓

xenodium:
> @xenodium thanks! I've made the change 🏓

Thanks for the follow-up

Issue #68, opened 2025-10-14
Show Edit Requests inline


Makesesama:
## Current State
Right now the if the agent wants to edit a file agent-shell asks for the tool permission. The user can view it if he wants.

## Proposal
Show the change or part of the change as preview in the agent-shell buffer like the default claude-code cli does.

xenodium:
Thanks for filing a feature request.

Ha! I did not expect this one :) Most folks I've spoken to are happy with the quick-diff buffer momentarily poping up (more room and less clutter), as well as the simple bindings/flow (n/p/q -> y/n/!) to manage the patch.

Are there specific bits about quick-diff that I could improve? I'm saying this as nailing a different diffing flow takes quite a bit of effort and iterations (quick-diff certainly did). I tried a handful of approaches before I landed on quick-diff.

Perhaps you could share more details about the things you like about claude-code cli diffing in addition to being in the same buffer? My hope with quick-diff was to make the transition fairly smooth.

Having said all this, as Emacsers we love the flexibility and options we have. I'm wondering if providing hooks for the edit events would cater for all? quick-diff as default, but can be replaced with another flow preferred by users.

Makesesama:
Thanks for responding so quickly.

I actually thought a bit before if the quick-diff workflow is really not what i would prefer, because i think it's actually really nice.

Nonetheless i really like seeing what is being done in the overview when using claude-code cli, because my workflow sometimes is, to just let claude-code go and implement a feature or fix a bug and review it afterwards. For that workflow i think actually seeing in which direction the llm is going is important, so i can stop it before the direction is totally wrong.

I also think providing a way to customize this would be the best for something in emacs.

Thanks for the awesome work :)

xenodium:
Thanks for this. It gives me more context. A couple of things that come to mind:

> For that workflow i think actually seeing in which direction the llm is going is important

Is the agent not pausing for you until you accept or reject proposed changes?

> I also think providing a way to customize this would be the best for something in emacs.

I'm wondering if the configuration point should be different than what I originally thought... Given you're interested on keeping sight of the agent shell, I'm thinking quick-diff window placement is what we're likely after. What if you could configure where quick-diff pops up to? Side window, bottom window, etc. This way the shell itself remains visible to you.

WDYT?

> Thanks for the awesome work :)

Thank you. That's nice to hear. Glad you like it.

Makesesama:
> Is the agent not pausing for you until you accept or reject proposed changes?

[...]

PR #89, opened 2025-10-21
Fix quick-diff window restoration for full-screen

nineluj:
## Summary

Fixes a bug in `quick-diff` where the calling buffer would not be properly restored after exiting the diff view when displayed in a full-screen window.

## Problem

When `quick-diff` is displayed in a full-screen window, the diff buffer replaces the calling buffer in that window. On exit, the original code only checked if the calling buffer was visible in *some* window, but didn't verify it was still in the *original* calling window. As a result, the calling buffer was not restored to the window where the user initiated the diff.

## Solution

Now captures the calling window at the start and checks on exit if:
1. The original calling window is still live
2. The calling buffer is still displayed in that window

If not, it explicitly restores the calling buffer to the original window before selecting it.

## Test plan
See videos for the old and new behavior. The behavior when there are multiple windows remains unchanged.

https://github.com/user-attachments/assets/5e400d94-b1fb-4afe-ae5d-42d7f6360015

https://github.com/user-attachments/assets/ac487b79-6f40-4127-843b-35d84ad2259b

### DIFF
--- a/README.org
+++ b/README.org
@@ -322,6 +322,114 @@ All of the above settings can be applied on a per-project basis using [[https://
 - =C-c C-c= - Interrupt current agent operation
 - =TAB and Shift-TAB= - Navigate interactive elements
 
+* Contributing
+
+** Before Contributing
+
+Before implementing new features, please *file a feature request first* to discuss the proposal. This helps ensure alignment with the project's direction and prevents unnecessary work.
+
+As the maintainer, I must be mindful of all features I accept since I inherit the code to maintain it. Some features may be better suited as separate packages (like [[https://github.com/cmacrae/agent-shell-sidebar][agent-shell-sidebar]]).
+
+I'll gladly promote your package wherever possible.
+
+** Style (or personal preference TBH)
+
+There are lots of ways to accomplish things in elisp. While the following are merely personal preferences, as maintainer, it really simplifies things for me to try to limit the number of ways to accomplish things.
+
+*** Maps (use alists)
+
+This project relies on [[https://www.gnu.org/software/emacs/manual/html_node/elisp/Association-Lists.html][alists]] for much of its functionality. Sure, we can also use plists, hashtables, etc.
+
+Unless we have a strong argument to use something else, please stick with =alists= (and =:= keywords).
+
+#+begin_src emacs-lisp
+  '((:species . "Cat")
+    (:name . "Whiskers")
+    (:age . 4)
+    (:color . "Gray")
+    (:favorite-toy . "Feather Wand"))
+#+end_src
+
+*** map.el
+
+Accessing and working with =alists=? Please prefer =map.el= unless we have a strong argument to use an alternative.
+
+#+begin_src emacs-lisp :lexical no
+  (setq animal (seq-first animals))
+  (map-elt animal :species)
+#+end_src
+
+*** seq.el
+
+Accessing and working with lists? Please prefer =seq.el=, unless we have a strong argument to use an alternative.
+
+#+begin_src emacs-lisp :lexical no
+  (setq animals
+        (list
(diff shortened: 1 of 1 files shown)


nineluj:
Tangentially related to #68. I decided not to add any more logic, like configuration for where the diff should pop up, while that gets figured out in that other issue thread.

xenodium:
Btw, I noticed you get a red section in quick-diff, which I don't.

Screenshot 2025-10-22 at 08 58 49

Screenshot 2025-10-22 at 08 57 30

Mind placing point on it, doing M-x describe-face, and reporting what face that is?

ps. While I'm a dark-theme guy, your light one looks awesome!

nineluj:
Thanks! I'm using the `ef-summer` theme from [ef-themes](https://github.com/protesilaos/ef-themes). I used to be a die hard dark theme person but I'm loving this one.

`M-x describe-char`
```
             position: 168 of 333 (50%), column: 0
            character: @ (displayed as @) (codepoint 64, #o100, #x40)
              charset: ascii (ASCII (ISO646 IRV))
code point in charset: 0x40
               script: latin
               syntax: . 	which means: punctuation
             category: .:Base, a:ASCII, l:Latin, r:Roman
             to input: type "C-x 8 RET 40" or "C-x 8 RET COMMERCIAL AT"
          buffer code: #x40
            file code: #x40 (encoded by coding system utf-8-unix)
              display: by this font (glyph code):
    mac-ct:-*-Iosevka Nerd Font Mono-bold-normal-normal-*-20-*-*-*-m-0-iso10646-1 (#x23)

Character code properties: customize what to show
  name: COMMERCIAL AT
  general-category: Po (Punctuation, Other)
  decomposition: (64) ('@')

There are 3 overlays here:
 From 168 to 334
  diff--font-lock-refined t
  diff-mode            fine
  evaporate            t
  modification-hooks   (diff--overlay-auto-delete)
 From 168 to 334
  diff--font-lock-syntax t
  diff-mode            syntax
  evaporate            t
  modification-hooks   (diff--overlay-auto-delete)
 From 168 to 184
  before-string        [Show]
  category             diff-header
  display              [Show]
  evaporate            t


There are text properties here:
  face                 diff-hunk-header
  fontified            t

[back]
```

and `M-x describe-face`
```
Face: diff-hunk-header (sample) (customize this face)

Documentation:
‘diff-mode’ face used to highlight hunk header lines.

Defined in ‘diff-mode.el’.


           Family: unspecified
          Foundry: unspecified
            Width: unspecified
           Height: unspecified
           Weight: unspecified
            Slant: unspecified
       Foreground: unspecified
DistantForeground: unspecified
       Background: #f7ebee
        Underline: unspecified
         Overline: unspecified
   Strike-through: unspecified
              Box: unspecified
          Inverse: unspecified
          Stipple: unspecified
             Font: unspecified
          Fontset: unspecified
           Extend: unspecified
          Inherit: bold
```

The theme is based on the framework from modus-themes, which does override this value:
https://github.com/protesilaos/modus-themes/blob/main/modus-themes.el#L5012

I tried out another theme (doom-palenight) and everythings looks normal, so there's probably just an issue with the ef theme:
image

xenodium:
Thanks for this. I suspect it's the combination of the theme and the overlay usage. From the light-themed quick-diff buffer, can you M-x eval-expression RET (remove-overlays) RET and post a screenshot please?

nineluj:
Oh odd, after updating my version of ef-themes to the last commit and it now seems to be good. I'm also using a transparent Emacs window so the desktop wallpaper is showing.

# Light theme
image

# Dark theme
image

Thanks for the follow-up @xenodium !

xenodium:
Ah awesome! Much better. Glad its all good now.

PR #91, opened 2025-10-22
Add container execution support via agent-shell-container-command-runner

### LINKED ISSUE #85: Support executing shell commands in containers
## Problem

Currently, `agent-shell-insert-shell-command-output` executes shell commands locally using `shell-file-name`, even when the agent itself runs in a container. This creates inconsistency:

1. **User-initiated shell commands** (via `agent-shell-insert-shell-command-output`) execute on the local machine
2. **Agent-initiated commands** execute in the container (e.g., devcontainer)

This means when the user runs a shell command and inserts its output, Claude sees the local filesystem instead of the container's filesystem where it's actually running, leading to confusion and errors.

### Option 1: Unified Container Wrapper

(Have a version of the first option in my fork https://github.com/ElleNajt/agent-shell/tree/container-wrapper-support , though only refactored the claude-command appropriately.)

Add a single configuration for container execution that applies to both agent and shell commands:

```elisp
(defcustom agent-shell-container-wrapper nil
  "Command prefix for executing commands in a container.
When non-nil, both the agent command and shell commands will be
executed using this wrapper. Should be a list of command arguments.

Example for devcontainer:
  '(\"devcontainer\" \"exec\" \"--workspace-folder\" \".\")"
  :type '(repeat string)
  :group 'agent-shell)
```

Configuration:
```elisp
(setq agent-shell-container-wrapper
      '("devcontainer" "exec" "--workspace-folder" "."))
(setq agent-shell-anthropic-claude-command 
      '("claude-code-acp"))
```

**Benefits:**
- Single place to configure container execution
- Easy to toggle between container and local mode by setting this variable to be nil

**Tradeoffs:**
- Requires custom container setups to use the same wrapper for both agent and shell commands
- Less flexible if different execution contexts are needed

### Option 2: Separate Shell Wrapper

Keep agent command configuration unchanged and add separate shell wrapper:

```elisp
(defcustom agent-shell-container-shell-wrapper nil
  "Command prefix for executing shell commands in a container.
When non-nil, shell commands will be executed using this wrapper."
  :type '(repeat string)
  :group 'agent-shell)
```

Configuration:
```elisp
(setq agent-shell-anthropic-claude-command 
      '("devcontainer" "exec" "--workspace-folder" "." "claude-code-acp"))
(setq agent-shell-container-shell-wrapper
      '("devcontainer" "exec" "--workspace-folder" "."))
```

**Benefits:**
- More flexible - allows different execution contexts for agent vs shell if needed
- No changes to existing agent command configuration

**Tradeoffs:**
- Repetitive configuration for common case where both use same container
- Harder to toggle modes (need to update both variables)

## Example Configuration

Using devcontainer with Option 1:
```elisp
(setq agent-shell-container-wrapper
      '("devcontainer" "exec" "--workspace-folder" "."))
(setq agent-shell-anthropic-claude-command 
      '("claude-code-acp"))
(setq agent-shell-path-resolver-function 
      #'agent-shell--resolve-devcontainer-path)
```

Using devcontainer with Option 2:
```elisp
(setq agent-shell-anthropic-claude-command 
      '("devcontainer" "exec" "--workspace-folder" "." "claude-code-acp"))
(setq agent-shell-container-shell-wrapper
      '("devcontainer" "exec" "--workspace-folder" "."))
(setq agent-shell-path-resolver-function 
      #'agent-shell--resolve-devcontainer-path)
```

## Additional Context

- This builds on PR #19's container support, which only handles path mapping
- Both approaches would provide parity between file operations and shell command execution in containerized environments

cc @fritzgrabo who implemented PR #19

ElleNajt:
Fixes #85

## Summary

Fixes environment inconsistency when running agents in containers.

**Problem:** Currently, when an agent runs in a container (e.g., via devcontainer), there's an inconsistency:
- Agent-initiated commands execute inside the container
- User-initiated shell commands (via `agent-shell-insert-shell-command-output`) execute on the local machine

This causes the agent to see the local filesystem when users insert shell command output, while the agent itself operates in the container's filesystem, leading to confusion and errors.

**Solution:** Add `agent-shell-container-command-runner` to wrap both agent commands and shell commands with the same container execution prefix, ensuring consistent execution context.

## Changes

- Add `agent-shell-container-command-runner` defcustom to specify container wrapper command prefix
- Add `agent-shell--make-acp-client` wrapper function to optionally prepend container runner to agent commands

## Usage

Set `agent-shell-container-command-runner` to a list of command arguments:

```elisp
;; For devcontainer
(setq agent-shell-container-command-runner '("devcontainer" "exec" "--workspace-folder" "."))
```

When set, both the agent command and shell commands will be executed using this runner.

## Example

Without container runner:
```elisp
;; Command: claude-code-acp --flag
```

With container runner:
```elisp
(setq agent-shell-container-command-runner '("devcontainer" "exec" "--workspace-folder" "."))
;; Command becomes: devcontainer exec --workspace-folder . claude-code-acp --flag
```

## Implementation Details

- The wrapper is applied in `agent-shell--make-acp-client` by prepending runner args to command params
- When `agent-shell-container-command-runner` is nil (default), behavior is unchanged (backward compatible)
- Both the main agent command and any shell commands spawned by the agent will use the container runner

## Testing

- ✅ Verified agent starts successfully with devcontainer wrapper
- ✅ Verified agent executes commands inside container
- ✅ Verified backward compatibility when runner is nil (default)
- ✅ Byte-compile passes
- ✅ Checkdoc passes

### Per-Session Container Override

It would be useful to allow switching between container and host execution modes on a per-session basis (e.g., via prefix argument like `C-u SPC o c`). This would let users:
- Run some agent sessions in the container (the default when `agent-shell-container-command-runner` is set)
- Run other agent sessions on the host machine (by overriding with a prefix arg)

**Working workaround:** Per-session container override can be implemented with buffer-local state and advice:

We ... may want to design parts of the system to make this simpler.

```elisp
(defvar-local my/agent-shell--force-local nil
  "When non-nil, disable container wrapper for this buffer.")

;; Advice to disable path resolver for host-mode buffers
(defun my/agent-shell--resolve-path-advice (original-fn path)
  "Disable resolver when force-local is set."
  (if (and (local-variable-p 'my/agent-shell--force-local)
           my/agent-shell--force-local)
      path  ; Return path unchanged
    (funcall original-fn path)))

(advice-add 'agent-shell--resolve-path :around #'my/agent-shell--resolve-path-advice)

;; Wrapper to start agent with prefix arg support
(defun my/start-claude-code (force-local)
  "Start Claude Code, optionally on host with C-u prefix."
  (interactive "P")
  (let ((agent-shell-container-command-runner
         (if force-local nil agent-shell-container-command-runner))
        (agent-shell-path-resolver-function
         (if force-local nil agent-shell-path-resolver-function)))
    (let ((config (agent-shell-anthropic-make-claude-code-config)))
      ;; Replace client-maker to establish fresh bindings when called
      (map-put! config :client-maker
                (lambda (buffer)
                  (with-current-buffer buffer
                    (setq-local my/agent-shell--force-local force-local))
                  (let ((agent-shell-container-command-runner
                         (if force-local nil agent-shell-container-command-runner))
                        (agent-shell-path-resolver-function
                         (if force-local nil agent-shell-path-resolver-function)))
                    (agent-shell-anthropic-make-claude-client :buffer buffer))))
      (agent-shell-start :config config))))
```

**Per Claude: Why this is messy - the timing issue:**

When you start an agent, here's what happens in order:
1. Your wrapper creates config with `let` bindings (container-runner=nil, path-resolver=nil)
2. Config is passed to `agent-shell-start`
3. `agent-shell-start` creates the buffer and sets up shell-maker
4. **Minutes later**, user types a command
5. NOW the client-maker lambda is finally called
6. Problem: The `let` bindings from step 1 are long gone

This creates three challenges:
1. **Client creation timing**: Can't use simple `let` bindings - they expire before the client-maker runs. Solution: Create fresh bindings inside the lambda itself.
2. **Path resolution timing**: Happens even later during `session/new` (after client creation). Solution: Use buffer-local flag + advice to intercept path resolution.
3. **Isolation**: The advice must only affect the specific buffer, not other sessions. Solution: Check `local-variable-p` before applying override.

The verbosity comes from repeating the binding pattern (outer + inner), and needing advice because path resolution happens in a separate call chain from client creation.

**Potential framework improvement:** If `agent-shell-start` created the client eagerly (during step 3) instead of lazily (during step 5), the `let` bindings would still be active and the lambda wrapper wouldn't be needed. This would simplify per-session overrides significantly:

```elisp
(defun my/start-claude-code (force-local)
  (interactive "P")
  (let ((agent-shell-container-command-runner
         (if force-local nil agent-shell-container-command-runner))
        (agent-shell-path-resolver-function
         (if force-local nil agent-shell-path-resolver-function)))
    (agent-shell-anthropic-start-claude-code)))  ;; Just works!
```

(Path resolution would still need advice since it happens during `session/new`, but the client-maker complexity would disappear.)


xenodium (approved):

xenodium:
Thank you for the discussion and iterating on things. PR is now merged. 🎉

I'll have a think about the switching mechanism.

PR #95, opened 2025-10-23
Fall back to default devcontainer workspace folder if none given

### LINKED ISSUE #94: Don't require `workspaceFolder` setting for devcontainers, fall back to default value
The current implementation of `agent-shell--get-devcontainer-workspace-path` requires a `workspaceFolder` to be set in `.devcontainer/devcontainer.json` to work, and signals an error otherwise.

This is inconvenient since the [Dev Container metadata reference](https://containers.dev/implementors/json_reference/#image-specific) says that the `workspaceFolder` setting is optional. If none is given, a default value of `/workspaces/<repository-name>` is used.

Ideally, `agent-shell--get-devcontainer-workspace-path` would also fall back to the same default value, and not signal an error.

PR coming up.

fritzgrabo:
Changes `agent-shell--get-devcontainer-workspace-path`, which determines the path under which the current project is mounted inside a devcontainer. This path is used to resolve devcontainer paths on the local filesystem and vice versa.

Update: if no `workspaceFolder` setting is found in `.devcontainer/devcontainer.json`, do not signal an error. Instead, fall back to the default value for this setting, which is `/workspaces/<repository-name>`.

Fixes #94

### DIFF
--- a/README.org
+++ b/README.org
@@ -297,18 +297,16 @@ Use =agent-shell-container-command-runner= to prefix the command that starts the
 (setq agent-shell-container-command-runner '("devcontainer" "exec" "--workspace-folder" "."))
 #+end_src
 
-Note that any =:environment-variables= you may have passed to =acp-make-client= will not apply to the agent process running inside the container.
-It's expected to inject environment variables by means of your devcontainer configuration / Dockerfile.
+Note that any =:environment-variables= you may have passed to =acp-make-client= will not apply to the agent process running inside the container. It's expected to inject environment variables by means of your devcontainer configuration / Dockerfile.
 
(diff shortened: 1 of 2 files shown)


xenodium:
Thanks for the PR. Merged. I'm glad you proposed the fix to the issue, as I've yet to start using devcontainers.

fritzgrabo:
Thanks for the quick merge. Also no rush wrt starting to use devcontainers -- I realize this is a sideshow right now, and marked as experimental in the README. If an issue related to devcontainers pops up, please do tag me in it and I'll do my best to help. @ElleNajt seemed interested as well. Thanks!

ElleNajt:
Thanks for this! I have a hack in my personal config that does something like this atm, will switch over. :)

PR #97, opened 2025-10-24
Add live transcript auto-save to .agent-shell/transcripts/

### LINKED ISSUE #63: Configuration to save transcripts automatically to a given location
I like how you have the option to save a transcript of the conversation when closing the shell. I'd like to propose:

1. An option to save this automatically without prompting
2. A second option to configure a directory to save to (with some sensible default)
3. A way to name the transcript automatically, if possible (e.g. how Claude Code sets the terminal title based on the work done), or otherwise using a template or sensible default.

Would like to contribute this if it seems useful.

ElleNajt:
This PR implements automatic transcript saving that captures complete conversation history for archival purposes.

## Features

**Auto-saves transcripts as conversations progress:**
- User prompts with timestamps
- Agent responses (streamed in real-time)
- Tool calls with status, commands, and output

**Enabled by default** - transcripts save to `.agents/transcripts/{timestamp}.txt`

**Note on format:** Transcripts are optimized for human readability (timestamped, formatted text). [Edit: This next sentence is wrong - ACP does support[ restoring sessions](https://agentclientprotocol.com/protocol/session-setup#loading-sessions)] While ACP doesn't currently support restoring sessions, a future enhancement could add a parallel machine-readable format for session restoration.

**Configurable:**
```elisp
;; Disable if desired
(setq agent-shell-auto-save-transcript nil)

;; Change directory
(setq agent-shell-transcript-directory ".my-transcripts")

;; Customize filename generation
(setq agent-shell-transcript-filename-function
      (lambda () (format "chat-%s.txt" (format-time-string "%Y%m%d"))))
```

## Relationship to shell-maker

This is **separate** from shell-maker's manual save functionality:
- shell-maker: Manual save on-demand or prompted save-on-close
- This feature: Automatic background archival to project directory

Users can use both, either, or neither depending on their needs.

## Example transcript format

```
Agent Shell Transcript
Started: 2025-10-24 18:00:00
Working Directory: /Users/elle/code/agent-shell

============================================================

[2025-10-24 18:00:05] USER:
list files

[2025-10-24 18:00:06] AGENT:
I'll list the files in the current directory.

[2025-10-24 18:00:07] TOOL CALL [completed]:
  Tool: bash - Execute bash command
  Title: List files
  Command: ls -la
  Output:
total 256
drwxr-xr-x  15 elle  staff   480 Oct 24 18:00 .
...
```

Closes #63 - Auto-save transcripts without prompting

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1971,9 +1971,6 @@ Returns list of alists with :start, :end, and :path for each mention."
     (agent-shell-heartbeat-start
      :heartbeat (map-elt agent-shell--state :heartbeat))
 
-    (when attached-files (agent-shell--display-attached-files attached-files))
-
-
     (map-put! agent-shell--state :last-entry-type nil)
 
     (agent-shell--append-transcript


ElleNajt:
Another benefit of doing something like this is that the buffers start to get laggy around 1M, and I'd like to have something like 

```elisp

(add-hook 'agent-shell-mode-hook
          (lambda ()
            (setq-local comint-buffer-maximum-size 50000)
            (add-hook 'comint-output-filter-functions 'comint-truncate-buffer nil t)))
```

In my personal config to keep the buffers manageable, and be able to look up the history if desired. 

Perhaps I should open the that as a separate issue though.

xenodium:
Thanks for the PR. Sorry, not had a chance to review. May need a little time...

> Another benefit of doing something like this is that the buffers start to get laggy around 1M, and I'd like to have something like

1M characters? Something else? Could you please file a bug about this with more details? Bonus points if you investigate your laggy case. I spent some time recently chasing optimizations after a friend mentioned laggy buffers for codex agent. These helped, but I'm sure we can optimize elsewhere.

- https://github.com/xenodium/agent-shell/commit/01b17aad9d4ec5210f3d4a85e94289dfb62f1f63
- https://github.com/xenodium/agent-shell/commit/f1597397ae3cdcd7ecd06b81618508dedba34842

Also, when reproducing the laggy scenario, make sure you update to the latest acp.el and logging is turned off (including saving any traffic), via M-x agent-shell-toggle-logging. Older acp.el versions saved traffic regardless.

Btw, you prolly already know this (and not a solution), but the "clear" command wipes the content of the buffer but doesn't wipe agent history.

Also, the markdown-overlay implementation is super expensive, specially for code blocks. If you comment out all of the `(markdown-overlays-put)`, does your buffer still get laggy at 1M?

ElleNajt:
1M characters, yes.

> Btw, you prolly already know this (and not a solution), but the "clear" command wipes the content of the buffer but doesn't wipe agent history.

Ah, I didn't know this. I thought it was like /clear in claude, which does clear the agents context window. tbh, probably clear *should* mean to clear the context window ? I don't see a method in ACP for this though. 

> Could you please file a bug about this with more details?
Yes!

btw, looked more at ACP, and it does support restoring sessions: https://agentclientprotocol.com/protocol/session-setup#loading-sessions

xenodium:
> Ah, I didn't know this. I thought it was like /clear in claude, which does clear the agents context window. tbh, probably clear should mean to clear the context window ? I don't see a method in ACP for this though.

We're getting the "clear" command from "shell-maker" which merely clears shell content. I'm in two minds about also making it clear the agent. The convention seems to be that / is needed to operate on agent, though I can see how what we have today is confusing.

> btw, looked more at ACP, and it does support restoring sessions: https://agentclientprotocol.com/protocol/session-setup#loading-sessions

While it's in the protocol, not all agents seem to support it. I've added a display of agent capabilities to surface these details more prominently

Screenshot 2025-10-26 at 15 14 11

agent-shell.el
```diff
@@ -91,6 +91,32 @@ Example for devcontainer:
   :type '(repeat string)
   :group 'agent-shell)
 
+(defcustom agent-shell-auto-save-transcript t
```
xenodium: Before opening the feature up for customizations, I'd like to give it some time to settle. This gives us time to use it, get a feel for it, and tweak without necessarily breaking users (if we decide to change settings around). Also gives us an opportunity to figure out how to reconcile this feature with the transcript feature in `shell-maker` (we need to turn that off, for example).

I'd say let's drop `agent-shell-auto-save-transcript` (we default to autosave) and `agent-shell-transcript-directory` for now. Let's rename `agent-shell-transcript-filename-function` to `agent-shell--transcript-file-path-function` (also private defvar).

For now if anyone would like to disable things, set `agent-shell--transcript-file-path-function` to nil.

This private var/function can take care of not just creating the file name but the entire path to write to.

agent-shell.el
```diff
@@ -91,6 +91,32 @@ Example for devcontainer:
...
+  "Whether to automatically save transcripts of agent conversations.
+
+When non-nil, transcripts are saved live as the conversation progresses
+to the directory specified by `agent-shell-transcript-directory'.
+
+This is separate from shell-maker's manual save functionality."
+  :type 'boolean
+  :group 'agent-shell)
+
+(defcustom agent-shell-transcript-directory ".agents/transcripts"
```
xenodium: For now let's inline this in the `agent-shell--transcript-file-path-function`. Let's make it `$project-cwd/.agent-shell/transcripts` so it's obvious to the user who owns these files.

agent-shell.el
```diff
@@ -234,8 +259,51 @@ AUTHENTICATE-REQUEST-MAKER."
...
+  "Initialize a new transcript file for this buffer.
+Returns the path to the transcript file."
+  (when agent-shell-auto-save-transcript
+    (condition-case err
+        (let* ((dir (agent-shell--transcript-directory))
+               (filename (funcall agent-shell-transcript-filename-function))
+               (filepath (expand-file-name filename dir)))
+          (make-directory dir t)
+          (write-region
+           (format "Agent Shell Transcript\nStarted: %s\nWorking Directory: %s\n\n%s\n\n"
```
xenodium: We should consider including the agent name (ie. Claude Code, Gemini, etc)

agent-shell.el
```diff
@@ -234,8 +259,51 @@ AUTHENTICATE-REQUEST-MAKER."
...
+                   (format-time-string "%Y-%m-%d %H:%M:%S")
+                   (agent-shell-cwd)
+                   (make-string 60 ?=))
+           nil filepath)
+          filepath)
+      (error
+       (message "Failed to initialize transcript: %S" err)
+       nil))))
+
+(defun agent-shell--append-to-transcript (text &optional buffer)
```
xenodium: Let's swap buffer for file-path to make things more explicit.

cl-defun agent-shell--append-transcript (&key text file-path)

agent-shell.el
```diff
@@ -520,6 +592,25 @@ Flow:
                                                "│ changes │\n"
                                                "╰─────────╯\n\n" diff-text)
                                      output)))
+                   ;; Log tool call to transcript when completed or failed
```
xenodium, line 628: Let's add a helper we can use here... Maybe `agent-shell--make-transcript-tool-call-entry`, so it looks something like:

(agent-shell--append-to-transcript
  :file-path ...
  :text (agent-shell--make-transcript-tool-call-entry ...)

agent-shell.el
```diff
@@ -234,8 +259,51 @@ AUTHENTICATE-REQUEST-MAKER."
 (defvar-local agent-shell--state
     (agent-shell--make-state))
 
+(defvar-local agent-shell--transcript-file nil
+  "Path to the transcript file for this buffer.")
+
 (defvar agent-shell--shell-maker-config nil)
 
+(defun agent-shell--transcript-directory ()
```
xenodium: Let's group all the transcript-related things under a ;;; Transcript section.

xenodium (commented):
Thanks for the PR. Here are some initial comments to get things going...

xenodium:
> Example transcript format

Not for this PR, but we should consider iterating on log format later on. For example, we can maybe consider saving as .md. An upside is that opening in Emacs itself automatically gives us the ability to collapse and expand bits.

agent-shell.el
xenodium: > For now if anyone would like to disable things, set agent-shell--transcript-file-path-function to nil.

On second thought, can we make this nil by default and create an internal function we can set it to for the time being? This gives us the opportunity to override ourselves and use it for a bit of time before we roll it out to others.

ElleNajt: Makes sense to me!

ElleNajt:
Thanks for the feedback. Made the changes. Also, since it was a quick change, swapped from .txt to .md. Agreed that it makes sense to iterate on the log format.

agent-shell.el
```diff
@@ -234,8 +233,73 @@ AUTHENTICATE-REQUEST-MAKER."
...
+;;; Transcript
+
+(defvar-local agent-shell--transcript-file nil
+  "Path to the transcript file for this buffer.")
+
+(defvar agent-shell--transcript-file-path-function nil
+  "Function to generate the full transcript file path.
+Called with no arguments, should return a string path or nil to disable.
+When nil, transcript saving is disabled.
+Set this to an internal function for testing before rolling out to users.")
```
xenodium: We can remove this: "Set this to an internal function for testing before rolling out to users."

agent-shell.el
```diff
@@ -234,8 +233,73 @@ AUTHENTICATE-REQUEST-MAKER."
...
+         (filename (format-time-string "%Y-%m-%d-%H-%M-%S-transcript.md"))
+         (filepath (expand-file-name filename dir)))
+    filepath))
+
+(defun agent-shell--init-transcript (config)
+  "Initialize a new transcript file for this buffer using CONFIG.
+Returns the path to the transcript file, or nil if disabled."
+  (when agent-shell--transcript-file-path-function
+    (condition-case err
+        (let* ((filepath (funcall agent-shell--transcript-file-path-function))
```
xenodium: optionally, we can flatten by using when-let*

agent-shell.el
```diff
@@ -234,8 +233,73 @@ AUTHENTICATE-REQUEST-MAKER."
...
+               (dir (file-name-directory filepath))
+               (agent-name (or (map-elt config :mode-line-name)
+                               (map-elt config :buffer-name)
+                               "Unknown Agent")))
+          (make-directory dir t)
+          (write-region
+           (format "# Agent Shell Transcript\n\n**Agent:** %s  \n**Started:** %s  \n**Working Directory:** %s\n\n---\n\n"
```
xenodium: We can use multi-line strings (easier on the reader).

agent-shell.el
```diff
@@ -234,8 +233,73 @@ AUTHENTICATE-REQUEST-MAKER."
...
+                   (agent-shell-cwd))
+           nil filepath)
+          filepath)
+      (error
+       (message "Failed to initialize transcript: %S" err)
+       nil))))
+
+(cl-defun agent-shell--append-transcript (&key text file-path)
+  "Append TEXT to the transcript at FILE-PATH.
+Both TEXT and FILE-PATH are required."
```
xenodium: We can likely remove this from docstring. There's no failure if they are not included. Alternatively use unless/error to keep.

agent-shell.el
```diff
@@ -234,8 +233,73 @@ AUTHENTICATE-REQUEST-MAKER."
...
+  (when (and file-path (file-exists-p file-path))
+    (condition-case err
+        (write-region text nil file-path t 'no-message)
+      (error
+       (message "Error writing to transcript: %S" err)))))
+
+(cl-defun agent-shell--make-transcript-tool-call-entry (&key status title kind description command output)
+  "Create a formatted transcript entry for a tool call.
+Returns a formatted string with tool call details."
+  (format "<details>\n<summary>🔧 Tool Call [%s]: %s (%s)</summary>\n\n**Tool:** %s%s  \n**Timestamp:** %s%s\n\n**Output:**\n```\n%s\n```\n\n</details>\n\n"
```
xenodium: Let's use multiline here too please.

agent-shell.el
```diff
@@ -470,8 +534,16 @@ Flow:
                (map-put! state :last-entry-type "agent_thought_chunk"))
               ((equal (map-elt update 'sessionUpdate) "agent_message_chunk")
                (unless (equal (map-elt state :last-entry-type) "agent_message_chunk")
-                 (map-put! state :chunked-group-count (1+ (map-elt state :chunked-group-count))))
+                 (map-put! state :chunked-group-count (1+ (map-elt state :chunked-group-count)))
+                 (agent-shell--append-transcript
+                  :text (format "## 🤖 Agent (%s)\n\n" (format-time-string "%Y-%m-%d %H:%M:%S"))
```
xenodium: Can we remove the emoji please? While they can be fun, I can understand how folks are starting to get saturated by LLM emoji overuse. I a future PR, we can look at maybe offering transcript template, but let's not do it just yet. Let's use the current format for a little while to get a feel for it.

agent-shell.el
```diff
@@ -470,8 +534,16 @@ Flow:
...
+                  :file-path (with-current-buffer (map-elt state :buffer)
```
xenodium: Did you need this? if so, it's probably a bug on ACP or agent-shell. agent-shell--make-acp-client should be taking :buffer such that all ACP callbacks are invoked in the shell buffer.

agent-shell.el
```diff
@@ -520,6 +592,25 @@ Flow:
...
+                            (description (map-elt tool-call :description))
+                            (status (map-elt update 'status)))
+                       (agent-shell--append-transcript
+                        :text (agent-shell--make-transcript-tool-call-entry
+                               :status status
+                               :title title
+                               :kind kind
+                               :description description
+                               :command command
+                               :output body-text)
```
xenodium: We likely don't need these local vars. I've seen LLMs being overly keen on adding local vars when inlining remains reasonable and more explicit. For example:

```
(agent-shell--append-transcript
 :text (agent-shell--make-transcript-tool-call-entry
        :status (map-elt update 'status)
        :title (map-nested-elt state `(:tool-calls ,.toolCallId :title))
        :kind (map-nested-elt state `(:tool-calls ,.toolCallId :kind))
        :description (map-nested-elt state `(:tool-calls ,.toolCallId :description))
        :command (map-nested-elt state `(:tool-calls ,.toolCallId :command))
        :output body-text))
```

agent-shell.el
```diff
@@ -1764,6 +1803,15 @@ Returns list of alists with :start, :end, and :path for each mention."
                                      (text . ,(substring-no-properties prompt)))])))
          (attached-files (agent-shell--collect-attached-files content-blocks)))
 
+    ;; Reset last-entry-type for new user prompt
```
xenodium: This comment is a bit misleading. Let's remove it.

agent-shell.el
```diff
@@ -1773,6 +1821,11 @@ Returns list of alists with :start, :end, and :path for each mention."
                :prompt content-blocks)
      :buffer (current-buffer)
      :on-success (lambda (response)
+                   (when (equal (map-elt (agent-shell--state) :last-entry-type) "agent_message_chunk")
+                     (agent-shell--append-transcript
+                      :text "\n\n"
+                      :file-path (with-current-buffer (map-elt (agent-shell--state) :buffer)
```
xenodium: Same here. Did we need with-current-buffer? ACP/agent-shell didn't use the shell buffer?

agent-shell.el
ElleNajt: It was unnecessary, yeah. Ty!

xenodium:
Hi @ElleNajt I got notification of some commits addressing comments. Thank you! Are you finished with the changes? If so, I'll look into merging.

ElleNajt:
Hi -- yes, finished with the changes! :)

xenodium (approved):
Thanks for iterating on the PR.

xenodium:
Merged!

PR #101, opened 2025-10-25
Return nil for "no header" style

rynffoll:
Previously 'no header' style returned an empty string, which will be shown as a blank line. Return nil to hide a header line.

```elisp
(setq agent-shell-header-style nil) ;; no header
```

Before:
image

After:
image


xenodium (approved):
Thank you for fixing!

rynffoll:
Thanks for the super fast review!

PR #153, opened 2025-12-16
fix: SVG header width to respect max-image-size

reynard93:
## Summary
- Add `agent-shell--safe-svg-width()` helper function that respects `max-image-size`
- Handle both numeric MB limits and cons `(width . height)` forms
- Provide fallback width of 1200px and minimum of 400px  
- Update SVG creation to use safe width instead of raw `frame-pixel-width`
- Prevents "Invalid image size" errors when frame width exceeds limits

## Problem
`agent-shell-manager-goto` was failing with "Invalid image size" errors because the SVG header used `(frame-pixel-width)` directly, which could exceed Emacs' `max-image-size` limit (default: 10MB).

## Solution  
The new `agent-shell--safe-svg-width()` function:
- Respects `max-image-size` setting (both number and cons forms)
- Calculates safe width based on estimated header height (~100px)
- Provides fallback width of 1200px if frame width unavailable
- Ensures minimum width of 400px for usability

## Testing
- ✅ With default max-image-size (10MB): 1920px frame width passes through unchanged
- ✅ With restrictive max-image-size (0.1MB): 1920px gets clamped to 1048px  
- ✅ Minimum width guarantee ensures usability even with very restrictive limits

Fixes the issue where `agent-shell-manager-goto` would fail due to SVG image size exceeding `max-image-size` limits.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1897,7 +1897,7 @@ BINDINGS is a list of alists defining key bindings to display, each with:
                   (icon-text-y text-height)
                   ;; Bindings positioned right after the bottom text (2 text lines) plus spacing
                   (bindings-y (+ (* 3 text-height) row-spacing))
-                  (svg (svg-create (frame-pixel-width) total-height))
+                   (svg (svg-create (agent-shell--safe-svg-width 1200) total-height))
                   (icon-filename
                    (if (map-nested-elt state '(:agent-config :icon-name))
                        (agent-shell--fetch-agent-icon (map-nested-elt state '(:agent-config :icon-name)))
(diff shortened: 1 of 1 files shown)


xenodium:
Thank you

xenodium:
Looking a little closer, we may be complicating things a bit...

> agent-shell-manager-goto was failing with "Invalid image size" errors because the SVG header used (frame-pixel-width) directly, which could exceed Emacs' max-image-size limit (default: 10MB).

Are we confident this is the case? Reading `max-image-size` docstring, my interpretation is that 10.0 is a ratio, not MB?

```
Maximum size of images.

Emacs will not load an image into memory if its pixel width or
pixel height exceeds this limit.

If the value is an integer, it directly specifies the maximum
image height and width, measured in pixels.  If it is a floating
point number, it specifies the maximum image height and width
as a ratio to the frame height and width.  If the value is
non-numeric, there is no explicit limit on the size of images.
```

Other things to consider:

- `fallback-width` is unnecessary. Svg's are only ever created when `(display-graphic-p)`.
- `(bound-and-true-p max-image-size)`. Is this check necessary? isn't max-image-size guaranteed granted this function is invoked when `(display-graphic-p)`.
- We're converting MBs in the calcultion, but it's unclear why we're using MBs.

We can likely simplify. In the mean time, I've reverted the merge. Sorry I jumped the gun and merged. My mistake.

Also, could you share more details of when things failed for ya? A stack trace would be helpful.

reynard93:
Will get the stack trace, sorry it's my bad! I didn't fully understand the issue

Issue #154, opened 2025-12-17
Gemini CLI doesn't need authorization if already logged in.


ccqpein:
I have logged into the Gemini CLI in my terminal, and then I started to run it in Emacs. I don't want to use my API key, so `agent-shell` tries to make me log in through `google.com` (I guess), and an error occurs.

There are two issues that happened in this workflow:

*   Web login doesn't work for me after I abandoned the API key.
*   My Gemini is actually logged in, but `agent-shell` cannot skip the authorization.

skyler544:
You might try this:

```lisp
(use-package agent-shell
  :custom (agent-shell-goose-authentication
           (agent-shell-make-goose-authentication :none t)))
```

It works for github copilot at least, but you need to do the authorization from the `goose` cli outside of Emacs first.

EDIT: of course you should use the variables for `gemini` and not `goose` when you try this 👍 

EDIT2: sorry, this comment won't apply to `gemini`. I didn't realize that the authentication interface is not uniform across providers.

xenodium:
> (agent-shell-make-goose-authentication :none t)

Good shout @skyler544.

> I didn't realize that the authentication interface is not uniform across providers.

I try to keep them uniform but it's also been fairly ad-hoc. Don't think anyone needed this feature on gemini until now :) We should indeed introduce authentication `:none` as an option for gemini also.

ccqpein:
Just made the change. And btw the first issue for some reason works to me now.

ccqpein:
PR merged.

PR #155, opened 2025-12-17
Give the new custom flag to gemini for skipping the authorization

ccqpein:
Try to solve the second issue of #154

### DIFF
--- a/agent-shell-google.el
+++ b/agent-shell-google.el
@@ -36,7 +36,7 @@
 (declare-function agent-shell-make-agent-config "agent-shell")
 (declare-function agent-shell--dwim "agent-shell")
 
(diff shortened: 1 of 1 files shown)


xenodium:
Thanks for the PR. While this works, it'd be good to converge on how we do this across providers. We can achieve a similar thing using authentication as `:none`. This isn't yet supported by the gemini setup, but we should likely go in that direction. You can take a look at either `agent-shell-goose.el` or `agent-shell-qwen.el` for how it's done and replicate for gemini, if that's ok?

ccqpein:
@xenodium just make the change for gemini (copy implementation from the qwen).

xenodium:
Thank you. Merged!

PR #166, opened 2025-12-27
Use project-name for buffer names

catern:
Buffer names now use project-name.  By default this is just the last directory component, but it's configurable by the user or by the project implementation if they want to refer to their projects with a different name.

## Checklist

- [X] *I've reviewed all code in PR myself and I'm happy with its quality*.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1571,8 +1571,7 @@ variable (see makunbound)"))
                              new-session
                              (concat (map-elt config :buffer-name)
                                      " Agent @ "
-                                     (file-name-nondirectory
-                                      (string-remove-suffix "/" default-directory)))
+                                     (agent-shell--project-name))
                              (map-elt config :mode-line-name))))
     (with-current-buffer shell-buffer
       ;; Initialize buffer-local state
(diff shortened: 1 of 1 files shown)


xenodium:
Thank you for the PR. agent-shell has both project.el users as well as projectile. Mind adding a helper like `agent-shell-cwd`, maybe name it `agent-shell--project-name` catering for both and with a `(agent-shell-cwd)` last directory component fallback?

catern:
How should it determine whether to use projectile or not?  Simply checking whether projectile is installed is insufficient - I have projectile installed (as a dependency of other packages) but I don't use it.  Do you want a defcustom to control that which defaults to project.el?

BTW, projectile already has its own compatibility code with project.el.  Every projectile project is also a project.el project.  I don't think you actually need to have compatibility code in agent-shell, but if you still want it, that's fine.

xenodium:
> How should it determine whether to use projectile or not? Simply checking whether projectile is installed is insufficient

You're right, `agent-shell-cwd` was missing an explicit check for `projectile-mode` to be t. I had this check elsewhere but not in `agent-shell-cwd`. With that check, we can actually check the user intends to use projectile over project.el, as they had to enable projectile-mode.

The check is now added in https://github.com/xenodium/agent-shell/commit/121c182fe48be1fef8c442641bcb9c042a5b9fd2

> I have projectile installed (as a dependency of other packages) but I don't use it.

With the projectile-mode check, we should have this case now covered.

> Do you want a defcustom to control that which defaults to project.el?

Shouldn't be needed after the internal projectile-mode check, or was there something else missing?

> BTW, projectile already has its own compatibility code with project.el. Every projectile project is also a project.el project. I don't think you actually need to have compatibility code in agent-shell, but if you still want it, that's fine.

Mind adding it for now until I have a chance to consolidate if needed?

catern:
Checking projectile-mode sounds good to me.  Here's a new version, tested with projectile and project.el

xenodium (approved):

xenodium:
Thank you. Merged.

PR #169, opened 2025-12-30
Resolve `@mentioned` paths in devcontainers

fritzgrabo:
Hi @xenodium!

I'm running agent-shell in a devcontainer setup and noticed that `@path` mentions are not currently resolved / mapped from the local filesystem to the devcontainer one in that scenario. In other words, a mention of `@bin/foo` would build a `resource` or `resource-link` content block that points to the file's absolute path _on the host filesystem_.

For smaller files, it seems that doesn't matter much since they're embedded right into the `resource` block anyway (if embedding is supported), and the agent just takes the file contents from there and doesn't bother re-reading from the  provided`uri`. For larger files though, it has to start guessing what file you actually meant.

Talking about embedding: I believe we should _not_ embed files unless the agent was started with `text file capabilities`? Would that not be a breach of contract? I've added a check for that, but happy to discuss if you think otherwise.

The PR also contains a small fix for the function that detects the devcontariner's working directory: in case it has to fallback to the default, it adds a trailing `/` -- I had missed that previously.

Also, fixes a couple failing tests!

I'm aware I lumped a few unrelated changes into a single PR (I just kept going, lol). If you'd rather have me split things off, let me know! Thanks!

### DIFF
--- a/tests/agent-shell-tests.el
+++ b/tests/agent-shell-tests.el
@@ -189,13 +189,12 @@
 (ert-deftest agent-shell--format-plan-test ()
   "Test `agent-shell--format-plan' function."
   ;; Test homogeneous statuses
-  (should (equal (agent-shell--format-plan [((content . "Update state initialization")
+  (should (equal (substring-no-properties (agent-shell--format-plan [((content . "Update state initialization")
                                              (status . "pending"))
                                             ((content . "Update session initialization")
-                                             (status . "pending"))])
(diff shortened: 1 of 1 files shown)


xenodium:
Happy New Year and thanks for the PR! Merged.

fritzgrabo:
@xenodium Thanks for the quick turnaround! Happy New Year to you as well! 🙌

PR #179, opened 2026-01-09
Auggie Support

Pacane:
As discussed in the Auggie support issue, I have implemented this. Here's the work in progress.

I couldn't get it to work 100% but according to others who have used it (and reported it works fine in the issue) it should work decently.

I'm having some issues with my augment subscription lately, but it was working OK before that.


xenodium:
Lovely. Thank you! Merged.

PR #196, opened 2026-01-16
fix agent names in prompt when agent-shell-show-config-icons is nil

budevg:
Previously nil was converted to string "nil", so the display was showing
"nilnilCodex" instead of "Codex"


xenodium:
Thank you!

Issue #176, opened 2026-01-03
Make `agent-shell-container-command-runner` a function


rpoisel:
Thanks for providing agent-shell. I really like it so far! 👍

I run agents in DevContainers, so I set `agent-shell-container-command-runner` to use the [devcontainer CLI](https://github.com/devcontainers/cli) (per the [README](https://github.com/xenodium/agent-shell?tab=readme-ov-file#running-agents-in-devcontainers--docker-containers-experimental)). This works when all agents share a single container.

**Use case / motivation:** I want to isolate each agent in its own container. With [Docker Compose](https://containers.dev/guide/dockerfile#docker-compose), I need to pick a different service depending on the agent.

**Suggestion:** allow `agent-shell-container-command-runner` to be either a list of strings (current behavior) or a function. If it’s a function, `agent-shell` would call it with the agent type (e.g., symbols `anthropic`, `openai`, `mistral`) and use the returned list of strings as the command prefix. If the function returns nil/empty, treat it as “no container runner.”

Before implementing, I’d love your (@xenodium and others) opinion.

Thanks in advance,  
Rainer

xenodium:
Thanks for the proposal.

I've yet to use the containers. cc'ing @fritzgrabo who's been driving the container work in `agent-shell`.

> Suggestion: allow agent-shell-container-command-runner to be either a list of strings (current behavior) or a function. If it’s a function, agent-shell would call it with the agent type (e.g., symbols anthropic, openai, mistral) and use the returned list of strings as the command prefix. If the function returns nil/empty, treat it as “no container runner.”

Allowing functions as a `agent-shell-container-command-runner` value SGTM. I'd suggest a slight change and pass the shell buffer (instead of agent type). This allows clients to draw whichever info they need from the buffer itself, beyond agent type.

@fritzgrabo WDYT?

fritzgrabo:
@xenodium @rpoisel I really like the proposal: it allows for a more flexible configuration, it is backward compatible with what we have now, and the "string-or-function" approach is an established pattern in Elisp. I'd love to see this implemented and am happy to help if I can.

[...]

PR #197, opened 2026-01-16
feat: support function-based container command runner configuration

rpoisel:
This PR adds support for dynamic, per-agent container configuration by allowing `agent-shell-container-command-runner` to accept either a static command list or a function that takes the agent-shell buffer and returns a command list.

This enables running different agents in different containers based on the agent's `:identifier` field, while maintaining backward compatibility with the existing static list configuration.

## Changes

  - Allow `agent-shell-container-command-runner` to be a function
  - Add `:identifier` field to all agent configs
  - Extract command wrapping logic into `agent-shell--wrap-command-for-container`
  - Add tests for this configuration setting
  - Update documentation with examples

The commit history reflects the chronological development sequence; I recommend squashing when merging, but I'm flexible if you prefer a different approach.

## Checklist

- [x] I've filed a feature request/discussion for this change: #176 
- [x] *I've reviewed all code in PR myself and I'm happy with its quality*.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -127,11 +127,26 @@ returns the resolved path.  Set to nil to disable mapping."
   "Command prefix for executing commands in a container.
 
 When non-nil, both the agent command and shell commands will be
-executed using this runner.  Should be a list of command arguments.
+executed using this runner. Can be a list of strings or a function
+that takes a buffer and returns a list.
 
-Example for devcontainer:
-  \\='(\"devcontainer\" \"exec\" \"--workspace-folder\" \".\")"
-  :type '(repeat string)
+Example for static devcontainer:
+  \\='(\"devcontainer\" \"exec\" \"--workspace-folder\" \".\")
+
+Example for dynamic per-agent containers:
+  (lambda (buffer)
+    (let ((config (agent-shell-get-config buffer)))
+      (pcase (map-elt config :identifier)
+        (\\='claude-code \\='(\"docker\" \"exec\" \"claude-dev\" \"--\"))
+        (\\='gemini-cli \\='(\"docker\" \"exec\" \"gemini-dev\" \"--\"))
+        (_ \\='(\"devcontainer\" \"exec\" \".\")))))
+
+Example for per-session containers:
+  (lambda (buffer)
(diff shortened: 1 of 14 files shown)


agent-shell.el
```diff
@@ -119,11 +119,19 @@ returns the resolved path.  Set to nil to disable mapping."
...
-  \\='(\"devcontainer\" \"exec\" \"--workspace-folder\" \".\")"
-  :type '(repeat string)
+executed using this runner. Can be a list of strings or a function
+that takes config and returns a list.
+
+Example for static devcontainer:
+  \\='(\"devcontainer\" \"exec\" \"--workspace-folder\" \".\")
+
+Example for dynamic per-agent containers:
+  (lambda (config)
```
xenodium: Passing config works and happy to merge as is. One thing to consider... passing the buffer offers a little more flexibility. For example, using a different container per shell session (even if they are of the same agent type identifier).

rpoisel:
> Left an optional comment, in case you'd like to tweak the param.

Thanks for your approval and comment!

Having different containers per shell session sounds great. I’ll think this through and check whether it can be implemented easily.

> @fritzgrabo may also like to take a look at this PR, since he’s more familiar with the container integration and actually using it ;)

Fritz (a friend of mine) has already kindly reviewed my changes twice. :-) I’ll implement his latest suggestions in one more commit for this PR.

rpoisel:
> Passing config works and happy to merge as is. One thing to consider... passing the buffer offers a little more flexibility. For example, using a different container per shell session (even if they are of the same agent type identifier).

@xenodium and @fritzgrabo, I just changed the implementation to pass around the agent-shell buffer instead of the associated config.

In order to access the config of aforementioned buffer, I introduced the `agent-shell-get-config` function. It's a oneliner, but imo still better than accessing the internal `agent-shell--state` directly. Let me know if you would like to change that.

I'd suggest to merge this PR. In my view it should support any use-case that might come up in the near future. :smile: 

I’m planning to implement a reference container setup and write a blog post about it soon.

xenodium (approved):

xenodium:

fritzgrabo:
Awesome, thanks all! @rpoisel, looking forward to the blog post you mentioned :)

PR #205, opened 2026-01-18
Add TRAMP support for remote agent execution

### LINKED ISSUE #122: Add tramp support
If you open agent-shell in a remote project (accessed via tramp), it seems to run the agent (EG opencode, goose etc) on the local machine. The agent therefore can't see any of the files in the project. Ideally the agent would be run on the remote machine in the working directory of the project. 

I don't know much about it, but for example eglot will run the appropriate lsp server on the remote machine for you, so it all works seamlessly. Perhaps agent-shell could do the same thing?
Thanks

csheaff:
Fixes #122

I've tested this with Claude Code and Copilot CLI. Unfortunately I realized in the process that I need persistent sessions with remote, as I often times lose connection. So SSH isn't ideal. But perhaps this PR will be useful for others with stable connections.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1175,6 +1175,72 @@ If the buffer's file has changed, prompt the user to reload it."
   "Resolve PATH using `agent-shell-path-resolver-function'."
   (funcall (or agent-shell-path-resolver-function #'identity) path))
 
+;;; TRAMP Support (experimental)
+
+(declare-function tramp-tramp-file-p "tramp")
+(declare-function tramp-dissect-file-name "tramp")
+(declare-function tramp-file-name-method "tramp")
+(declare-function tramp-file-name-user "tramp")
+(declare-function tramp-file-name-host "tramp")
+(declare-function tramp-file-name-port "tramp")
+(declare-function tramp-file-name-localname "tramp")
+(declare-function tramp-make-tramp-file-name "tramp")
+
+(defun agent-shell--tramp-command-runner (buffer)
+  "Return command prefix for running commands on TRAMP remote host.
+BUFFER is the agent-shell buffer.
+Returns nil for non-TRAMP buffers, allowing local execution."
+  (require 'tramp)
+  (with-current-buffer buffer
+    (let ((cwd (agent-shell-cwd)))
+      (when (tramp-tramp-file-p cwd)
+        (let* ((vec (tramp-dissect-file-name cwd))
+               (method (tramp-file-name-method vec))
+               (user (tramp-file-name-user vec))
+               (host (tramp-file-name-host vec))
+               (port (tramp-file-name-port vec)))
+          (unless (member method '("ssh" "scp" nil))
+            (error "TRAMP method '%s' not supported; only SSH is supported" method))
+          (append
+           (list "ssh")
+           (when port (list "-p" port))
+           (list (if user (format "%s@%s" user host) host))
+           (list "--")))))))
+
+(defun agent-shell--resolve-tramp-path (path)
+  "Resolve PATH between TRAMP format and remote-local format.
+
+For example:
+- /ssh:host:/project/README.md => /project/README.md
+- /project/README.md => /ssh:host:/project/README.md"
+  (require 'tramp)
+  (let* ((cwd (agent-shell-cwd))
+         (tramp-vec (and (tramp-tramp-file-p cwd)
+                         (tramp-dissect-file-name cwd))))
+    (cond
+     ;; Path is already a TRAMP path - strip the prefix for the agent
+     ((tramp-tramp-file-p path)
+      (tramp-file-name-localname (tramp-dissect-file-name path)))
+     ;; Path is a remote-local path - add TRAMP prefix for Emacs
+     (tramp-vec
+      (tramp-make-tramp-file-name tramp-vec path))
+     ;; Not in a TRAMP context
+     (t path))))
+
+(defun agent-shell-enable-tramp-support ()
+  "Enable TRAMP support for agent-shell (experimental)."
+  (interactive)
+  (setq agent-shell-container-command-runner #'agent-shell--tramp-command-runner)
+  (setq agent-shell-path-resolver-function #'agent-shell--resolve-tramp-path)
+  (message "TRAMP support enabled for agent-shell"))
+
+(defun agent-shell-disable-tramp-support ()
+  "Disable TRAMP support for agent-shell."
+  (interactive)
+  (setq agent-shell-container-command-runner nil)
+  (setq agent-shell-path-resolver-function nil)
+  (message "TRAMP support disabled for agent-shell"))
+
 (defun agent-shell--get-devcontainer-workspace-path (cwd)
   "Return devcontainer workspaceFolder for CWD, or default value if none found.
 
@@ -2190,7 +2256,12 @@ Icon names starting with https:// are downloaded directly from that location."
                            url))
                        ;; For lobe-icons names, use the original filename
                        (file-name-nondirectory url)))
-           (cache-dir (file-name-concat (temporary-file-directory) "agent-shell" mode))
+           ;; Always use local temp directory, even when default-directory is remote
+           (local-temp-dir (if (and (fboundp 'tramp-tramp-file-p)
+                                    (tramp-tramp-file-p default-directory))
+                               (or (getenv "TMPDIR") "/tmp")
+                             (temporary-file-directory)))
+           (cache-dir (file-name-concat local-temp-dir "agent-shell" mode))
            (cache-path (expand-file-name filename cache-dir)))
       (unless (file-exists-p cache-path)
         (make-directory cache-dir t)
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -4229,8 +4229,22 @@ When nil, transcript saving is disabled."
 
 For example:
 
- project/.agent-shell/transcripts/."
-  (let* ((dir (expand-file-name ".agent-shell/transcripts" (agent-shell-cwd)))
+ project/.agent-shell/transcripts/.
+
+For TRAMP paths, saves locally in ~/.agent-shell/transcripts/<host>/<remote-path>/."
+  (let* ((cwd (agent-shell-cwd))
+         (dir (if (and (fboundp 'tramp-tramp-file-p)
+                       (tramp-tramp-file-p cwd))
+                  ;; For TRAMP paths, save transcripts locally
+                  (let* ((vec (tramp-dissect-file-name cwd))
+                         (host (tramp-file-name-host vec))
+                         (localname (tramp-file-name-localname vec))
+                         ;; Sanitize path for use as directory name
+                         (safe-path (replace-regexp-in-string "/" "_" (string-trim localname "/"))))
+                    (expand-file-name (format ".agent-shell/transcripts/%s/%s" host safe-path)
+                                      (expand-file-name "~")))
+                ;; Local paths use project root as before
+                (expand-file-name ".agent-shell/transcripts" cwd)))
          (filename (format-time-string "%F-%H-%M-%S.md"))
          (filepath (expand-file-name filename dir)))
     filepath))
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1175,74 +1175,12 @@ If the buffer's file has changed, prompt the user to reload it."
   "Resolve PATH using `agent-shell-path-resolver-function'."
   (funcall (or agent-shell-path-resolver-function #'identity) path))
 
-;;; TRAMP Support (experimental)
-
-(declare-function tramp-tramp-file-p "tramp")
-(declare-function tramp-dissect-file-name "tramp")
-(declare-function tramp-file-name-method "tramp")
-(declare-function tramp-file-name-user "tramp")
-(declare-function tramp-file-name-host "tramp")
-(declare-function tramp-file-name-port "tramp")
-(declare-function tramp-file-name-hop "tramp")
-(declare-function tramp-file-name-localname "tramp")
-(declare-function tramp-make-tramp-file-name "tramp")
-
-(defun agent-shell--tramp-command-runner (buffer)
-  "Return command prefix for running commands on TRAMP remote host.
-BUFFER is the agent-shell buffer.
-Returns nil for non-TRAMP buffers, allowing local execution."
-  (require 'tramp)
-  (with-current-buffer buffer
-    (let ((cwd (agent-shell-cwd)))
-      (when (tramp-tramp-file-p cwd)
-        (let* ((vec (tramp-dissect-file-name cwd))
-               (method (tramp-file-name-method vec))
-               (user (tramp-file-name-user vec))
-               (host (tramp-file-name-host vec))
-               (port (tramp-file-name-port vec)))
-          (unless (member method '("ssh" "scp" nil))
-            (error "TRAMP method '%s' not supported; only SSH is supported" method))
-          (when (tramp-file-name-hop vec)
-            (error "Multi-hop TRAMP paths not supported"))
-          (append
-           (list "ssh")
-           (when port (list "-p" port))
-           (list (if user (format "%s@%s" user host) host))
-           (list "--")))))))
-
-(defun agent-shell--resolve-tramp-path (path)
-  "Resolve PATH between TRAMP format and remote-local format.
-
-For example:
-- /ssh:host:/project/README.md => /project/README.md
-- /project/README.md => /ssh:host:/project/README.md"
-  (require 'tramp)
-  (let* ((cwd (agent-shell-cwd))
-         (tramp-vec (and (tramp-tramp-file-p cwd)
-                         (tramp-dissect-file-name cwd))))
-    (cond
-     ;; Path is already a TRAMP path - strip the prefix for the agent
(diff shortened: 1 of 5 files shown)


xenodium:
Thank you for kicking this off!

I need to start breaking the `agent-shell.el` sections into sub packages. Mind if we move the TRAMP functionality to a new file, say `agent-shell-tramp.el`?

ps. I'm not using TRAMP myself, so may be good to ping the feature request at https://github.com/xenodium/agent-shell/issues/122 and ask for feedback or folks to try it out.

fritzgrabo:
This is awesome -- love how support for containers and now TRAMP make `agent-shell` such a flexible tool for use with agents.

@csheaff re. the unstable connection issues you mention in the issue, maybe have a look at [detached.el](https://sr.ht/~niklaseklund/detached.el), and more importantly, the [dtach](https://github.com/crigler/dtach) tool it uses under the hood. It certainly adds another layer of indirection, but would help the session survive a reconnect.

Re. the PR, I was wondering if we could make TRAMP transparent, in the sense that it prefixes the command to run and resolves paths to/from remote machines automatically? After all, you should be able to defer from the current `cwd` and file path whether it's using TRAMP or not, so could bake that into the default logic, without the need to use `agent-shell-container-command-runner` or `agent-shell-path-resolver-function` at all.

**UPDATE:** just realized that not everybody would want to run the agent on the remote machine. which is likely why you went with the custom functions to manually toggle the feature -- got it 💡. The PR's issue mentions how eglot runs the server on the remote machine transparently, and if I [read the code right](https://github.com/emacs-mirror/emacs/blob/master/lisp/progmodes/eglot.el#L1788), that happens automatically for remote working directories (at least I didn't find anything in `eglot.el` that would prefix the command to run with `ssh ...`). Should the same behavior should be the default in `agent-shell` then, and opting out of it should be toggled by a flag?

What do you think?

csheaff:
Hey @fritzgrabo , thanks for the feedback. In addition to removing the disable/enable functionality, I've attempted to use `make-process` as Eglot does. However agent-shell stalls for some reason and I need to move on to other work for rest of the week. @xenodium This requires some updates to acp.el i think:

- adding :file-handler t to the make-process call in acp.el (line 130)
- addressing the fact that TRAMP's file handler isn't compatible with the stderr handling in acp.el  ( eglot.el line ~1794: uses `(get-buffer-create ...)` for stderr, not a pipe process)

I will attempt to fix next weekend if i can find some time.

junyi-hou:
I tried it a little bit and here's what I found:
- the `ssh HOST -- claude-code-acp` trick does not work if `claude-code-acp` is install in a customized place because this does not source the profile/rc file on the remote, changing the prefix to `ssh HOST -- zsh -lc` (or bash lc) fix this. i.e.
```elisp
(defun agent-shell--tramp-command-runner (buffer)
  "Return command prefix for running commands on TRAMP remote host.
BUFFER is the agent-shell buffer.
Returns nil for non-TRAMP buffers, allowing local execution."
  (require 'tramp)
  (with-current-buffer buffer
    (let ((cwd (agent-shell-cwd)))
      (when (tramp-tramp-file-p cwd)
        (let* ((vec (tramp-dissect-file-name cwd))
               (method (tramp-file-name-method vec))
               (user (tramp-file-name-user vec))
               (host (tramp-file-name-host vec))
               (port (tramp-file-name-port vec)))
          (unless (member method '("ssh" "scp" nil))
            (error "TRAMP method '%s' not supported; only SSH is supported" method))
          (when (tramp-file-name-hop vec)
            (error "Multi-hop TRAMP paths not supported"))
          (append
           (list "ssh")
           (when port (list "-p" port))
           (list (if user (format "%s@%s" user host) host))
           (list "--" "zsh" "-lc")))))))
```

- the environment variable set in `agent-shell-anthropic-claude-environment` doesn't seem to get pass onto the remote process. I was trying to add `(when agent-shell-anthropic-claude-environment agent-shell-anthropic-claude-environment)` to the list but this doesn't work with complex env variable (like `ANTHROPIC_CUSTOM_HEADERS=OpenAI-Organization: xxx`)

junyi-hou:
environment variables can be passed in the following way:
```
(defun agent-shell-tramp-pass-env-var (var)
  "VAR should be output from `agent-shell-make-environment-variables`"
  (mapcar (lambda (env)
            (if-let* ((env-list (string-split env "="))
                      (value (cadr env-list))
                      (_ (string-match-p " " value)))
                (string-join `(,(car env-list) ,(format "'%s'" value)) "=")
              env))
          var))
```
to add quotes to the env variables

neonmei:
Hi! first of all, thank you a lot for contributing on this 🙌

I took the branch for a spin to see how it works with `gemini` and `qwen`. Just as general context I do have some odd paths because I use nixos, but that generally works out with some adjusting of `tramp-remote-path`.

The steps I performed:
1. I installed the `gh-122-add-tramp-support` branch in Doom Emacs and ensured that `agent-shell-enable-tramp-support` ran.
2. opened a golang project on one of my local `microvm.nix` VMs (`projectile-switch-project` -> `/ssh:ouroboros.vm:~/kaifa/go/yinglong`). These generally work fast and functionality like LSPs, dape, etc run pretty well.
3. picked a file and ran `agent-shell`
4. said a "hello!" in the chat

~for `gemini`, it seems stuck in Initializing but don't understand why~

Edit:  just cleaned up some caches and gemini seems to start! but it does exhibit same behavior as qwen for linked files.

As for `qwen`, it managed to start the shell with `C-c a a` and it answered the "hello", but current opened file gets linked as `docs/docs/control-plane.org` instead of `docs/control-plane.org`. Perhaps there's some quirk in path resolution for current opened buffer inside a remote project? (projectile, project.el, etc). 

If anything needs some additional test, I'm happy to help!

fritzgrabo:
Great call on the ENV vars, much agree that it's important to be able to set those on the remote host.

TRAMP seems to have solved a lot of the issues around remote process execution already (including passing ENV vars and finding executables, see `tramp-remote-process-environment` and `tramp-remote-path`), so I believe that the more TRAMP functionality we can reuse here, the better: it'll keep the `agent-shell` code base simple(r) and allows users to fine-tune to their requirements using TRAMP built-in functionality.

csheaff:
Thanks everyone, I've pushed an update that changes the approach significantly.

**What changed:**
- Now using TRAMP's native `:file-handler` mechanism (like eglot does) instead of the SSH prefix approach
- This required changes to acp.el - see PR xenodium/acp.el#9
- Removed `agent-shell-enable/disable-tramp-support` commands - TRAMP now works automatically when `default-directory` is a TRAMP path

**@junyi-hou** - The `:file-handler` approach should address both your concerns:
1. TRAMP's file-handler respects `tramp-remote-path`, so custom install locations should work
2. Environment variables are passed via `tramp-remote-process-environment`

**@fritzgrabo** - Agreed on using TRAMP built-ins. The file-handler approach inherently uses `tramp-remote-process-environment` and `tramp-remote-path`.

**@neonmei** - I wasn't able to reproduce the `docs/docs/...` path duplication. Could you try the latest push and let me know if you still see it? If so, what does `(agent-shell-cwd)` return when you're in the buffer? And what's the full TRAMP path of the file you have open?

csheaff:
Edit: I'm getting the stall again at 

> ▶  in progress  Starting agent
> 
>   Creating client...
> 
>   Subscribing...
> 
>   Initializing...

not sure what's going on here...

csheaff:
Okay I think I fixed the stall issue. Seems it was a race condition - the SSH connection wasn't fully established before the first message was sent. The fix is in https://github.com/xenodium/acp.el/pull/9

agent-shell.el
```diff
@@ -1175,6 +1181,18 @@ If the buffer's file has changed, prompt the user to reload it."
...
+;; TRAMP support is in agent-shell-tramp.el
+(declare-function agent-shell--tramp-transcript-dir "agent-shell-tramp")
+
+(defun agent-shell--local-temp-directory ()
+  "Return a local temporary directory, even when `default-directory' is remote.
+This ensures temp files (like cached icons) are always stored locally."
+  (if (and (fboundp 'tramp-tramp-file-p)
+           (tramp-tramp-file-p default-directory))
+      ;; When in a TRAMP buffer, use Emacs-standard cache directory
+      (locate-user-emacs-file "agent-shell/cache/")
```
fritzgrabo, line 1193: Should this be just `(locate-user-emacs-file "cache/")`?

Further down in lines 2213 and 2240, you're `file-name-concat`-ing another `"agent-shell"`, so would end up with `".../agent-shell/cache/agent-shell"`.

agent-shell.el
```diff
@@ -1175,6 +1181,18 @@ If the buffer's file has changed, prompt the user to reload it."
   "Resolve PATH using `agent-shell-path-resolver-function'."
   (funcall (or agent-shell-path-resolver-function #'identity) path))
 
+;; TRAMP support is in agent-shell-tramp.el
+(declare-function agent-shell--tramp-transcript-dir "agent-shell-tramp")
+
+(defun agent-shell--local-temp-directory ()
```
fritzgrabo, line 1187: Nit: you might want to consider naming this `agent-shell--tramp-temp-dir` and moving it into `agent-shell-tramp.el`. It's quite similar in spirit to `agent-shell--tramp-transcript-dir` imho.

agent-shell.el
```diff
@@ -115,11 +115,14 @@ When non-nil, user message sections are expanded."
   :type 'boolean
   :group 'agent-shell)
 
-(defcustom agent-shell-path-resolver-function nil
+(autoload 'agent-shell--resolve-tramp-path "agent-shell-tramp")
+
+(defcustom agent-shell-path-resolver-function #'agent-shell--resolve-tramp-path
```
fritzgrabo, line 120: Just a thought: since `agent-shell--resolve-tramp-path` does nothing for non-tramp paths (so is save to call in all cases), you might want to hardcode calling it into `agent-shell--resolve-path`, probably even before `funcall`-ing `agent-shell-path-resolver-function`. That way, TRAMP support would be entirely transparent.

fritzgrabo (commented):
This looks great to me, thanks for your updates!

Left a couple thoughts and nits in time for the weekend (sorry 😬).

CeleritasCelery:
I just tested this with the acp.el updates, and it worked for me with claude code. Hope this can get merged.

JuanPabloMF:
What is the status of this functionality? It's a game changer for me

xenodium:
I'm not using agents on other machines so been waiting for some other feedback on the pull request. Seems like it's working for folks. Can you try it out and report back? I should be able to merge this week, but please report back before I merge. You'll need the acp.el patch also.

csheaff:
Sorry everyone, been meaning to make requested updates but it's been hard to find the time.

fritzgrabo:
@csheaff I'll gladly make the requested / suggested changes myself in a follow-up PR, don't worry about them. Happy to see your PR get merged, it seems to have a bunch of folks eagerly waiting for it already, which is awesome! 💛

JuanPabloMF:
I tried it with my local SBC and it worked, but I had to do the following minor changes:

```
Two changes were made:                                                                                                                                                                                           
                                         
  1. straight/repos/agent-shell/agent-shell-diff.el — Added (require 'diff) and (require 'map) to fix native-compiler warnings about diff-no-select and map-elt not being known to be defined.                     
  2. emacs.org — Had to change a few things from my conf, there were no problems before using the PR:                                                                                                                                                               
    - agent-shell-default-agent-config → agent-shell-preferred-agent-config (correct variable name)                                                                                                                
    - agent-shell-get-agent-config-by-name "claude-code" → agent-shell-anthropic-make-claude-code-config (correct function that returns a full config alist instead of just a symbol, which was causing the
  cl-no-applicable-method: No applicable method: map-elt, claude-code, :shell-prompt error)
```

<img width="1925" height="630" alt="Screenshot from 2026-02-17 23-03-10" src="https://github.com/user-attachments/assets/d87f2618-cef1-4570-9dd3-5602c92d406d" />

xenodium:
Hi folks. Been thinking this more thoroughly... While I like the idea of agent-shell potentially having TRAMP support, I have some reservations about merging #205 and https://github.com/xenodium/acp.el/pull/9.

In a nutshell, I know fairly little (I'm being optimistic here) about TRAMP internals/infrastructure and seldom use TRAMP myself. I don't have a coding agent running on a separate machine nor have had a need for it just yet. My worry here is that TRAMP support will invite bug reports and I won't be in a good position to maintain the code nor provide prompt support. With that in mind, I feel uncomfortable taking ownership of this feature.

I can offer two paths forward:

1. [My preference] I add extension points to both agent-shell and acp.el so TRAMP support can be injected somehow. This would enable adding TRAMP support via a separate Emacs package (agent-shell-tramp?). We have [precedence for this](https://github.com/xenodium/agent-shell/?tab=readme-ov-file#related-projects). I can possibly volunteer to port/rebase the PRs to use the extension points in a separate Emacs package, but someone else would have to take ownership and maintenance of this new package.
2. Have at least 2 folks who commit to maintaining this area of the agent-shell/acp.el codebase that would maintain and support it. That would mean being fairly responsive to reported issues (I would forward those to you). I understand this may not be feasible for some.

JuanPabloMF:
I can volunteer if needed.

dangom:
Would love to see this feature merged as well. @xenodium, given TRAMP is a killer emacs app, it would be nice to be able to have it work out of the box without an extra package. Unsolicited 2 cents, one could also add a note at the readme.md that the TRAMP support is experimental/alpha with 0 guarantees and subject to breakage. 
From my side happy to contribute with testing as needed.

jasonriedy:
TRAMP setups can involve incredibly ornate and brittle configurations. There *would* be such bug reports, and they very well may prove impossible to debug (e.g. in multi-hop TRAMP setups). ACP *wants* to support remote agents, but the docs say it may not all be there right now...

One possible path forward could follow gnus's model for IMAP. In gnus, you can provide an arbitrary command (or function) that fires up an IMAP server that communicates over a TTY. In this case, you could just allow an arbitrary function that fires up a process rather than various auth, hostname, etc. parameters. Talk to it via jsonrpc-process-connection. It's a clear delineation of responsibilities. (Sorry, I can't delved into acp.el enough to see if that already is how you're handling different providers.)

That's a hook not only for TRAMP but also any other bizarre situations, even intercepting and re-routing agent communication that might help with embedded systems.

AlecVercruysse:
I'm very grateful for the awesome package, as well as this PR for TRAMP support--thank you all. I've been using the PR successfully for a few days. 

I wonder about the utility of saving transcripts locally. AFAIK, this means sessions do not have access to previous transcripts, which seems like a big limitation, especially since sessions are fickle (i.e., do not survive tramp connection cleanups). Might it be reasonable to write transcripts to the remote project directory? If TRAMP writes add too much latency, could they be done in a deferred or async manner?

Thanks again.

agent-shell.el
thinckleyTT, line 1707: This needs to add the remote argument:

 `(executable-find command t)`

junyi-hou:
@xenodium I understand your hesitation in merging this, but was wondering about the other option you mentioned earlier - i.e., creating extension points in `acp.el` and `agent-shell` allowing tramp to take over. Let me know if you still open to this!

CeleritasCelery:
Also the changes in acp.el are not tramp specific and just change the method to one that supports tramp. So that will not require extra maintenance, and will make it easier for extension packages to modify agent-shell if tramp support is not wanted.

junyi-hou:
Hi all - I have created [a package](https://github.com/junyi-hou/agent-shell-tramp/tree/main) that adds TRAMP support via one of the earlier approach mentioned in this PR - I have briefly tested it and it worked for me. It can be a temporary(?) solution until we have the extension points ready (mostly merging the acp change really). Feel free to try it out!

csheaff:
Quick update: I moved the TRAMP-specific integration out of this core `agent-shell` PR and into a small external package:

https://github.com/csheaff/agent-shell-tramp-rpc

The package uses `agent-shell`'s existing resolver/transcript extension points and is primarily tested with [`tramp-rpc`](https://github.com/ArthurHeymans/emacs-tramp-rpc). The remaining upstream dependency is generic ACP file-handler process support, now proposed separately in xenodium/acp.el#21.

For users who want an approach that does not require patched `acp.el`, Junyi Hou also published `agent-shell-tramp`: https://github.com/junyi-hou/agent-shell-tramp

Given the maintenance concerns around owning TRAMP behavior in core, I think this PR can be considered superseded by the external package + smaller `acp.el` PR path.

csheaff:
Update: acp.el now has upstream TRAMP/file-handler process support via xenodium/acp.el#20. I updated agent-shell-tramp-rpc so it no longer requires my patched acp.el branch: https://github.com/csheaff/agent-shell-tramp-rpc/commit/ccb0133. That makes this old PR even more obsolete; the external-package route can now sit on top of upstream acp.el.

jasonriedy:
Thank you! I finally managed to make tramp-rpc work (via agent-shell atop pi).

The next "level" may be determining which tools only are available on the far side... The agent-shell author/maintainer is correct to delegate this to users, imho. There are too many variables.

xenodium:
> That makes this old PR even more obsolete; the external-package route can now sit on top of upstream acp.el.

Perfect. Will go ahead and close. Please ping if there's something else needed from my end.

By the way, ok to link to your tramp package in the README?

Issue #207, opened 2026-01-19
[Bug] Some colors do not update after changing the Emacs theme (e.g., modus-themes-toggle)


whhone:
After changing the light/dark theme (e.g., via `modus-themes-toggle`), the color of some UI remains the same, resulting in low color contract.

Examples
- status label
- header bar
- etc

## Checklist

- [X] I've read the README's [Issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#issues) section.
- [X] I'm running the latest agent-shell version.
- [X] I'm running the latest agent version (ie. Claude Code, Geminig CLI).
- [X] I'm running the latest agent ACP package (ie. claude-code-acp).
- [X] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [X] For issues, I'm including ACP traffic (as per README).

xenodium:
https://github.com/xenodium/agent-shell/pull/208 was merged. Closing.

PR #208, opened 2026-01-19
Make status-label color respect new theme

whhone:
**Problem**: After changing the light/dark theme (e.g., via `modus-themes-toggle`), the color of the status label remain the same as the previous theme.

It makes the text less readable because the color contrast is low between the background and the text.

**Root Cause:** The `font-lock-face` uses the exact color value. The face won't change if the underlying face is changed.

**Fix**: Set `font-lock-face` to uses the face with a box. This way, when changing the theme, the face can change with the underlying face.

**Tested**:
- Tested in GUI Emacs. Seeing the color is changed relatively after `modus-themes-toggle`.
- Fixed unit test. (I might had broken it in e11a66bf8b5bc9c3bc3274049075a4d16fa27bf2)

## Checklist

- [X] I've filed a feature request/discussion for this change (https://github.com/xenodium/agent-shell/issues/207).
- [X] *I've reviewed all code in PR myself and I'm happy with its quality*.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1517,13 +1517,11 @@ DESTINATION-DIR is required and must be provided."
                    (_ '("unknown" warning))))
          (label (car config))
          (face (cadr config))
-         (color (face-foreground face nil t))
          ;; Wrap the label in [ and ] in TUI which cannot render the box border.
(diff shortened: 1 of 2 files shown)


xenodium (approved):
Thanks for the PR!

PR #212, opened 2026-01-21
fix: Simplify language detection

necaris:
I use `delight` to hide some minor modes, which means that `mode-name` is a mode-line format string and doesn't play well with the existing language detection functionality. Switching to `major-mode` seems to be more robust.

## Checklist

- [ ] *I've reviewed all code in PR myself and I'm happy with its quality*.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -3809,12 +3809,7 @@ Available values:
     (let ((start (region-beginning))
           (end (region-end))
           (content (buffer-substring-no-properties (region-beginning) (region-end)))
-          (language (cond ((listp mode-name)
-                           (downcase (car mode-name)))
-                          ((stringp mode-name)
-                           (downcase mode-name))
-                          (t
-                           "")))
+          (language (string-remove-suffix "-mode" (string-remove-suffix "-ts-mode" (symbol-name major-mode))))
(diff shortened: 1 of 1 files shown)


xenodium:
Thanks for the PR.

> I use delight to hide some minor modes, which means that mode-name is a mode-line format string and doesn't play well with the existing language detection functionality. 

Not familiar with `delight`, mind sharing what `mode-name` looks like so I better understand the incompatibility?

necaris:
@xenodium sure, apologies for the lack of information in the pull request description. [`delight`](https://elpa.gnu.org/packages/delight.html) is one of the things I use to keep my mode-line uncluttered. As a result, in for e.g. a Python-mode buffer `M-x describe-variable` for `mode-name` produces:
```elisp
(delight-mode-name-inhibit "Python" (:eval (doom-modeline-icon-for-buffer)))
```
Here, `delight-mode-name-inhibit` is a boolean symbol and my understanding is that it's used [as a conditional for whether to show the string or the icon](https://www.gnu.org/software/emacs/manual/html_node/elisp/Mode-Line-Data.html)

xenodium:
Thanks for the details and contribution. Merged PR.

PR #213, opened 2026-01-21
fix: Improve file completion (and command completion)

necaris:
With the current file- and command-completion implementation, completion is only triggered _immediately after_ the trigger symbol (e.g. `@`):
Screenshot 2026-01-21 at 14 16 31

However, typing any prefix to narrow the completions results in nothing at all:
Screenshot 2026-01-21 at 14 16 52

This PR aligns the completions more closely with other completion-at-point functions (mostly following implementations in [`cape`](https://github.com/minad/cape)) and adds metadata so that [`corfu`](https://github.com/minad/corfu) can annotate results:
Screenshot 2026-01-21 at 13 57 16

## Checklist

- [x] *I've reviewed all code in PR myself and I'm happy with its quality*.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -3107,49 +3107,54 @@ The captured screenshot file path is then inserted into the shell prompt."
               (project-files proj))))
    (t nil)))
 
+(defun agent-shell--completion-bounds (char-class trigger-char)
+  "Find completion bounds for CHAR-CLASS, if TRIGGER-CHAR precedes them.
+Returns (start . end) if TRIGGER-CHAR is found before the word, nil otherwise."
+  (save-excursion
+    (let ((end (progn (skip-chars-forward char-class) (point)))
+          (start (progn (skip-chars-backward char-class) (point))))
+      (when (eq (char-before start) trigger-char)
+        (cons start end)))))
+
+(defun agent-shell--capf-exit-with-space (_string _status)
+  "Insert space after completion."
+  (insert " "))
+
 (defun agent-shell--file-completion-at-point ()
   "Complete project files after @."
-  (when-let* ((triggered (eq (char-before) ?@))
-              (start (point))
-              (end (save-excursion
-                     (skip-chars-forward "[:alnum:]/_.-")
-                     (point)))
-              (prefix (buffer-substring start end))
+  (when-let* ((bounds (agent-shell--completion-bounds "[:alnum:]/_.-" ?@))
               (files (agent-shell--project-files)))
-    (list start end
-          (seq-filter
-           (lambda (f) (string-prefix-p prefix (file-name-nondirectory f)))
-           files)
+    (list (car bounds) (cdr bounds)
+          files
           :exclusive 'no
-          :exit-function
-          (lambda (_status _string)
-            (insert " ")))))
+          :company-kind (lambda (f) (if (string-suffix-p "/" f) 'folder 'file))
+          :exit-function #'agent-shell--capf-exit-with-space)))
 
 (defun agent-shell--command-completion-at-point ()
   "Complete available commands after /."
-  (when (eq (char-before) ?/)
-    (let* ((start (point))
-           (end (save-excursion
-                  (skip-chars-forward "[:alnum:]_-")
-                  (point)))
-           (prefix (buffer-substring start end))
-           (commands (map-elt agent-shell--state :available-commands)))
-      (list start end
-            (mapcar (lambda (command)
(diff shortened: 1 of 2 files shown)


xenodium:
Thanks for the PR. Could you please share details on the improvement? Is it fixing an issue? Performace? Something else?

necaris:
@xenodium apologies for the lack of information in the PR description -- I've updated it with some screenshots and more explanation!

agent-shell.el
```diff
@@ -253,8 +253,8 @@ Can be one of:
  \='text: Display simple text-only header.
  nil: Display no header."
   :type '(choice (const :tag "Graphical" graphical)
```
xenodium, line 255: Can we keep the existing indent please?

agent-shell.el
```diff
@@ -3109,47 +3109,55 @@ The captured screenshot file path is then inserted into the shell prompt."
...
-          (seq-filter
-           (lambda (f) (string-prefix-p prefix (file-name-nondirectory f)))
-           files)
-          :exclusive 'no
-          :exit-function
-          (lambda (_status _string)
-            (insert " ")))))
+  (let ((path-chars "[:alnum:]/_.-"))
+    (save-excursion
+      (let* ((end (progn (skip-chars-forward path-chars) (point)))
```
xenodium: Can't we flatten here a bit? Something like:

```emacs-lisp
      (when-let* ((end (progn (skip-chars-forward path-chars) (point)))
                  (start (progn (skip-chars-backward path-chars) (point)))
                  (triggered (eq (char-before start) ?@))
                  (files (agent-shell--project-files)))
```

agent-shell.el
```diff
@@ -3109,47 +3109,55 @@ The captured screenshot file path is then inserted into the shell prompt."
...
-                    (seq-filter
-                     (lambda (command)
-                       (string-prefix-p prefix (map-elt command 'name)))
-                     commands))
-            :exclusive t ; prevent / file completion fallback (already covered by @).
-            :exit-function
-            (lambda (_status _string)
-              (insert " "))))))
+  (save-excursion
+    (let* ((cmd-chars "[:alnum:]_-")
```
xenodium: Same here. Can't we flatten?

agent-shell.el
necaris: Yes, sorry, that was my auto-formatting getting over-enthusiastic. Fixed!

agent-shell.el
```diff
@@ -3109,47 +3109,55 @@ The captured screenshot file path is then inserted into the shell prompt."
...
+           (end (progn (skip-chars-forward cmd-chars) (point)))
+           (start (progn (skip-chars-backward cmd-chars) (point))))
+      (when-let* ((triggered (eq (char-before start) ?/))
+                  (commands (map-elt agent-shell--state :available-commands))
+                  ;; Build hash table for O(1) annotation lookup instead of O(n) seq-find
```
xenodium: We can likely do without this comment. I'm guessing it was added by an agent, referring to previous code?

ps. While technically more performant, likely negligeable considering how many commands we can have in the more extreme cases. 20? 50? Ok to keep code though.

agent-shell.el
```diff
@@ -4164,7 +4167,7 @@ Mark model using CURRENT-MODEL-ID."
 Called with no arguments, should return a string path or nil to disable.
 When nil, transcript saving is disabled."
   :type '(choice (const :tag "Disabled" nil)
-                 (function :tag "Custom function"))
+          (function :tag "Custom function"))
```
xenodium: Can we keep existing indent here too please?

xenodium (commented):
Thanks for adding PR details. I've left some minor comments. Mind taking a look?

agent-shell.el
necaris: Simplified it to an alist rather than a hash-table, which should reduce it from O(n<sup>2</sup>) to O(n), with the benefit of simpler handling than the hash table :-)

necaris:
Thank you for the review @xenodium ! I've addressed your comments and added a couple of tests on top of a slightly bigger refactor -- would appreciate your thoughts!

xenodium (approved):

xenodium:

xenodium:
Made a couple of minor tweak in https://github.com/xenodium/agent-shell/commit/7a02091f48387fb3d5ba7eadd35d9fd80e3dd9e0

nineluj:
Nice, this works great. Thanks for the improvement!

Issue #170, opened 2026-01-01
Suppressing effects of `find-file-noselect`


ultronozm:
When the agent modifies a file not currently visited by a buffer (`fs/write_text_file` handler), it does so using `find-file-noselect`, which runs `find-file-hook`.  With `auto-insert-mode` enabled, this hook causes `auto-insert` to fire, which, when the file is empty, prompts the user to "Perform auto-insertion? y/n".  This behavior is undesired (e.g., the agent might just be creating temp files to work with while testing something).

Proposed minimal fix:

## buffer: *vc-diff*
```diff
diff --git a/agent-shell.el b/agent-shell.el
index f331ea8..1ba9521 100644
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1022,7 +1022,8 @@ If the buffer's file has changed, prompt the user to reload it."
                (content .params.content)
                (dir (file-name-directory path))
                (buffer (or (find-buffer-visiting path)
-                           (find-file-noselect path))))
+                           (let ((auto-insert nil))
+                             (find-file-noselect path)))))
           (when (and dir (not (file-exists-p dir)))
             (make-directory dir t))
           (with-temp-buffer
```

Alternatively, we could avoid `find-file-noselect` and instead use `write-region` and friends, although I haven't considered what might be the trade-offs there.

xenodium:
Thank you

> Proposed minimal fix:

Applied in https://github.com/xenodium/agent-shell/commit/f314f54e025d3c78280c402df9bdbf15a0797702

> Alternatively, we could avoid find-file-noselect and instead use write-region

Using find-file-noselect (or similar) is handy as it keeps the buffer open, so you get quick access to it thereafter.

ultronozm:
thanks, good for me

PR #214, opened 2026-01-21
Fix auto-insert variable declaration in agent-shell

ultronozm:
This is a touch-up on #170, fixing the fix I had provided there
(commit f314f54).  That fix worked when agent-shell.el was evaluated
via C-c C-e, but not after byte-compilation in a fresh Emacs session.

* agent-shell.el (auto-insert): Declare as special variable to prevent
byte-compilation from treating let-bindings as lexical.
(agent-shell--on-fs-write-text-file-request): Remove identity call
that had been used to silence unused variable warning.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -73,6 +73,10 @@
 (declare-function projectile-project-root "projectile")
 (declare-function projectile-project-p "projectile")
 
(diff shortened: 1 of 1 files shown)


xenodium:
Thank you

Issue #222, opened 2026-01-24
Inhibit some minor modes during agent writes


ultronozm:
Some minor modes (e.g., `aggressive-indent-mode`) react to all buffer modifications.  When `agent-shell` applies write updates to a file, it visits the file.  Such modes may run as a side effect and introduce unwanted changes, for instance, to whitespace.  It would be useful to have a way to inhibit such changes.

Using `inhibit-modification-hooks` would be overly broad, because some side effects are desired.

In my local fork https://github.com/ultronozm/agent-shell/tree/write-inhibit-minor-modes, I've added a user option `agent-shell-write-inhibit-minor-modes` that allows the user to supply a list of minor modes that get temporarily disabled during write edits.  Happy to send as a PR if the direction seems reasonable, or maybe you have other ideas.

xenodium:
Thanks for filing. While I share the view that `inhibit-modification-hooks` is broad, I'm torn. On the other hand, temporaribly disabling and enabling minor modes may invite additional side effects every time a buffer is modified (ie. remove/add overlays, kill/create buffer-local variables, run hooks, cancel/add timers, etc).

> because some side effects are desired

Do we have concrete examples of modification-hooks we do want?

I'm thinking that even when inhibit-modification-hooks is set, we still get the benefits from: `before-save-hook`, `write-file-functions`, and `after-save-hook`. Perhaps these are more relevant anyway since agents typically replace the entire file (I've yet to see one that doesn't)?

ultronozm:
> Do we have concrete examples of modification-hooks we do want?

eglot/lsp-mode come to mind (and other packages that use before/after change functions to stay in sync, such as overleaf.el).

I also see a few things mentioned in the docstring for `inhibit-modification-hooks` (e.g., file locks) that sound risky to disable here.

I agree that temporarily disabling minor modes can have unwanted side effects, but we only pay that cost for the modes we specify (`aggressive-indent-mode` for me, where the overhead is negligible).

xenodium:
Fair. Let's give it a go. Want to send a PR? I'd say if you already know of problematic modes that we include them in agent-shell-write-inhibit-minor-modes by default?

ultronozm:
Sure.  The only one that's been trouble for me is `aggressive-indent-mode` (agents seem to assume that nothing changes between successive edits, so this can trip them up), which I've now included by default

xenodium:
Thank you for the contribution! Merged it.

PR #224, opened 2026-01-24
Add option to inhibit minor modes during file writes #222

ultronozm:
* agent-shell.el (agent-shell-write-inhibit-minor-modes): New
defcustom to specify minor modes to disable temporarily during agent
file writes.
(agent-shell--call-with-inhibited-minor-modes): New helper function.
(agent-shell--on-fs-write-text-file-request): Use it.
* README.org (Per-session containers, Customizations): Document it.


xenodium:
Thanks for the PR. Merged.

Issue #136, opened 2025-11-24
Support for emacs MCPs like claude-code-ide.el


acmorrow:
## Checklist

- [x] I've read the README's [Issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#issues) section.
- [n/a] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [n/a] For issues, I'm including ACP traffic (as per README).

I've been using https://github.com/manzaltu/claude-code-ide.el for a while now, with good results. However, I'm quite interested in this project: the terminal-in-emacs aspect of `claude-code-ide.el` introduces many limitations and glitches.

So, I'd like to switch over to `agent-shell`! However, I've also invested in creating my own `claude-code-ide.el` Emacs MCPs that allow claude-code-ide.el to interact with emacs directly (e.g. [running lsp-format-buffer](https://github.com/acmorrow/claude-code-ide-extras/blob/bafa6385a129016655a4e9b5ab0115babec501a4/claude-code-ide-extras-lsp.el#L86-L104)).

Having looked at the `README.md` for this project, it isn't clear whether or not `agent-shell` has support for custom "in emacs" MCPs. If it does, that's great, you can close this issue and just let me know where in the docs to find details.

If not, I think it would be a worthy addition to the project: the ability to expose native emacs capabilities to the agent is quite powerful.

Finally, if I've just misunderstood the relationship between ACP / MCP / emacs / agent-shell, please let me know how I should reframe my thinking here and I'll try to use the intended mechanisms to achieve my desired effect.

xenodium:
Update to the latest and check out @jinnovation's introduction of `agent-shell-mcp-servers` for this purpose. More details at https://github.com/xenodium/agent-shell/issues/137 and https://github.com/xenodium/agent-shell/pull/138.

matthewbauer:
I think a piece missing here is the ability to run an MCP server from Emacs. But I don't think that actually has to live in agent-shell. You can use:

- [claude-code-ide-mcp-server](https://github.com/manzaltu/claude-code-ide.el/blob/main/claude-code-ide-mcp-server.el)
- [emacs-mcp-server](https://github.com/rhblind/emacs-mcp-server)

It might be nice to document them though!

xenodium:
> It might be nice to document them though!

While I haven't used either of these packages, they should work with something like the following:

[...]

PR #237, opened 2026-01-31
Allow lambdas in agent-shell-mcp-servers

matthewbauer:
This makes it possible to setup a session for claude-code-ide.

## Checklist

- [x] I've filed a feature request/discussion for this change. (https://github.com/xenodium/agent-shell/issues/136)
- [x ] I've run `M-x checkdoc` and `M-x byte-compile-file`.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -2677,6 +2677,15 @@ Must provide ON-SESSION-INIT (lambda ())."
    :on-failure (agent-shell--make-error-handler
                 :state agent-shell--state :shell shell)))
 
+(defun agent-shell--eval-dynamic-values (obj)
+  "Recursively evaluate any function/lambda values in OBJ."
+  (cond
+   ((and (functionp obj) (not (symbolp obj))) (agent-shell--eval-dynamic-values (funcall obj)))
+   ((consp obj)
+    (cons (agent-shell--eval-dynamic-values (car obj))
+          (agent-shell--eval-dynamic-values (cdr obj))))
+   (t obj)))
+
 (defun agent-shell--mcp-servers ()
   "Return normalized MCP servers configuration for JSON serialization.
 
@@ -2686,15 +2695,7 @@ normalized server configs."
   (when agent-shell-mcp-servers
     (apply #'vector
            (mapcar (lambda (server)
-                     (when (functionp server)
-                       (setq server (funcall server)))
-                     (when (consp server)
-                       (setq server (mapcar (lambda (obj)
-                                              (cons
-                                               (car obj)
-                                               (if (functionp (cdr obj))
-                                                   (funcall (cdr obj))
-                                                 (cdr obj)))) server)))
+                     (setq server (agent-shell--eval-dynamic-values server))
                      (let ((normalized (copy-alist server)))
                        (when (map-contains-key normalized 'args)
                          (let ((args (map-elt normalized 'args)))


agent-shell.el
```diff
@@ -402,8 +402,31 @@ Example configuration with multiple servers:
            (command . \"npx\")
            (args . (\"-y\"
                     \"@modelcontextprotocol/server-filesystem\" \"/tmp\"))
-           (env . ()))))"
-  :type '(repeat (alist :key-type symbol :value-type sexp))
+           (env . ()))))
+
+You can also pass a function so that you can provide extra context like
```
matthewbauer, line 407: Feel free to reword this as you see fit. I just thought it might be nice to have this extra guide.

agent-shell.el
```diff
@@ -402,8 +402,31 @@ Example configuration with multiple servers:
...
+the current working directory (`agent-shell-cwd') .
+
+For example, you can set up an Emacs MCP using external `claude-code-ide'
+package. See documentation of that package for more configuration:
+
+  (require \='claude-code-ide-mcp-server)
+  (setq agent-shell-mcp-servers
+        `((lambda ()
+            (let* ((project-dir (agent-shell-cwd))
```
matthewbauer, line 416: need a lambda so we can get cwd of agent shell

agent-shell.el
```diff
@@ -402,8 +402,31 @@ Example configuration with multiple servers:
...
+                   (session-id (format \"agent-shell-%s-%s\"
+                                 (file-name-nondirectory
+                                   (directory-file-name project-dir))
+                                 (format-time-string \"%Y%m%d-%H%M%S\"))))
+              (puthash session-id `(:project-dir ,project-dir)
```
matthewbauer, line 421: there's some more things that claude-code-ide stores (like :last-active-buffer, :buffer) but I don't really use them for my tools. I don't think the provided tool (claude-code-ide-emacs-tools-setup) do either.

agent-shell.el
xenodium: Good call. Makes sense to delay it.

I'm thinking we should consider making the delay system a little more generic so we can embed lambdas anywhere to delay evaluation.

For this case, maybe something like this (untested):

```emacs-lisp
(setq agent-shell-mcp-servers
        '(((name . "emacs")
           (type . "http")
           (headers . ())
           (url . (lambda ()
                    (require 'claude-code-ide-mcp-server)
                    (let* ((project-dir (agent-shell-cwd))
                           (session-id (format "agent-shell-%s-%s"
                                         (file-name-nondirectory
                                           (directory-file-name project-dir))
                                         (format-time-string "%Y%m%d-%H%M%S"))))
                      (puthash session-id `(:project-dir ,project-dir)
                               claude-code-ide-mcp-server--sessions)
                      (format "http://localhost:%d/mcp/%s"
                         (claude-code-ide-mcp-server-ensure-server)
                         session-id)))))))
```

agent-shell.el
```diff
@@ -2663,6 +2686,8 @@ normalized server configs."
   (when agent-shell-mcp-servers
     (apply #'vector
            (mapcar (lambda (server)
+                     (when (functionp server)
```
xenodium: We can eval/expand using something like this:

```emacs-lisp
(setq server (agent-shell--eval-dynamic-values server))
```

Implementation looking like (untested):

```emacs-lisp
(defun agent-shell--eval-dynamic-values (obj)
    "Recursively evaluate any function/lambda values in OBJ."
    (cond
     ((functionp obj) (agent-shell--eval-dynamic-values (funcall obj)))
     ((consp obj)
      (cons (agent-shell--eval-dynamic-values (car obj))
            (agent-shell--eval-dynamic-values (cdr obj))))
     (t obj)))
```

xenodium:
ps. More out of curiosity than anything... what are some of the typical uses you're getting out of the Emacs mcp integration? Like what are you favorite or more popular ones?

matthewbauer:
> ps. More out of curiosity than anything... what are some of the typical uses you're getting out of the Emacs mcp integration? Like what are you favorite or more popular ones?

For work, we have a really large Haskell code base that I normally work on with GHCi, the Haskell repl. I have a comint buffer that runs it (+ compilation-shell-minor-mode) and my normal workflow is using this. So I give Claude some tools that let it reload GHCi, check output, make changes, repeat. I could just have it create a new process after each change but that's going to be much slower and different from my normal workflow. It's worked pretty well so far, albeit some kinks to still work out.

There's some interesting other tools people are working on too: https://github.com/acmorrow/claude-code-ide-extras

agent-shell.el
matthewbauer: hmm the problem seems to be that `(functionp 'headers)` is t (from `web-server`).

matthewbauer: ok updated with a similar idea but slightly different implementation

xenodium: What if we restrict the check to lambdas only? Maybe something like:

`(and (functionp obj) (not (symbolp obj)))`

xenodium (changes requested):
We may be able to restrict the function check to lambdas only. Check the comment.

agent-shell.el
matthewbauer: works for me!

xenodium (approved):
Thanks for the follow-ups.

xenodium:
Merged. Tweaked the docstrings in https://github.com/xenodium/agent-shell/commit/06ddf28ef4eb8918cc5362270dbbc31a608ea7fb and https://github.com/xenodium/agent-shell/commit/9bd60cca9f7ddb57e300270861af7b484d67dd33

Issue #236, opened 2026-01-31
Change send key from Enter to C-c C-c


HIRANO-Satoshi:
Currently, the send key is Enter, which makes writing multi-line input awkward. I think C-c C-c would be a better default for sending, allowing natural multi-line editing.

I know C-c C-c is already used as the send key in viewport mode, but viewport mode cannot reference previous turns, so it’s not a full workaround.

As a local workaround, I customized the keybindings like this:

```
(with-eval-after-load 'agent-shell
  (define-key agent-shell-mode-map (kbd "RET") 'newline)
  (define-key agent-shell-mode-map (kbd "C-c C-c") 'shell-maker-submit)
  (define-key agent-shell-mode-map (kbd "C-c C-k") 'agent-shell-interrupt))
```

xenodium:
Thank you for filing!

> As a local workaround, I customized the keybindings like this

This is actually the typical way of customizing bindings, so perfectly ok for anyone to override these to their liking.

> I think C-c C-c would be a better default for sending, allowing natural multi-line editing.

The challenge here is that `agent-shell` is a comint buffer, so changing the default behaviour will feel foreign to folks who are used to emacs shells. Also keep in mind that `C-c C-c` is already a common binding in comint buffers to cancel ongoing commands (in our case prompts).

With that in mind, I think it makes sense to keep the familiar bindings. Having said that, you found `agent-shell-mode-map` which enables ya to customize to your preferred bindings.

HIRANO-Satoshi:
It's okay, but C-c C-c is also a common key sequence to send.

How about add something like the following to index.html? If you like it, would you mind to add?

```
If you prefer C-c C-c to send and C-c C-k to interrupt, please add the following remapping.

(use-package agent-shell
  :config
  (with-eval-after-load 'agent-shell
    (define-key agent-shell-mode-map (kbd "RET") 'newline)
    (define-key agent-shell-mode-map (kbd "C-c C-c") 'shell-maker-submit)
    (define-key agent-shell-mode-map (kbd "C-c C-k") 'agent-shell-interrupt)))
```

xenodium:
Yup. We can add a section about customising bindings. 

One thing, the bindings you're describing are indeed used by the viewport interaction mode. 

`(setq agent-shell-prefer-viewport-interaction t)`

Check out https://xenodium.com/agent-shell-0-25-updates

HIRANO-Satoshi:
I know. Since they are separate modes, I think no problem is there.

[...]

PR #238, opened 2026-01-31
Change send key from Enter to C-c C-c

Muzamilkhan35713:
I change RET to C-c C-c.  it is first Contribution so if there are any error so tell me. I try to best to correct it.
Screenshot 2026-01-31 230740

### DIFF
--- a/README.org
+++ b/README.org
@@ -489,7 +489,7 @@ always go to Evil modes if you need to with ~C-z~).
     :config
     ;; Evil state-specific RET behavior: insert mode = newline, normal mode = send
     (evil-define-key 'insert agent-shell-mode-map (kbd "RET") #'newline)
-    (evil-define-key 'normal agent-shell-mode-map (kbd "RET") #'comint-send-input)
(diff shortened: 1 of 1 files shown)


Rutherther:
Hey, have you seen #236?

Muzamilkhan35713:
yes

xenodium:
Not an evil user myself, but this looks like it will create a binding conflict with cancelling.

cc'ing @juboba who is more familiar with evil than I am.

juboba:
@xenodium I agree, C-c C-c in normal or visual mode is used to cancel the on-flight request, so you should also rebind that to something else.
Also  C-c C-c is more idiomatic in Emacs than in Evil world for "commit".
That said, since these are examples it could be added as an alternative (?).

xenodium:
@Muzamilkhan35713 thanks for the PR.

I think for this instance we should likely keep the current docs as they are. Thankfully, we have the means to customize the behaviour, so anyone can change the bindings to their preferred keys.

Issue #44, opened 2025-10-01
Add compose buffer where it's easier to craft multi-line


xenodium:
Some form of compose buffer (a la `M-x chatgpt-shell-prompt-compose`):

![Image](https://github.com/user-attachments/assets/804e1402-61cf-4fcf-a50b-a468fc072ec0)

I'm also considering merging the concept of compose buffer and [viewport/pager](https://xenodium.com/an-experimental-e-shell-pager). These are separate buffers, though I'm wondering if it can just be a "viewport-mode or pager-mode" on the shell itself?

This is what paging currently looks like in the compose buffer:

![Image](https://github.com/user-attachments/assets/18210534-24fa-4609-a518-dc1613aad768)

hexmode:
Can you bump this up in priority?  This is the one enhancement I keep needing.

xenodium:
I've started working on it. I have a pretty good idea of how it should behave (based on chatgpt-shell's version). That one went through a bunch of iterations and has a lot of cruft and internal complexity, but the UX is good. I'll try to distill and simplify for agent-shell.

xenodium:
@hexmode https://github.com/xenodium/agent-shell/commit/e59cd30a8c29f911a9078615602f26891c857420 adds a basic `M-x agent-shell-prompt-compose` command. Composes and sends only for now.

hexmode:
Thanks.  I tried it and it does what I want 👍

xenodium:
@hexmode  Heads up, I'm working on a much richer compose experience. If you're familiar with chatgpt-shell's compose experience, that's more or less the target UX.

Of the main compose features, paging is still missing. If you'd like to give it a try, you'll need v0.21.1 and use:

```emacs-lisp
(setq agent-shell-prompt-compose--experimental-compose t)
```

@djr7C4 FYI

xenodium:
Heads-up, things have changed quite a bit:

The new mode is called “viewport interaction” and can become the primary way to interact with the agent. Needs agent-shell v0.24.1, shell-maker 0.84.2 and can be enabled with

```emacs-lisp.
(setq agent-shell-prefer-viewport-interaction t)
```

From then on, just do `M-x agent-shell` as usual.

hexmode:
loving viewport mode

xenodium:
> loving viewport mode

Great to hear!

[...]

PR #240, opened 2026-01-31
Fix viewport shell buffer association 

Rutherther:
Hey, the first commit here fixes a bug where if you had multiple viewports associated with a project, you could not really use them. Instead, all the submissions (C-c C-c) would keep getting submitted to the latest agent you've opened for the project. This meant that for example `Claude Code` viewport buffer could suddenly become a `Gemini CLI` viewport buffer, despite the name still being Claude Code viewport...

The second commit could be said to add a new feature. But I personally think this is what the user would expect, always when you're already in a shell buffer (or viewport buffer), return that buffer, not the latest one associated with the project... But feel free to discard that commit and just apply the first one if you do not feel that's the right thing.

IMO there is still room for improvement with multiple agents in the same session, somehow allowing to choose which one is the current 'primary' one, preferring that to the others. Currently always the latest one is used in the send functions. But that's a different topic and I am going to open an issue for that.

PS: I think there could be some refactoring done with this new function and also the already existing `shell-buffer--viewport-buffer`. But I've left that for now.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -2952,20 +2952,22 @@ If FILE-PATH is not an image, returns nil."
 (cl-defun agent-shell--shell-buffer (&key viewport-buffer no-error no-create)
   "Get an `agent-shell' buffer for the current project.
 
-When VIEWPORT-BUFFER is non-nil, return counterpart.
+Resolution order:
+1. If VIEWPORT-BUFFER is provided, derive shell buffer from its name.
+2. If inside of a viewport buffer, derive shell bufer from its name.
+3. If currently in an `agent-shell-mode' buffer, return it.
+4. Otherwise, return the first shell buffer in the current project.
+
 When NO-CREATE is nil (default), prompt to create a new shell if none exists.
 When NO-CREATE is non-nil, return existing shell or nil/error if none exists.
 When NO-ERROR is non-nil, return nil instead of raising an error.
 
 Returns a buffer object or nil."
-  (let ((shell-buffer (if viewport-buffer
-                          (seq-first (seq-filter (lambda (shell-buffer)
-                                                   (equal (agent-shell-viewport--buffer
-                                                           :shell-buffer shell-buffer
-                                                           :existing-only t)
-                                                          viewport-buffer))
-                                                 (agent-shell-buffers)))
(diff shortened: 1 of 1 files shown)


xenodium:
Thanks for the PR!

> Hey, the first commit here fixes a bug where if you had multiple viewports associated with a project, you could not really use them. Instead, all the submissions (C-c C-c) would keep getting submitted to the latest agent you've opened for the project. This meant that for example `Claude Code` viewport buffer could suddenly become a `Gemini CLI` viewport buffer, despite the name still being Claude Code viewport...

Yep. Thanks for this. Been meaning to look at this issue.

> The second commit could be said to add a new feature. But I personally think this is what the user would expect, always when you're already in a shell buffer (or viewport buffer), return that buffer, not the latest one associated with the project... But feel free to discard that commit and just apply the first one if you do not feel that's the right thing.

Agree. This is more predictable.

> IMO there is still room for improvement with multiple agents in the same session, somehow allowing to choose which one is the current 'primary' one, preferring that to the others. Currently always the latest one is used in the send functions. But that's a different topic and I am going to open an issue for that.

Funny you mention this. Before I built agent-shell, I built chatgpt-shell, which had the concept of primary shell. You could only use a single viewport buffer, always associated to the primary shell. This became a pain point, so what we have in agent shell is somewhat of an evolution away from that ;)

With your latest changes, I suspect the mental model will be predictable and flexible enough.

> PS: I think there could be some refactoring done with this new function and also the already existing `shell-buffer--viewport-buffer`. But I've left that for now.

No doubt. There was a lot of iterating and experimentation to build the viewport/compose experience in agent-shell. I'm sure there's plenty of crust left ;) Some of that history is in this issue https://github.com/xenodium/agent-shell/issues/44

I'll take a look at the changes tomorrow or the day after.

Rutherther:
> Funny you mention this. Before I built agent-shell, I built chatgpt-shell, which had the concept of primary shell. You could only use a single viewport buffer, always associated to the primary shell. This became a pain point, so what we have in agent shell is somewhat of an evolution away from that ;)

I think I understood the actual problem here now. My word latest has been used as the last one opened, while you probably understood the last one used. But I was wrong, it isn't really the last opened currently.I have moved the explanation and proposal to #241.

PR #245, opened 2026-02-01
Fix maintenance of :tool-calls state

vermiculus:
`map-delete` is not destructive; we need to use `map-put!` here in order for the deletion to have any persistent effect (as is clearly the intent).

Note that since `map-delete` is not destructive, the `map-copy` is also not doing anything, so that has been removed.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -3339,9 +3339,8 @@ MESSAGE-TEXT: Optional message to display after sending the response."
   ;; block-id must be the same as the one used as
   ;; agent-shell--update-fragment param by "session/request_permission".
   (agent-shell--delete-fragment :state state :block-id (format "permission-%s" tool-call-id))
-  (let ((updated-tool-calls (map-copy (map-elt state :tool-calls))))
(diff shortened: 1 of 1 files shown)


vermiculus:
I think the change is obviously correct, though I'm not 100% sure what downstream impacts it will have. It looks like at least one place *may* have been working around this issue already, but this package's code is still pretty new to me:

https://github.com/xenodium/agent-shell/blob/b593b2f572fb00d1ad42d7c4214f4a94a01d687c/agent-shell.el#L970-L978

Truthfully, I would give all accessors of `:tool-calls` another look; I may have missed something in my own review.

It does look like this fixes potential bugs here:

https://github.com/xenodium/agent-shell/blob/b593b2f572fb00d1ad42d7c4214f4a94a01d687c/agent-shell.el#L638-L647

and it does look like `:tool-calls` is not intended to accumulate:

https://github.com/xenodium/agent-shell/blob/b593b2f572fb00d1ad42d7c4214f4a94a01d687c/agent-shell.el#L2900-L2903

Issue #241, opened 2026-01-31
Multiple agents in one project


Rutherther:
Hi,

I have been trying to compare multiple agents on similar topics and I have been struggling with this. First point of struggle is described and fixed in #240. But there is more.

Specifically when I open multiple agents, I would like to be sending data from buffer to either of them. I am usually using the main dwim for this. Currently it seems to always send to the latest agent opened. Meaning the older agents are not as usable. I can still talk to them, yes, but I cannot really interact with them from the editor.

Personally what I am imagining is that we could have sort of a 'primary' agent buffer that would be the one generally preferred. You could call a function, let's say `agent-shell-choose-primary` that will give you a menu with the buffers tochoose from, based on the project you're in. Then all the functions would be sending the data to the primary agent.

WDYT?

## Checklist

- [x] I've read the README's [Issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#issues) section.
- [x] I'm running the latest agent-shell version.
- [x] I'm running the latest agent version (ie. Claude Code, Geminig CLI).
- [x] I'm running the latest agent ACP package (ie. claude-code-acp).
- [] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [ ] For issues, I'm including ACP traffic (as per README).

Rutherther:
Bit of a discussion in #240, I said

> IMO there is still room for improvement with multiple agents in the same session, somehow allowing to choose which one is the current 'primary' one, preferring that to the others. Currently always the latest one is used in the send functions. But that's a different topic and I am going to open an issue for that.

@xenodium said:
> Funny you mention this. Before I built agent-shell, I built chatgpt-shell, which had the concept of primary shell. You could only use a single viewport buffer, always associated to the primary shell. This became a pain point, so what we have in agent shell is somewhat of an evolution away from that ;)

now that you mention this I have experimented some more and it seems that the idea has been that you always send to (with `agent-shell-send*` to the agent you visited as last one? But similarly to viewport <-> shell association this seems to be broken for viewports. I do have to switch to shell so that an agent becomes the one chosen. If this is fixed, we definitely do not have to make primary agents as I suggested initially. Let's keep these latest. That sounds useful to me. 

I haven't realized this because I use viewports most of the time, so I haven't really tried switching to shells every single time I wanted to switch the agent.

Rutherther:
[...]

PR #246, opened 2026-02-01
Respect viewport visits when searching for last visited shell buffer.

Rutherther:
This is the solution I proposed in #241. It seems to work as I would expect. This depend on my #240 PR, because currently there is no function to get a shell buffer from a viewport reliably (the agent-shell--shell-buffer cannot be used here, because that causes stack overflow agent-shell--shell-buffer calls agent-shell-buffers to determine the viewport buffer. So that one cannot be called from agent-shell-buffers in the current codebase.

I am not sure if this is the way forward, but I will be using this for now, it makes the experience much better for me when trying to use only viewports.

## Checklist

- [x] I've filed a feature request/discussion for this change.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -595,11 +595,21 @@ Returns an empty string if no icon should be displayed."
     (map-elt choices selected-name)))
 
 (defun agent-shell-buffers ()
-  "Return all shell buffers."
-  (seq-filter (lambda (buffer)
-                (with-current-buffer buffer
-                  (derived-mode-p 'agent-shell-mode)))
-              (buffer-list)))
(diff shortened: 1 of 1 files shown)


Rutherther:
Note that this PR adds only one commit Look at viewports when determining visited order.
The first two are taken from #240.

Rutherther:
No problem at all. Thanks for this package, it's suiting me the best from the ones I've seen so far.

xenodium:
Great to hear. I'm surprised I don't hear more from folks using viewport interaction. I'm guessing because it's not enabled by default. In any case, thanks for improving it.

PR #249, opened 2026-02-01
Fix tool call title duplication with special shell characters

ElleNajt:
Use `regexp-quote` to escape regex metacharacters when checking if command is already in title.

Without this, commands containing `[`, `]`, `*`, `+`, etc. cause `string-match` to fail, resulting in the command being prepended repeatedly.

```elisp
(string-match "ls *.txt" "ls *.txt")     => nil  ;; broken
(string-match-p (regexp-quote "ls *.txt") "ls *.txt") => 0  ;; fixed
```


ElleNajt:
Thanks -- fixed!

PR #250, opened 2026-02-01
Show [C] in modeline when running in container

ElleNajt:
Display a container indicator in the modeline when `agent-shell-container-command-runner` is set.

This makes it clear when the agent is running inside a container vs locally.


xenodium:
Thanks!

PR #252, opened 2026-02-02
C-c C-c to send

### LINKED ISSUE #236: Change send key from Enter to C-c C-c

HIRANO-Satoshi:
fix #236

## Checklist

- [X] I've filed a feature request/discussion for this change.

### DIFF
--- a/README.org
+++ b/README.org
@@ -487,6 +487,17 @@ All of the above settings can be applied on a per-project basis using [[https://
                   ("M-RET" . shell-maker-submit)))
   #+end_src
 
+  Or if you prefer C-c C-c to send and C-c C-k to interrupt, add the following remapping.
+
+  #+begin_src emacs-lisp :lexical no
+    (use-package agent-shell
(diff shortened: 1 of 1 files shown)


xenodium (approved):

HIRANO-Satoshi:
Thank you, sir.

PR #256, opened 2026-02-03
Agent shell buffer name format customization

nhojb:
Adds a new defcustom `agent-shell-buffer-name-format` to permit customization of the agent-shell buffer name.

- default: "Claude Code Agent @ My Project"
- kebab-case: "claude-code-agent @ my-project"
- custom: function which takes the agent-name and project-name

Notes: 
- Personally I'm not a fan of the default agent-shell buffer-name format. I prefer kebab-case e.g. like magit buffers.
- Since it's a small change I thought it was easier to just open the PR than a FR.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 /.agent-shell/
+
+*.elc


nhojb:
Closing until I fix some issues with `agent-shell-toggle`

nhojb:
Checked the MELPA version and the same issue with `agent-shell-toggle` is present:

- Works from agent-shell -> other-project-buffer
- Fails to find a agent-shell buffer when going other-project-buffer -> agent-shell

So not an issue with this PR 😅

xenodium:
> Checked the MELPA version and the same issue with agent-shell-toggle is present

Fixed in https://github.com/xenodium/agent-shell/commit/2c0c25f11901082d25b45e92cf83cc1cf99a4070

> Personally I'm not a fan of the default agent-shell buffer-name format. I prefer kebab-case e.g. like magit buffers

I'll accept this one, but please keep in mind that some package behaviour will have to remain opinionated. We Emacs users are a diverse bunch, so I would have to offer an unmaintainable number of options to satisfy everyone's preferences.

nhojb:
Thanks for accepting. Seems like a minor thing, but it improves my QoL 🙏 

One thing: I was wondering whether it would be better to transform the `agent-name` only for `kebab-case`? Was going to push this as a fix, but you beat me to it.

xenodium:
Sure. Send a PR.

PR #269, opened 2026-02-08
Add missing paren

Rutherther:
Seems that there is a missing paren, leading to compilation/evaluation errors.


xenodium:
Bumped version for MELPA in https://github.com/xenodium/agent-shell/commit/4d7f5a2607a3032095b307b41955a4a1f00a8680

Issue #267, opened 2026-02-08
FR: Session usage tracking


Lenbok:
I just update my `opencode` and found that I'm now getting some extra uninterpreted json being inserted into the shell buffer. It corresponds to a feature they've recently implemented to allow tracking session and context usage:

https://agentclientprotocol.com/rfds/session-usage

Here's the opencode PR, where you can see a screenshot of how it renders in an alternate ACP client:

https://github.com/anomalyco/opencode/pull/12299

Is support for that something you'd be interested in having in agent-shell? Presumably we could display the session usage status in the header somewhere.

Small example traffic file including the `usage_update` notification:
[agent-shell-opencode-usage-traffic.txt](https://github.com/user-attachments/files/25160185/agent-shell-opencode-usage-traffic.txt)

xenodium:
> Is support for that something you'd be interested in having in agent-shell?

Certainly. Interested in sending a PR?

We can potentially add `(cons :usage nil)` to `agent-shell--make-state` to store all usage details as they come. We can subsequently display wherever we find useful (we may have to play around with this). Could be the header, but who knows, it's getting a little crowded there. Could also be that we output usage into the shell itself when agent turn ends.

Maybe something like:

```
╭─ Usage ──────────────────────────────────────────────────────────╮
│ Tokens  11,602 total     Context 2 / 200,000     Cost    $0.0439 │
╰──────────────────────────────────────────────────────────────────╯
```

Lenbok:
Yep, I'll take a look at it.

Lenbok:
I think this is fine to close now #270 is merged.

PR #270, opened 2026-02-08
Track session usage details sent by server

Lenbok:
This adds tracking of session usage details (see #267).

In terms of exposing the information, at the moment there is:

- An interactive function `agent-shell-show-usage` that displays the usage information in the minibuffer whenever desired (I can add to the readme documentation if you're happy to keep it). 
- An end-of-turn summary as suggested in https://github.com/xenodium/agent-shell/issues/267#issuecomment-3867006102 (enabled by a custom var `agent-shell-show-usage-at-turn-end`, default nil)
- A single character fuel-gauge indicator in the header showing context usage (enabled by a custom var `agent-shell-show-context-usage-indicator`, default t)

Feel free to try it out and give feedback on the UI aspect.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -63,6 +63,7 @@
 (require 'agent-shell-pi)
 (require 'agent-shell-project)
 (require 'agent-shell-qwen)
+(require 'agent-shell-usage)
 (require 'agent-shell-worktree)
 (require 'agent-shell-ui)
 (require 'agent-shell-viewport)
@@ -175,14 +176,6 @@ See https://github.com/xenodium/agent-shell/issues/119"
   :type 'boolean
   :group 'agent-shell)
 
-(defcustom agent-shell-show-usage-at-turn-end nil
-  "Whether to display usage information when agent turn ends.
-
-When non-nil, displays a formatted box showing token counts,
-context window usage, and cost information after each agent response."
-  :type 'boolean
-  :group 'agent-shell)
-
 (defcustom agent-shell-context-sources '(files region error line)
   "Sources to consider when determining \\<agent-shell-mode-map>\\[agent-shell] automatic context.
 
@@ -281,15 +274,6 @@ Can be one of:
 
 (defcustom agent-shell-show-busy-indicator t
   "Non-nil to show the busy indicator animation in the header and mode line."
-  :type 'boolean
-  :group 'agent-shell)
-
-(defcustom agent-shell-show-context-usage-indicator t
-  "Non-nil to show the context usage indicator in the header and mode line.
-
-The indicator displays context window usage as a vertical bar character,
-color-coded from green (low) to yellow (high) to red (critical).
-Only appears when the ACP server provides usage information."
   :type 'boolean
   :group 'agent-shell)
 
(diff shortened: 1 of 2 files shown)


Lenbok:
@xenodium (or anyone else), would you like to try this out and give any feedback on the UI aspects? (I pretty much only use emacs in the terminal so haven't looked much at the GUI version). Also not sure if you're OK with the glyph approach to boxes etc. Maybe the header indicator popup could contain more info (essentially the same as the minibuffer / end-of-turn indicators)?

As an aside, while building this I've been using https://github.com/keegancsmith/emacs-mcp-server so the agent can inspect and make changes to your running emacs session to live-update / debug as we go. It's pretty cool. Do you use something like that in your workflow?

Lenbok:
@xenodium I think this is worth a review now.

xenodium:
Thanks for the nudge!

agent-shell.el
```diff
@@ -4283,6 +4468,36 @@ See https://agentclientprotocol.com/protocol/session-modes for details."
...
+             ;; Unicode vertical block characters from empty to full
+             (indicator (cond
+                         ((>= percentage 100) "█")  ; Full
+                         ((>= percentage 87.5) "▇")
+                         ((>= percentage 75) "▆")
+                         ((>= percentage 62.5) "▅")
+                         ((>= percentage 50) "▄")
+                         ((>= percentage 37.5) "▃")
+                         ((>= percentage 25) "▂")
+                         ((>= percentage 12.5) "▁")
```
xenodium: We should make this bracket `(> percentage 0)`.

I couldn't work out why I could not see the usage indicator even though I had usage data in the state. Turns out my percentage usage was 10.4 from just sending a "Hello" prompt.

agent-shell.el
```diff
@@ -4283,6 +4468,36 @@ See https://agentclientprotocol.com/protocol/session-modes for details."
...
+                         (t " ")))  ; Empty/near-empty
```
xenodium: We should likely return nil otherwise and make sure the usage indicator is hidden. When the context indicator is " ", and the status indicator is active, the latter looks a little off (separated by two spaces).

agent-shell.el
```diff
@@ -506,11 +523,141 @@ HEARTBEAT, and AUTHENTICATE-REQUEST-MAKER."
...
+  (let* ((total (map-elt usage :total-tokens))
+         (input (map-elt usage :input-tokens))
+         (output (map-elt usage :output-tokens))
+         (thought (map-elt usage :thought-tokens))
+         (cached-read (map-elt usage :cached-read-tokens))
+         (cached-write (map-elt usage :cached-write-tokens))
+         (context-used (map-elt usage :context-used))
+         (context-size (map-elt usage :context-size))
+         (cost (map-elt usage :cost-amount))
+         (currency (map-elt usage :cost-currency)))
```
xenodium: We can inline (map-elt usage :total-tokens) and all the other let vars (no need for local vars). The LLM models love to add unnecessary local vars. This is somewhat the equivalent of saving usage.totalTokens to a local var in other languages).

agent-shell.el
```diff
@@ -506,11 +523,141 @@ HEARTBEAT, and AUTHENTICATE-REQUEST-MAKER."
...
+                           (cons :cached-write-tokens 0)
+                           (cons :context-used 0)
+                           (cons :context-size 0)
+                           (cons :cost-amount 0.0)
+                           (cons :cost-currency nil)))))
 
 (defvar-local agent-shell--state
     (agent-shell--make-state))
 
+(cl-defun agent-shell--update-usage-from-response (&key state usage)
```
xenodium: nit: can we name agent-shell--save-usage (similar to agent-shell--save-tool-call)?

xenodium: Can we name `usage` as `acp-usage`? I'm trying to introduce a naming convention to know if the value that's being used is an acp structure.

agent-shell.el
```diff
@@ -506,11 +523,141 @@ HEARTBEAT, and AUTHENTICATE-REQUEST-MAKER."
...
+  (cond
+   ((>= num 1000000000) (format "%.0fb" (/ num 1000000000.0)))
+   ((>= num 1000000) (format "%.0fm" (/ num 1000000.0)))
+   ((>= num 1000) (format "%.0fk" (/ num 1000.0)))
+   (t (format "%d" num))))
+
+(defun agent-shell--usage-has-data-p (usage)
+  "Return non-nil if USAGE contains meaningful data.
+Checks if any token counts or context size are non-zero."
+  (let ((total (or (map-elt usage :total-tokens) 0))
```
xenodium: The LLM models love to add unnecessary local vars. Inlining would be fine, including both ors. More details about LLMs and local vars in next comment.

agent-shell.el
```diff
@@ -506,11 +523,141 @@ HEARTBEAT, and AUTHENTICATE-REQUEST-MAKER."
...
+    (when-let ((size (map-elt update 'size)))
+      (map-put! usage-state :context-size size))
+    (when-let ((cost (map-elt update 'cost)))
+      (when-let ((amount (map-elt cost 'amount)))
+        (map-put! usage-state :cost-amount amount))
+      (when-let ((currency (map-elt cost 'currency)))
+        (map-put! usage-state :cost-currency currency)))
+    (map-put! state :usage usage-state)))
+
+(cl-defun agent-shell--reset-usage (&key state)
```
xenodium: Do we need this? Don't we always start a session with reset values from agent-shell--make-state? I think we can remove?

agent-shell.el
```diff
@@ -506,11 +523,141 @@ HEARTBEAT, and AUTHENTICATE-REQUEST-MAKER."
...
+                               (cons :output-tokens 0)
+                               (cons :thought-tokens 0)
+                               (cons :cached-read-tokens 0)
+                               (cons :cached-write-tokens 0)
+                               (cons :context-used 0)
+                               (cons :context-size 0)
+                               (cons :cost-amount 0.0)
+                               (cons :cost-currency nil))))
+
+(defun agent-shell--get-usage (state)
```
xenodium: One-line function. We should remove and inline everywhere. 

for example: (map-elt (agent-shell--state) :usage)

agent-shell.el
```diff
@@ -2356,6 +2514,19 @@ BINDINGS is a list of alists defining key bindings to display, each with:
                                                                                    "#6699cc"))
                                                                       (dx . "8"))
                                                                     (map-elt header-model :mode-name))))
+                                      (when (map-elt header-model :context-indicator)
+                                        (let* ((indicator (map-elt header-model :context-indicator))
```
xenodium: Similarly. We can inline `(map-elt header-model :context-indicator)`

agent-shell.el
```diff
@@ -3068,8 +3241,20 @@ If FILE-PATH is not an image, returns nil."
                    ;; a session prompt request is finished.
                    ;; Avoid accumulating them unnecessarily.
                    (map-put! (agent-shell--state) :tool-calls nil)
+                   ;; Extract usage information from response
+                   (when-let ((usage (map-elt response 'usage)))
```
xenodium: Inline :)

(when (map-elt response 'usage)
  ...)

agent-shell.el
```diff
@@ -3068,8 +3241,20 @@ If FILE-PATH is not an image, returns nil."
...
+                     (agent-shell--update-usage-from-response :state (agent-shell--state) :usage usage))
                    (let ((success (equal (map-elt response 'stopReason)
```
xenodium, line 3108: We can flatten (drop inner when) with:

```
(when-let* ((success (equal (map-elt response 'stopReason)
                            "end_turn"))
            ((and success
                  agent-shell-show-usage-at-turn-end
                  (agent-shell--usage-has-data-p (agent-shell--get-usage (agent-shell--state)))))
....
```

agent-shell.el
```diff
@@ -4283,6 +4468,36 @@ See https://agentclientprotocol.com/protocol/session-modes for details."
                              available-session-modes)))
     (map-elt mode :name)))
 
+(defun agent-shell--context-usage-indicator ()
+  "Return a single character indicating context usage percentage.
+Uses Unicode vertical block characters to show fill level.
+Only returns an indicator if enabled and usage data is available."
+  (when (and agent-shell-show-context-usage-indicator
+             (agent-shell--usage-has-data-p (agent-shell--get-usage (agent-shell--state))))
+    (when-let* ((usage (agent-shell--get-usage (agent-shell--state)))
```
xenodium: Similarly, I think we can combine when and when-let*

agent-shell.el
Lenbok: Yeah, agree, that got me a couple of times too.

agent-shell.el
Lenbok: Might be worth adding that to the style section of the README (I did tell the agent to review the changes for compliance with that style information).

Lenbok:
Should be good to go. I didn't update the customization/command tables in the README, I assume you'll regenerate whenver you next do it.

xenodium (approved):
Thank you for the changes. Looks great.

agent-shell.el
xenodium: > Might be worth adding that to the style section of the README 

Good call. Added https://github.com/xenodium/agent-shell/commit/ecc08b65a08de96f0fa234e1370d9a65b0c3c957

> I did tell the agent to review the changes for compliance with that style information

Thank you

xenodium:
Heads-up, inline usage is now rendered as other collapsable sections

Screenshot 2026-02-11 at 17 25 02

Lenbok:
Thanks, looks good. You didn't like the color on the indicator in the header, that's gone?

xenodium:
> You didn't like the color on the indicator in the header, that's gone?

Ah no, that's my mistake. I'll bring color back. Having said that, I'm gravitating towards replacing the bar with actual percentage (and keep colors), like: `75%`. Something doesn't feel quite right with the percentage bar in the header, specially next to the throbber.

Lenbok:
OK, cool. I went with the single char bar gauge to minimize the amount of extra space usage (I do like the mouseover to see detailed info, but I think that's only with text header). Whatever you think looks best.

PR #274, opened 2026-02-09
Include current prompt when switching from shell to viewport

nhojb:
I often find myself starting a prompt in shell mode, then realising it will be easier to continue in viewport-mode. Currently after entering viewport mode via `C-c C-o`, the partial shell prompt is not sent to the viewport. So I have to switch back and manually copy it.

This PR automatically sends the current prompt to the viewport - assuming the viewport composer is empty.

## Checklist

### DIFF
--- a/agent-shell-viewport.el
+++ b/agent-shell-viewport.el
@@ -358,7 +358,7 @@ Optionally set its PROMPT and RESPONSE."
     (agent-shell-viewport--initialize
      :prompt (car current)
      :response (cdr current))
-    (goto-char (point-min))
+    (goto-char (point-max))
     current))
 
 (defun agent-shell-viewport-next-item ()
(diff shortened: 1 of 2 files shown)


xenodium:
Thank you for the PR.

`C-c C-o` (agent-shell-other-buffer) should more or less just jump between the two buffers without changing them. For the use-case you're describing, we should rely on `agent-shell-prompt-compose`, but that didn't really support this flow. I've modified it in https://github.com/xenodium/agent-shell/commit/318f080b773225347503866f7cf5b99c91e7e01c to support it and also handle additional edge cases: override vs append prompt, support region selection, automatically clear shell input when transfered, etc. Update to the latest and give `agent-shell-prompt-compose` a try.

nhojb:
Thanks! Probably a case were a tentative FR would have made sense. I'll give `agent-shell-prompt-compose` a try.

PR #275, opened 2026-02-09
Graphical Header improvements

nhojb:
Minor improvements to the graphical header layout:

- Adds some padding at the bottom of the header so that prompt text is not immediately adjacent to the bottom text in the header.
- Improves alignments of the header icon.
- Layout should be responsive

Tested with all agents and various fonts, sizes and themes. See example screenshots:

Screenshot 2026-02-20 at 10 29 58
Screenshot 2026-02-20 at 10 30 09
Screenshot 2026-02-20 at 10 30 26
Screenshot 2026-02-20 at 10 30 34
Screenshot 2026-02-20 at 10 16 46
Screenshot 2026-02-20 at 10 17 06
Screenshot 2026-02-20 at 10 17 27
Screenshot 2026-02-20 at 10 17 35
Screenshot 2026-02-20 at 10 21 34
Screenshot 2026-02-20 at 10 21 41
Screenshot 2026-02-20 at 10 22 03
Screenshot 2026-02-20 at 10 22 10

## Checklist

- [x] I've read the README's [Contributing](https://github.com/xenodium/agent-shell?tab=readme-ov-file#contributing) section.
- [x] I've filed a feature request/discussion for a new feature.
- [x] My code follows the project [style](https://github.com/xenodium/agent-shell?tab=readme-ov-file#style-or-personal-preference-tbh).
- [x] I've added tests where applicable.
- [x] I've updated documentation where necessary.
- [x] I've run `M-x checkdoc` and `M-x byte-compile-file`.
- [x] *I've reviewed all code in PR myself and will vouch for its quality*.

PR #276, opened 2026-02-09
Experimenting with a more reliable graphical header layout?

xenodium:
Looking into graphical header improvements for #275

### DIFF
--- a/agent-shell-ui.el
+++ b/agent-shell-ui.el
@@ -342,10 +342,6 @@ NAVIGATION controls navigability:
     (when body
       (when (or label-left label-right)
         (insert "\n\n"))
-      ;; Drop any leading body newlines as newlines are
-      ;; already inserted between labels and body.
-      (when (string-prefix-p "\n" body)
-        (setq body (string-trim-left body "\n")))
       ;; Never leave more than two trailing newlines.
       (when (string-suffix-p "\n\n" body)
         (setq body (concat (string-trim-right body) "\n\n")))


nhojb:
Thanks for the comment on #275. Much better to use the current char/font-height than rely on my hard-coded values. I did wonder about that after I submitted my PR 🤷‍♂️

I think the changes in #276 generally work well. Couple of niggles:

- There's perhaps a little too much bottom padding in shell mode
- There's a slight jump in the header text y-position when toggling between shell and viewport modes (see screen-recording).

I don't mind looking into the above. But equally happy if you want to continue.

https://github.com/user-attachments/assets/5ad6d313-ddd6-440b-9d48-c085a7eacb59

xenodium:
Yup. Feel free to take a look. Happy for you to take over the PR if you'd like. We may have to apply whatever we arrive to on both of our ends to get some coverage (we have slightly different setups).

BTW, you may have noticed I have the same background color throughout, so it's good that you have a different header background color.

xenodium:
ps. There's also a text-based header if you'd like that as an alternative.

xenodium:
> I don't mind looking into the above. But equally happy if you want to continue.

Still keen on getting that PR over the finish line?

nhojb:
Apologies been caught up with other stuff. Will look into tomorrow!

nhojb:
Just pushed my changes to #275

xenodium:
> Just pushed my changes to #275

Thanks. Dropping #276 in favor of https://github.com/xenodium/agent-shell/pull/275

PR #279, opened 2026-02-10
Add agent-shell-send-dwim to transient menu

matthewbauer:
Small change to add `d` to agent-shell-help-menu.

Right now, most of my workflow goes through agent-shell-help-menu with:

```
  (use-package agent-shell
    :bind ("C-c C-'" . agent-shell-help-menu))
```

This seems like a reasonable addition? I know I can bind it myself but it seems like something others might want.


xenodium:
Thanks! Btw, probably not super discoverable, but M-x agent-shell handles most dwim sending behavior too.

PR #283, opened 2026-02-12
Add session-meta parameter for system prompt customization

ElleNajt:
## Summary

Add `:session-meta` parameter to `agent-shell-start` and related functions, allowing callers to pass metadata when creating ACP sessions. This enables customizing the system prompt by appending to the default, which is useful for multi-agent systems where certain agents need persistent instructions that survive context compaction.

## Changes

- Added `:session-meta` parameter to `agent-shell--make-state`
- Added `:session-meta` parameter to `agent-shell-start` and `agent-shell--start`
- Store session-meta in buffer-local state
- Pass it to `acp-make-session-new-request` when creating the session

## Example Usage

```elisp
(agent-shell-start
  :config (agent-shell--resolve-preferred-config)
  :session-meta '((systemPrompt . ((append . "Custom instructions...")))))
```

Or set it on an existing buffer before the first prompt:

```elisp
(map-put! agent-shell--state :session-meta
          '((systemPrompt . ((append . "Instructions here")))))
```

## Use Case

This is useful for building systems like [meta-agent-shell](https://github.com/ElleNajt/meta-agent-shell) where "dispatcher" agents need persistent instructions. Previously, instructions sent as user messages would be lost on context compaction. With session-meta, they're appended to the system prompt and persist.

## Dependencies

Requires acp.el with `:meta` parameter support (merged in https://github.com/xenodium/acp.el/pull/13).

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -481,11 +481,15 @@ the session and returns the appropriate endpoint:
   :type '(repeat (choice (alist :key-type symbol :value-type sexp) function))
   :group 'agent-shell)
 
-(cl-defun agent-shell--make-state (&key agent-config buffer client-maker needs-authentication authenticate-request-maker heartbeat)
+(cl-defun agent-shell--make-state (&key agent-config buffer client-maker needs-authentication authenticate-request-maker heartbeat session-meta)
   "Construct shell agent state with AGENT-CONFIG and BUFFER.
 
 Shell state is provider-dependent and needs CLIENT-MAKER, NEEDS-AUTHENTICATION,
-HEARTBEAT, and AUTHENTICATE-REQUEST-MAKER."
+HEARTBEAT, and AUTHENTICATE-REQUEST-MAKER.
+
+SESSION-META is an optional alist of metadata to pass when creating the
+ACP session.  Supported keys include `systemPrompt' for appending to the
+default system prompt."
   (list (cons :agent-config agent-config)
         (cons :buffer buffer)
         (cons :client nil)
@@ -500,6 +504,7 @@ HEARTBEAT, and AUTHENTICATE-REQUEST-MAKER."
         (cons :session (list (cons :id nil)
                              (cons :mode-id nil)
                              (cons :modes nil)))
+        (cons :session-meta session-meta)
         (cons :last-entry-type nil)
         (cons :chunked-group-count 0)
         (cons :request-count 0)
@@ -664,13 +669,19 @@ If currently visiting an `agent-shell', transfer latest input."
         (agent-shell-viewport--show-buffer :override input))
     (agent-shell-viewport--show-buffer)))
 
-(cl-defun agent-shell-start (&key config)
+(cl-defun agent-shell-start (&key config session-meta)
   "Programmatically start shell with CONFIG.
 
-See `agent-shell-make-agent-config' for config format."
+See `agent-shell-make-agent-config' for config format.
+
+SESSION-META is an optional alist of metadata to pass when creating the
+ACP session.  Supported keys include:
+- `systemPrompt': Either a string (replaces default) or an alist with
+  key `append' containing a string to append to the default system prompt."
   (agent-shell--start :config config
                       :no-focus nil
-                      :new-session t))
+                      :new-session t
+                      :session-meta session-meta))
 
 (cl-defun agent-shell--config-icon (&key config)
   "Create icon string for CONFIG if available and icons are enabled.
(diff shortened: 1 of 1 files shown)


xenodium:
I'm wondering how to scale this a bit better, so we don't keep injecting parameters every time we watn to customize a request. I'm thinking perhaps we can use request-decorating functions that receive and return a request, giving you the opportunity to customize the request before it's sent out. Lemme think a bit more about this. Would a solution along these lines work for your purpose?

ElleNajt:
Yeah that would be good! Mostly I just want to be able to pass other parameters to claude-code-acp.

ElleNajt:
That makes sense as a general direction. For this specific case though, `_meta` is the ACP protocol's [built-in extensibility mechanism](https://agentclientprotocol.com/protocol/extensibility.md) — it's on every request type specifically so clients can pass through opaque custom data without the protocol needing to know about it. There's also an [RFD on meta propagation](https://agentclientprotocol.com/rfds/meta-propagation.md) formalizing conventions for it.

So this isn't really adding a new parameter so much as exposing the one the protocol already provides for this purpose. A request-decorator approach could work too, but it might be more machinery than needed for forwarding a protocol-level field. Happy to adapt either way if you have a specific API in mind.

---
*Written by Claude*

xenodium:
Please check out https://github.com/xenodium/agent-shell/commit/3b443f2f2187820b08f33ec2209e9dbe22849302 which introduces an `:outgoing-request-decorator`, a function or lambda taking a request and returning a decorated request. You'll need to update acp.el also.

> or this specific case though, _meta is the ACP protocol's [built-in extensibility mechanism](https://agentclientprotocol.com/protocol/extensibility.md) — it's on every request type specifically so clients can pass through opaque custom data without the protocol needing to know about it. There's also an [RFD on meta propagation](https://agentclientprotocol.com/rfds/meta-propagation.md) formalizing conventions for it.

Sure, but that's not what this PR is offering. The PR adds a parameter to agent-shell that appends _meta onlyfor some requests. What do we do when we want to add _meta elsewhere? We'd have to add another parameter for those instances.

According to https://agentclientprotocol.com/protocol/extensibility.md:

> All types in the protocol include a `_meta` field with type `{ [key: string]: unknown }` that implementations can use to attach custom information. This includes requests, responses, notifications, and even nested types like content blocks, tool calls, plan entries, and capability objects.

> Written by Claude

@ElleNajt Please talk to me instead 😭 I still prefer talking to to humans, specially if I'm doing something for them.

ElleNajt:
Sry! Was really busy yesterday and just wanted to move this PR forward. Will do.

ElleNajt:
I'm trying to find the right social balance between Claude doing research for me and (tbqh) understanding the code better than I do, and therefore endorsing Claude's statements after I've cross examined it's conclusions (which is admittedly doing a lot of work, and degree of cross examination varies from asking it questions until I'm satisfied to following the logic), but also not framing everything Claude says as stuff that I'm saying. 

Genuinely unsure of what the right norms are here -- I think it probably depends on what you want since this your project. But since you prefer it happy to write messages here directly.

Perhaps the right move in these situations is to cite what the robot found and give my level of endorsement? :)

ElleNajt:
Claude makes the point that git commit + PRs are deliverables and that a PR conversation is a human relationship, and so it makes sense to draw a line there around what Claude gets to write. Which is an interesting point. (Also perhaps we should take this norms conversation elsewhere?)

ElleNajt:
On topic:

> Sure, but that's not what this PR is offering. The PR adds a parameter to agent-shell that appends _meta onlyfor some requests. What do we do when we want to add _meta elsewhere? 

Ty, objection makes sense to me now!

And ty for the update! 🙏♥️

xenodium:
> Perhaps the right move in these situations is to cite what the robot found and give my level of endorsement? :)

Sure. As long as you're citing the lightweight version of robot output and giving me your level of endorsement (after throroughly reviewing yourself). I (human) need to read whatever you send my way and give you my brain cycles.

xenodium:
> Also perhaps we should take this norms conversation elsewhere?

Happy to add it to github PR/issue templates, stating my preference, if ya reckon that's the right place.

ElleNajt:
Yeah I think that's a good place. You could also sneak it in a claud.md and/or an agents.md in this repo so that the robots read it too.

xenodium:
Good idea. I've yet to leverage these much. Thanks.

xenodium:
> Yeah I think that's a good place. You could also sneak it in a claud.md and/or an agents.md in this repo so that the robots read it too.

Thanks for this. https://github.com/xenodium/agent-shell/commit/74c58741eda2e6a2a1f82f2bc361a670c24d8902 should be a start.

xenodium:
Gonna close PR since these two should handle the request:

- https://github.com/xenodium/agent-shell/commit/3b443f2f2187820b08f33ec2209e9dbe22849302
- https://github.com/xenodium/acp.el/commit/5dcac092cfd8bf0eefe1ce759202f74d605e716b

Please ping if otherwise.

PR #285, opened 2026-02-13
add agent-shell-send-clipboard-image

dangom:
## Checklist

I noticed on MacOS that the screenshot feature doesn't really work because of screen recording permission issues. Also, sometimes I already have an image on the clipboard which I'd like to send, instead of taking another screenshot or saving to file.
This PR introduces agent-shell-send-clipboard-image, which will use pngpaste to add an image from clipboard to the prompt. I tested this on my end and it works fine, though I didn't write any tests. Code is similar to agent-shell-send-screenshot.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -295,6 +295,36 @@ Assume screenshot file path will be appended to this list."
   :type '(repeat string)
   :group 'agent-shell)
 
+(defcustom agent-shell-clipboard-image-handlers
+  (list
+   (list (cons :command "pngpaste")
+         (cons :save (lambda (file-path)
+                       (let ((exit-code (call-process "pngpaste" nil nil nil file-path)))
+                         (unless (zerop exit-code)
+                           (error "pngpaste failed with exit code %d" exit-code))))))
+   (list (cons :command "xclip")
+         (cons :save (lambda (file-path)
+                       (with-temp-buffer
+                         (set-buffer-multibyte nil)
+                         (let ((exit-code (call-process "xclip" nil t nil
+                                                        "-selection" "clipboard"
+                                                        "-t" "image/png" "-o")))
+                           (unless (zerop exit-code)
+                             (error "xclip failed with exit code %d" exit-code))
+                           (write-region (point-min) (point-max) file-path nil 'silent)))))))
+  "Handlers for saving clipboard images to a file.
+
+Each handler is an alist with the following keys:
+
+  :command  The executable name to look up via `executable-find'.
+  :save     A function taking FILE-PATH that saves the clipboard
+            image there, signaling an error on failure.
+
+Handlers are tried in order.  The first whose :command is found
+on the system is used."
+  :type '(repeat (alist :key-type symbol :value-type sexp))
+  :group 'agent-shell)
+
 (defcustom agent-shell-buffer-name-format 'default
   "Format to use when generating agent shell buffer names.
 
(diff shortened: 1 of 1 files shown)


xenodium:
Thank you for the PR! Nice idea.

Looks like we're introducing this great way of making the things configurable, while relying on platform checks. How about we base most of the logic checks on whether or not a utility exists instead (can use executable-find for this sort of thing)?

I'm thinking we can create a registry list of sorts and resolve the command based on whether or not utility is found. Bails if none found.

I've not tested this code, but I'm thinking something like:

```emacs-lisp
(defcustom agent-shell-clipboard-image-handlers
  (list
   (list (cons :utility "pngpaste")
         (cons :save (lambda (file-path)
                       (let ((exit-code (call-process "pngpaste" nil nil nil file-path)))
                         (unless (zerop exit-code)
                           (error "pngpaste failed with exit code %d" exit-code))))))
   (list (cons :utility "xclip")
         (cons :save (lambda (file-path)
                       (with-temp-buffer
                         (set-buffer-multibyte nil)
                         (let ((exit-code (call-process "xclip" nil t nil
                                                        "-selection" "clipboard"
                                                        "-t" "image/png" "-o")))
                           (unless (zerop exit-code)
                             (error "xclip failed with exit code %d" exit-code))
                           (write-region (point-min) (point-max) file-path nil 'silent)))))))
  "Handlers for saving clipboard images to a file.

Each handler is an alist with the following keys:

  :utility  The executable name to look up via `executable-find'.
  :save     A function taking FILE-PATH that saves the clipboard
            image there, signaling an error on failure.

Handlers are tried in order. The first whose :utility is found
on the system is used."
  :type '(repeat (alist :key-type symbol :value-type sexp))
  :group 'agent-shell)
```

This makes agent-shell--save-clipboard-image agnostic of the utility as follows:

```emacs-lisp
(cl-defun agent-shell--save-clipboard-image (&key destination-dir)
  "Save clipboard image to DESTINATION-DIR.
Returns the full path to the saved image file on success."
  (unless destination-dir
    (error "Destination-dir is required"))
  (let* ((file-path (expand-file-name
                     (format "clipboard-%s.png"
                             (format-time-string "%Y%m%d-%H%M%S"))
                     destination-dir))
         (handler (seq-find
                   (lambda (h)
                     (executable-find (map-elt h :utility)))
                   agent-shell-clipboard-image-handlers)))
    (unless handler
      (error "No clipboard image utility found (tried: %s)"
             (mapconcat (lambda (h) (map-elt h :utility))
                        agent-shell-clipboard-image-handlers ", ")))
    (unless (file-directory-p destination-dir)
      (make-directory destination-dir t))
    (funcall (map-elt handler :save) file-path)
    (cond
     ((not (file-exists-p file-path))
      (error "Clipboard image file was not created"))
     ((zerop (nth 7 (file-attributes file-path)))
      (delete-file file-path)
      (error "No image found in clipboard"))
     (t
      file-path))))
```

As I said, I've not tested this and may need cleanup, but may be a solution.... What do you think?

dangom:
Nice. I just changed `:utility` to `:command` since that seemed more emacsonic.

xenodium:
> Nice. I just changed :utility to :command since that seemed more emacsonic.

Awesome. Looks great. Thank you!

xenodium:
Heads-up, I've extended things so we can now make yanking smarter via agent-shell-yank-dwim https://github.com/xenodium/agent-shell/commit/565943f33c9a029444e94567a1f0fe54082c71a9

The agent-shell editable modes now remap yank to agent-shell-yank-dwim, so we no longer need to explicitly invoke agent-shell-send-clipboard-image. Just yank/paste as usual and the image appears in the buffer. Typically via C-y.

PR #288, opened 2026-02-14
Add experimental quiet mode for compact turn display

ewilderj:
## Summary

This is an **EXPERIMENTAL** mode to show what it would look like to conceal most of the steps between thoughts. In long running turns this shows only the most important agent thoughts and reduces visual noise considerably.

## What it does

When `agent-shell-quiet-mode` is enabled (`t`), all thought process and tool call fragments within a turn are grouped under collapsible sections instead of being shown individually. Each section:

- Shows the agent's current thought as a one-line label
- Animates a braille spinner (⠋⠙⠹…) while work is in progress
- Finalizes with ✓ when the phase completes
- Expands on click to reveal the individual tool calls underneath

When a new thought arrives after tool calls, a new collapsible group starts automatically — so long-running turns show multiple phases:

```
▶ ✓ Exploring the codebase...
▶ ⠹ Found the issue, now fixing...
```

## Implementation

- New file `agent-shell-quiet.el` with all quiet mode logic
- Uses `agent-shell-ui` fragments: a wrapper fragment per phase with child fragments for tool calls
- Spinner timer and state stored per-group on the state alist (multi-session safe)
- Late-completing tool calls are correctly associated with their originating group
- Toggle (expand/collapse) works via advice on `agent-shell-ui-toggle-fragment-at-point`

## Usage

```elisp
(setq agent-shell-quiet-mode t)
```

No other configuration needed. Set to `nil` to return to normal display.

### DIFF
--- a/agent-shell-quiet.el
+++ b/agent-shell-quiet.el
@@ -1,0 +1,259 @@
+;;; agent-shell-quiet.el --- Quiet mode for compact turn display -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2024 Alvaro Ramirez
+
+;; Author: Alvaro Ramirez https://xenodium.com
+;; URL: https://github.com/xenodium/agent-shell
+
+;; This package is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+
+;; This package is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;
+;; Report issues at https://github.com/xenodium/agent-shell/issues
+;;
+;; ✨ Please support this work https://github.com/sponsors/xenodium ✨
+
+;;; Commentary:
+;;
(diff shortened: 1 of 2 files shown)


ewilderj:
Here's a before/after comparison

BEFORE
image

(and goes on for 2.5 more screenfuls)

AFTER
image

ewilderj:
Noting again how experimental this is: it becomes quite vulnerable to the various whims of how each model outputs thinking tokens, and whether it streams or not. Pretty sure I found some upstream bugs for Copilot CLI. Still, am trying to make it do a good job on messy data.

xenodium:
Thanks for this. It looks great.

> Noting again how experimental this is: it becomes quite vulnerable to the various whims of how each model outputs thinking tokens, and whether it streams or not.

Yeah. I suspect it'd become pretty challenging for me to maintain it as we now have 12 agents and we already get subtle differences across agents that break UX in unique ways. For a more involved customization like this, I'd be more inclined to offer whatever hooks you need so it can be implemented as a separate package. We got quite a few related packages now https://github.com/xenodium/agent-shell?tab=readme-ov-file#related-projects

ewilderj:
Following up on your suggestion — I've extracted quiet-mode as a standalone package: https://github.com/ewilderj/agent-shell-quiet-mode

It works today via advice on four internal functions. I've opened #324 with a proposal for the hooks that would make this (and other add-ons) more robust:

1. **Notification hooks** — per-event-type hooks for `tool_call`, `agent_thought_chunk`, `tool_call_update` (with ability to suppress default handling)
2. **Turn completion hook** — fires when a turn ends
3. **State init hook** — lets add-ons extend the state alist
4. **Fragment toggle hook** — fires after a fragment is collapsed/expanded

Happy to help implement any of these upstream. Closing this PR since the separate package approach is the way forward.

Issue #254, opened 2026-02-02
Remove leading two spaces from each output line


HIRANO-Satoshi:
Each output line starts with two leading spaces. When I paste the output into emails or documents, I have to remove these spaces line by line, which is inconvenient.

According to the ACP log, the LLM output does not contain these leading spaces. This means the two spaces at the beginning of each line are being added by the code, not by the LLM.

If possible, it would be great to avoid adding leading spaces by default (or make it configurable).

## Checklist

- [X] I've read the README's [Issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#issues) section.
- [X] I'm running the latest agent-shell version.

xenodium:
https://github.com/xenodium/agent-shell/commit/9d72ff10c2c3d06b523d6ed1224ebecf330021f7 makes the indent visual so when you copy the text, the indent is not part of the text. Please update to v0.33.2 and try it out.

HIRANO-Satoshi:
I confirmed. It's good.

Thank you very much!

xenodium:
Good to hear. Thanks!

ultronozm:
With the change you mention here, when I kill/yank part of a code block into another file buffer, the text appears indented two spaces in display, even though leading space characters are not actually present.

This is because the line-prefix/wrap-prefix text properties are preserved by default under kill/yank.  The following patch fixes things for me (happy to send as a PR, or please feel free to apply directly, etc):

```diff
diff --git a/agent-shell.el b/agent-shell.el
index b5e178f..12e6907 100644
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -734,6 +734,19 @@ When FORCE is non-nil, skip confirmation prompt."
       :command command
       :shell shell))))
 
+(defun agent-shell--filter-buffer-substring (start end &optional delete)
+  "Return the buffer substring between BEG and END, after filtering.
+Strip `line-prefix' and `wrap-prefix' from the copied substring.  If
+DELETE is non-nil, delete the text between BEG and END from the buffer."
+  (let ((text (if delete
+                  (prog1 (buffer-substring start end)
+                    (delete-region start end))
+                (buffer-substring start end))))
+    (remove-text-properties 0 (length text)
+                            '(line-prefix nil wrap-prefix nil)
+                            text)
+    text))
+
 (defvar-keymap agent-shell-mode-map
   :parent shell-maker-mode-map
   :doc "Keymap for `agent-shell-mode'."
[...]

PR #292, opened 2026-02-16
Strip text properties from copied agent-shell buffer text

ultronozm:
* agent-shell.el (agent-shell--filter-buffer-substring): New function.
(agent-shell--start): Set buffer-local
filter-buffer-substring-function to use custom filter.  This ensures
copied text is clean of display properties, such as 'line-prefix' and
'wrap-prefix'.

Fixes an error introduced by the fix for #254.


xenodium:
Thank you!

Issue #301, opened 2026-02-18
[FR] Support configuring default model by name


lgfang:
### Problem

Currently, `agent-shell-anthropic-default-model-id` requires specifying a
technical model ID string (e.g., `anthropic.claude-opus-4-5-20251101-v1:0`).
These IDs are hard to remember, not user-friendly, and change across model
versions.

Users have to start a session first, inspect the available models, copy the 
exact model ID, then put it in their config. This is cumbersome compared to
just writing `"Opus"`.

### Proposal

Add a new custom variable `agent-shell-anthropic-default-model-name` that
accepts a human-readable model name (e.g., `"Opus"`, `"Sonnet"`, `"Haiku"`).

After session initialization (when the available model list is known), the
name is resolved to the corresponding model ID via case-insensitive matching.

lgfang:
https://github.com/xenodium/agent-shell/pull/302 implements this feature

xenodium:
Thank you for filing a feature request and offering a PR.

I introduced `agent-shell-anthropic-default-model-id` to overcome a pretty big limitation we had. That is, the need to submit a prompt before you could change or set things. I'm actually looking to deprecate that variable in favor of a new flow that does bootstrap shells before offering prompt input. This gives you the ability to use `agent-shell-cycle-session-mode` and `agent-shell-set-session-model` (with completion) before you send any prompt.

You can enable the new flow with either:

Closest to the current behaviour:

```emacs-lisp
(setq agent-shell-session-strategy 'new)
```

Enable resuming sessions:

```emacs-lisp
(setq agent-shell-session-strategy 'prompt)
```

Now that we can bootstratp shell sessions, I'll be looking into remembering the last model and mode used. Having said that, this is likely not as desperately needed as before, when we could not easily change models/modes before submitting a prompt.

Please give the new flows a try. I haven't communicated these too widely just yet as they are very new and may still have rough edges. Please report issues.

lgfang:
Thank you for the prompt response. I've tested `agent-shell-session-strategy` set to `'prompt` or `'new`, but the new workflow doesn't fully address the core issues:

1. New sessions default to the base model, requiring manual selection each time (I understand this will be addressed by remembering the last model/mode used)
2. Model updates still require manual intervention to switch to newer releases

This feature request solves both problems: specifying a model name ensures all sessions automatically use the latest release of the target model.

Please let me know if I missed anything.

xenodium:
> Please let me know if I missed anything.

[...]

PR #302, opened 2026-02-18
feat: Add support for model configuration by name

lgfang:
Previously, users had to configure the default Anthropic model using technical model IDs like "us.anthropic.claude-opus-4-5-20251101-v1:0", which are:
- Hard to remember and type correctly
- Not user-friendly
- Subject to change across model versions
- Difficult to discover without looking at server responses

This change allows users to configure models using human-readable names like "Opus", "Sonnet", or "Haiku" instead, making the configuration more intuitive and maintainable.

- Add `agent-shell-anthropic-default-model-name` custom variable for specifying models by name (e.g., "Opus", "Sonnet", "Haiku")
- Implement model name resolution that maps names to IDs after session initialization when available models are known
- Add validation to prevent confusing configurations where both name and ID are set - raises clear error messages
- Handle deferred model setting for name-based configuration, retrying after session initialization when models are available
- Add comprehensive tests for model name resolution and error cases

Users can now configure:
```elisp
(setq agent-shell-anthropic-default-model-name "Opus")
```

Instead of:
```elisp
(setq agent-shell-anthropic-default-model-id "claude-3-6-opus-20250109")
```

The implementation handles case-insensitive matching and provides clear error messages when models cannot be found or when configuration is invalid.

### DIFF
--- a/agent-shell-anthropic.el
+++ b/agent-shell-anthropic.el
@@ -80,6 +80,18 @@ when starting a new shell."
   :type '(choice (const nil) string)
   :group 'agent-shell)
 
+(defcustom agent-shell-anthropic-default-model-name
+  nil
+  "Default Anthropic model name.
+
+Must be one of the model names displayed under \"Available models\"
+when starting a new shell. For example: \"Opus\", \"Sonnet\", \"Haiku\".
+
+Cannot be used together with `agent-shell-anthropic-default-model-id'.
+If both are set, an error will be raised."
+  :type '(choice (const nil) string)
+  :group 'agent-shell)
+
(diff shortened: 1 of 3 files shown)


xenodium:
Thank you for the PR! Have a look at this comment https://github.com/xenodium/agent-shell/issues/301#issuecomment-3919457426 We got a new flow that will be deprecating `agent-shell-anthropic-default-model-id` in favor of bootstrapped shell sessions.

lgfang:
As agreed in the corresponding feature request, making `agent-shell-anthropic-default-model-id` accept a function is a better solution. Hence, closing this PR in favour of that direction.

PR #308, opened 2026-02-18
Fix heartbeat nil value crash in timer and busy indicator

ElleNajt:
## Summary

- Guard against nil `:heartbeat :value` in the timer lambda and busy indicator frame function

## Problem

`agent-shell-heartbeat-stop` sets `:value` to nil, but `cancel-timer` doesn't prevent an already-dispatched timer tick from running. When that final tick fires, it calls `(1+ nil)`, producing:

```
(wrong-type-argument number-or-marker-p nil)
```

The same nil value propagates to `agent-shell--busy-indicator-frame` via mode-line redisplay during the `force-mode-line-update` call, triggering the error there too. This results in repeated error spam in `*Messages*`.

## Changes

- `agent-shell-heartbeat.el`: Wrap timer lambda body in `when-let` so it skips the tick when value is nil
- `agent-shell.el`: Add `value` binding to `when-let*` in `agent-shell--busy-indicator-frame` so `mod` never receives nil

## Test plan

- [ ] Start an agent-shell session and let it run (busy indicator active)
- [ ] Interrupt/stop the session and verify no `wrong-type-argument` errors in `*Messages*`

---
*Written by Claude*

### DIFF
--- a/agent-shell-heartbeat.el
+++ b/agent-shell-heartbeat.el
@@ -84,12 +84,12 @@ Returns the heartbeat alist with the timer started."
       (map-put! heartbeat :heartbeat-timer
                 (run-at-time 0 (/ 1.0 (map-elt heartbeat :beats-per-second))
                              (lambda ()
-                               (map-put! heartbeat :status 'busy)
-                               (map-put! heartbeat :value
-                                         (1+ (map-elt heartbeat :value)))
-                               (funcall (map-elt heartbeat :on-heartbeat)
-                                        (map-elt heartbeat :value)
-                                        (map-elt heartbeat :status))))))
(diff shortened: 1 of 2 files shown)


xenodium:
Still keen to submit this? Marked as draft still.

ElleNajt:
I'm not running into this problem anymore (I always have  agent-shell-show-busy-indicator nil) The code looks harmless to merge and does protect against a plausible bug; so sure, let's merge it to not think about it anymore?

ElleNajt:
OTOH don't mind closing it. Up to you. :)

xenodium:
>  let's merge it to not think about it anymore?

Agreed. Done.

Issue #311, opened 2026-02-20
[FR] agent-shell-anthropic-default-model-id should accept a function


lgfang:
## Summary

  As discussed in #301, allow `agent-shell-anthropic-default-model-id` to accept a function in addition to a string. This enables users to implement custom model resolution logic (e.g., resolving a human-readable model name to a model ID) without
  requiring changes to agent-shell itself.

  ## Motivation

  Currently `agent-shell-anthropic-default-model-id` only accepts a string model ID. These IDs are hard to remember, change across versions, and vary by provider (e.g., `us.anthropic.claude-opus-4-6-v1` on AWS Bedrock). A function value lets users write
  their own resolution logic that runs at session bootstrap time when the available models are known.

  ## Proposed changes

  1. Update `agent-shell-anthropic-default-model-id` custom type to accept `nil`, `string`, or `function`.
  2. Update the `:default-model-id` lambda in `agent-shell-anthropic-make-claude-code-config` to call the value with `funcall` when it is a function.

  ## Example usage

  Users can configure a model by name in their init file:

  ```elisp
  (setq agent-shell-anthropic-default-model-id
        (lambda ()
          (when-let* ((models (and (boundp 'agent-shell--state)
                                   (map-nested-elt agent-shell--state '(:session :models))))
                      (model (seq-find (lambda (m)
                                         (string-equal (downcase (or (map-elt m :name) ""))
                                                       (downcase "Opus")))
                                       models)))
            (map-elt model :model-id))))
```

This resolves "Opus" to the actual model ID at session bootstrap time using case-insensitive matching against the session's available models.

Existing string-based configuration continues to work unchanged.

xenodium:
https://github.com/xenodium/agent-shell/pull/312 now merged

PR #312, opened 2026-02-20
Accept function value for agent-shell-anthropic-default-model-id #311

lgfang:
Previously, `agent-shell-anthropic-default-model-id` only accepted a string, requiring users to specify model IDs like
"us.anthropic.claude-opus-4-6-v1".

This change allows a function value, enabling users to implement custom resolution logic.  For example, resolving a human-readable model name to a model ID using the session's available models:

```elisp
  (setq agent-shell-anthropic-default-model-id
        (lambda ()
          (when-let* ((models (and (boundp 'agent-shell--state)
                                   (map-nested-elt agent-shell--state
                                                   '(:session :models))))
                      (model (seq-find
                              (lambda (m)
                                (string-equal
                                 (downcase (or (map-elt m :name) ""))
                                 (downcase "Opus")))
                              models)))
            (map-elt model :model-id))))
```

### DIFF
--- a/agent-shell-anthropic.el
+++ b/agent-shell-anthropic.el
@@ -76,8 +76,10 @@ For api key:
   "Default Anthropic model ID.
 
 Must be one of the model ID's displayed under \"Available models\"
-when starting a new shell."
-  :type '(choice (const nil) string)
+when starting a new shell.
+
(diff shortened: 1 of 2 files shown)


xenodium:
Thank you for the PR

PR #313, opened 2026-02-20
prompt-select-session preserves sorting

catern:
agent-shell--prompt-select-session now has completion table metadata such that it won't be sorted alphabetically by completion frontends which check display-sort-function (such as the default completion frontend), and sets eager-display and eager-update to reflect that it's a small list of candidates on which completion is cheap. (these are new in Emacs 31)

The "Start a new session" candidate is now at the end of the candidate list.  It's still the default, so a user can still select it easily by pressing RET, but now they more easily can select the candidates at the start of the candidate list, using whatever methods are specific to their completion frontend.

## Checklist

- [] I've run `M-x checkdoc` and `M-x byte-compile-file`.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -3372,11 +3372,12 @@ Falls back to latest session in batch mode (e.g. tests)."
                                                    (length (agent-shell--session-title s)))
                                                  acp-sessions)))
            (new-session-choice "Start a new session")
-           (choices (cons (cons new-session-choice nil)
-                          (mapcar (lambda (acp-session)
-                                    (cons (agent-shell--session-choice-label acp-session max-dir-width max-title-width)
-                                          acp-session))
-                                  acp-sessions)))
+           (choices (append
+                     (mapcar (lambda (acp-session)
+                               (cons (agent-shell--session-choice-label acp-session max-dir-width max-title-width)
+                                     acp-session))
+                             acp-sessions)
+                     (list (cons new-session-choice nil))))
            (candidates (mapcar #'car choices))
            ;; Some completion frameworks yielded appended (nil) to each line
            ;; unless this-command was bound.
@@ -3387,8 +3388,14 @@ Falls back to latest session in batch mode (e.g. tests)."
            ;; Let's optimize the rocket engine      Feb 12, 21:02 (nil)
            (this-command 'agent-shell))
       (agent-shell--emit-event :event 'session-prompt)
-      (let ((selection (completing-read "Resume session: "
-                                        candidates
+      (let ((selection (completing-read "Resume session (default: start new): "
+                                        (lambda (string pred action)
+                                          (if (eq action 'metadata)
+                                              '(metadata
+                                                (display-sort-function . identity)
+                                                (eager-display . t)
+                                                (eager-update . t))
+                                            (complete-with-action action candidates string pred)))
                                         nil t nil nil
                                         new-session-choice)))
         (map-elt choices selection))))))


agent-shell.el
```diff
@@ -3372,11 +3372,12 @@ Falls back to latest session in batch mode (e.g. tests)."
                                                    (length (agent-shell--session-title s)))
                                                  acp-sessions)))
            (new-session-choice "Start a new session")
-           (choices (cons (cons new-session-choice nil)
```
xenodium, line 3464: I was getting:

```emacs-lisp
Debugger entered--Lisp error: (wrong-type-argument listp "Start a new session")
  car("Start a new session")
  mapcar(car ((... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ...) "Start a new session"))
  (let* ((max-dir-width (apply #'max (mapcar #'(lambda (s) (length ...)) acp-sessions))) (max-title-width (apply #'max (mapcar #'(lambda (s) (length ...)) acp-sessions))) (new-session-choice "Start a new session") (choices (cons (mapcar #'(lambda (acp-session) (cons ... acp-session)) acp-sessions) (cons new-session-choice nil))) (candidates (mapcar #'car choices)) (this-command 'agent-shell)) (agent-shell--emit-event :event 'session-prompt) (let ((selection (completing-read "Resume session (default: start new): " (completion-table-with-metadata candidates '(... ... ...)) nil t nil nil new-session-choice))) (map-elt choices selection)))
```

Fixed with something like:

```emacs-lisp
           (choices (append
                     (mapcar (lambda (acp-session)
                               (cons (agent-shell--session-choice-label acp-session max-dir-width max-title-width)
                                     acp-session))
                             acp-sessions)
                     (list (cons new-session-choice nil))))
```

Coincidentally, and I know it's a nit/subjective but https://github.com/xenodium/agent-shell?tab=readme-ov-file#cl-lib-limited-to-cl-defun Sorry :)

agent-shell.el
```diff
@@ -3387,8 +3388,12 @@ Falls back to latest session in batch mode (e.g. tests)."
            ;; Let's optimize the rocket engine      Feb 12, 21:02 (nil)
            (this-command 'agent-shell))
       (agent-shell--emit-event :event 'session-prompt)
-      (let ((selection (completing-read "Resume session: "
-                                        candidates
+      (let ((selection (completing-read "Resume session (default: start new): "
+                                        (completion-table-with-metadata
```
xenodium: agent-shell.el:3392:42: Warning: the function ‘completion-table-with-metadata’
    is not known to be defined.

Is this new? Are you building Emacs from main? How about using something backwards compatible like:

```emacs-lisp
                                  (lambda (string pred action)
                                          (if (eq action 'metadata)
                                              '(metadata
                                                (display-sort-function . identity)
                                                (eager-display . t)
                                                (eager-update . t))
                                            (complete-with-action action candidates string pred)))
```

Does this work on your instance too?

xenodium (commented):
Thanks for the PR! The `Resume session (default: start new): ` feels more idiomatic. Left a couple of comments.

agent-shell.el
catern: Certainly, will use append, not a problem!

agent-shell.el
catern: Oops, yes I guess it's new in 31, sorry about that, a lambda is indeed better.

xenodium:
Thanks for the changes. Merged.

Heads-up, I may look into further changes. While I like the `(default: start new)` prompt, moving the "start new" to last means that when selected by default in a vertical completion framework (mine is ivy), the last item is naturally selected (but also scrolled). As a result, I now have a view of the oldest created sessions (and lose view of the most recent ones) by default.

xenodium:
Hmmm... feels a little strange for the default option to be the last option.

I've pushed https://github.com/xenodium/agent-shell/commit/5ec181ef7ab451bd34d1e6e21141347c33cbb750 bringing the new shell option back to top, but I'm hoping with the introduction of display-sort-function still remains compatible with your flow.

PR #314, opened 2026-02-20
Add agent-shell-start-in-project

catern:
This makes agent-shell configurable to behave more like project-shell or other commands, which don't prompt to start a new shell, they just create it.  (I personally think `t` should be the default but I'm guessing that breaks some use case?)

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -3916,6 +3916,19 @@ If FILE-PATH is not an image, returns nil."
                            (agent-shell-cwd))))
                 (agent-shell-buffers))))
 
+(defcustom agent-shell-new-shell-strategy 'prompt
+  "How to handle starting a shell when none exist for the current project.
+
+Available values:
+
+  `new': Always start a new shell immediately.
+  `prompt': Prompt whether to start a new shell.
+  `switch': Skip creation and prompt to switch to an existing shell."
+  :type '(choice (const :tag "Always start new shell" new)
+                 (const :tag "Prompt to start new shell" prompt)
+                 (const :tag "Switch to existing shell" switch))
+  :group 'agent-shell)
+
 (cl-defun agent-shell--shell-buffer (&key viewport-buffer no-error no-create)
   "Get an `agent-shell' buffer for the current project.
 
@@ -3941,7 +3954,14 @@ Returns a buffer object or nil."
       (if no-create
           (unless no-error
             (user-error "No agent shell buffers available for current project"))
-        (if (y-or-n-p "No shells in project.  Start a new one? ")
+        (if (pcase agent-shell-new-shell-strategy
+              ('prompt (y-or-n-p "No shells in project.  Start a new one? "))
+              ('new t)
+              ('switch nil)
+              (_
+               (message "Unknown value of `agent-shell-new-shell-strategy' '%s', starting new shell."
+                        agent-shell-session-strategy)
+               t))
             (get-buffer
              (agent-shell--start :config (or (agent-shell--resolve-preferred-config)
                                              (agent-shell-select-config


agent-shell.el
```diff
@@ -3916,6 +3916,19 @@ If FILE-PATH is not an image, returns nil."
                            (agent-shell-cwd))))
                 (agent-shell-buffers))))
 
+(defcustom agent-shell-start-in-project 'query
```
xenodium: Since we have `agent-shell-session-strategy` (and `'prompt`), how about naming something like this?

```emacs
(defcustom agent-shell-new-shell-strategy 'prompt
  "How to handle starting a shell when none exist for the current project.

Available values:

  `new': Always start a new shell immediately.
  `prompt': Prompt whether to start a new shell.
  `switch': Skip creation and prompt to switch to an existing shell."
  :type '(choice (const :tag "Always start new shell" new)
                 (const :tag "Prompt to start new shell" prompt)
                 (const :tag "Switch to existing shell" switch))
  :group 'agent-shell)
```

catern:
I see... that's an interesting use case.

So you might be in a file in one project, but want to send the current line or region as context to another project...

If there were two separate commands for "send context" and "open shell", then this wouldn't be an issue, but of course that's less nice.

I can see why this would be useful, so maybe this PR doesn't make sense.

Still, my reason for this PR is just that it's annoying for M-x agent-shell to prompt you twice in a row:
1. First a prompt about whether to start a new shell or switch to an existing one
2. Then a prompt about whether to start a new session or resume a previous one (when agent-shell-session-strategy is 'prompt)

What if we merged these prompts somewhat?  So it was instead:
1. A prompt about whether to A. start a new session (in a new shell), B. resume a previous session or C. switch to an existing shell in another project
2. If you chose A, no prompt; B, a prompt for a previous session; C, a prompt for an existing shell in another project.

Also, if the user selects A, they could get the agent-shell-session-strategy=new-deferred behavior (maybe based on their configuration?), which is maybe a little nicer.

That would also make session resumption a bit better-integrated: this triple prompt could be available by default.

A more extreme idea would be to merge 1 and 2 entirely, so there's a single prompt where you can choose between a list of existing shells, a list of previous sessions, and "start a new session".  That would mean eagerly starting up an ACP server in the project, and shutting it down if the user selects an existing shell.  That might be over-engineering things for now, but maybe it's not too crazy for the future.

xenodium:
> If there were two separate commands for "send context" and "open shell", then this wouldn't be an issue, but of course that's less nice.

We have those too:

- `M-x agent-shell-*send*-to`
- `M-x agent-shell-new-shell`

While I'm a fan of DWIM behaviour, it's not super discoverable so I often offer explicit alternatives like these.

> Still, my reason for this PR is just that it's annoying for M-x agent-shell to prompt you twice in a row:
> 
>     1. First a prompt about whether to start a new shell or switch to an existing one
> 
>     2. Then a prompt about whether to start a new session or resume a previous one (when agent-shell-session-strategy is 'prompt)

Ah interesting. Is creating new shells via `M-x agent-shell` a frequent event for you? When I know I'm definitely going to need a new shell, I either invoke:

1. `C-u M-x agent-shell`
2. `M-x agent-shell-new-shell`

There's also `M-x agent-shell-anthropic-start-claude-code` or whatever other agent equivalent. Been meaning to rename those to something like `M-x agent-shell-anthropic-*new*-claude-code`

Would any these work for you?

> If you chose A, no prompt; B, a prompt for a previous session; C, a prompt for an existing shell in another project.

I think this is doable, but would like to have play with it first to see what edge cases it unearths. Lemme have a play and see...

> if the user selects A, they could get the agent-shell-session-strategy=new-deferred behavior (maybe based on their configuration?)

With `agent-shell-session-strategy` and `agent-shell-prefer-session-resume` already in place, I have to be mindful of offering further configuration in the space. Since introducing bootstapped sessions and resuming, I'd like to try to simplify the whole bootstrapping/handshake internally while also offering the best possible default experience. That may mean dropping `new-deferred` altogether and figure out an alternate way offering a writeble prompt ASAP. The `new-deferred` option is currently there for backwards compatibility, but the deferred flow also invited quite a few user requests for bootrapped sessions (to enable switching mode/models prior to submitting prompts).

I'll have a play with consolidating prompts and see where it takes us.

xenodium:
> I'll have a play with consolidating prompts and see where it takes us.

Been experimenting with things.

2026-02-23-18:29:09-Emacs_optimized

This may be getting us pretty close to what we want (minus new-deferred). If we want to go further about enabling the shell to receive input wihle still handshaking, we'll have to start looking into the bigger task of replacing comint with a more custom major mode.

catern:
>Ah interesting. Is creating new shells via M-x agent-shell a frequent event for you?

Yes.  Or, more specifically, I'm used to a single command which brings me to the shell buffer for the current project, creating that buffer if it doesn't exist - that's how e.g. project-shell works.  Having a single entry-point command like this seems like it makes things nice and easy for users.

>With agent-shell-session-strategy and agent-shell-prefer-session-resume already in place, I have to be mindful of offering further configuration in the space.

Fully agreed, adding too much config here will cause problems.

>That may mean dropping new-deferred altogether and figure out an alternate way offering a writeble prompt ASAP. The new-deferred option is currently there for backwards compatibility, but the deferred flow also invited quite a few user requests for bootrapped sessions (to enable switching mode/models prior to submitting prompts).

Ah I didn't realize that getting rid of new-deferred is the plan.  That makes sense to me.

And your demo looks great!  I think that's pretty much exactly what I want.  A few thoughts:

- I think it would be a bit nicer to have the "Open an existing shell" option at the end, maybe?  Pressing <down> two (ivy) or three (non-ivy) times to get to the most recent session seems a bit painful.
- I'm guessing that when agent-shell-preferred-agent-config isn't set, agent-shell (if there's no current shell) would first prompt for an agent type, then prompt like in your demo?  If so, that seems reasonable to me.

xenodium:
> I think it would be a bit nicer to have the "Open an existing shell" option at the end, maybe?

The challenge here is that you and I will know about the "Open an existing shell" option hiding in the end of the list, but it won't be discoverable to users. Users will have many sessions accumulated, specially since resuming is a new agent-shell feature.

> Pressing two (ivy) or three (non-ivy) times to get to the most recent session seems a bit painful.

Can be done in two ways. Either fuzzy type to match the desired session or indeed `C-n C-n`. Neither felt particularly bad, but we're emacsers each like things in a very particular way. I'd say give it a try and see how it works out in practice. We can potentially look into customisation if its really getting in the way.

The first iteration of the consolidated prompt is now submitted https://github.com/xenodium/agent-shell/commit/6a33d6a2eae163a5dbd5ec47245f129fe2cd4139

Screenshot 2026-02-24 at 10 38 34

I'd like to rework some internals to simplify implementing this sort of interaction, but it's going to need bigger changes. Will keep interating on it over time.

catern:
Yes, that seems great!  I think that obsoletes this PR, thanks for the better change.

catern:
Just one further thought that I'll put here: perhaps the "Open existing shell" option should actually go in the agent-shell-select-config prompt (when there is one)?  That would avoid unnecessarily starting up an ACP server.

xenodium:
Normally, if you invoke M-x agent-shell (w/ agent-shell-select-config set) and there's no shell found in project, you are now presented with the new prompt. If there's a shell already in project, it's focused. However, C-u M-x agent-shell overrides this behaviour and goes straight into this new prompt. This is particularly handy if you'd like to send context from one project to another (both already shells running).

ps. I have a regression carrying context for this scenario since the latest changes, but will fix that.

catern:
>if you invoke M-x agent-shell (w/ agent-shell-select-config set) 

Do you mean with `agent-shell-preferred-agent-config` set?  I was talking about the scenario when `agent-shell-preferred-agent-config` isn't set, and we therefore call `agent-shell-select-config` to prompt for which ACP backend to use.

xenodium:
Oh. I see. We can do that. I assumed you meant moving "Open existing shell" from the combined prompt introduced in this discussion to the provider prompt, but I think we need it in both (with some heuristic in the second not to show it twice). We otherwise never get offered to "Open existing shell" when preferred config is set and we're in a CWD already running a shell (the case for sending context from one project to another).

PR #318, opened 2026-02-20
Include tool call parameters in transcript

systemfreund:
Include tool parameters (filePath, offset, limit, pattern, etc.) in transcript entries.

Transcripts show exactly which files were read, what patterns were searched, and other tool-specific parameters that were previously omitted.

Parameters like command, description, and plan are excluded as they are already displayed separately in transcript entries.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1126,7 +1126,7 @@ COMMAND, when present, may be a shell command string or an argv vector."
                                               (map-nested-elt update '(rawInput command))))
                               (cons :description (map-nested-elt update '(rawInput description)))
                               (cons :content (map-elt update 'content))
-                              (cons :rawInput (map-elt update 'rawInput)))
+                              (cons :raw-input (map-elt update 'rawInput)))
                         (when-let ((diff (agent-shell--make-diff-info :tool-call update)))
                           (list (cons :diff diff)))))
                (agent-shell--emit-event
@@ -1258,7 +1258,7 @@ COMMAND, when present, may be a shell command string or an argv vector."
                                       ((not (map-nested-elt state `(:tool-calls ,.toolCallId :command)))))
                             (list (cons :command command)))
                           (when-let ((raw-input (map-elt update 'rawInput)))
-                            (list (cons :rawInput raw-input)))
+                            (list (cons :raw-input raw-input)))
                           (when-let ((diff (agent-shell--make-diff-info :tool-call update)))
                             (list (cons :diff diff)))))
                  (agent-shell--emit-event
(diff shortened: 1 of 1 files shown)


agent-shell.el
```diff
@@ -5524,6 +5530,15 @@ When nil, transcript saving is disabled."
                  (function :tag "Custom function"))
   :group 'agent-shell)
 
+(defcustom agent-shell-transcript-include-parameters nil
```
xenodium: Let's not make configurable please (more to maintain at it becomes a public API). Let's just include the parameters always, which a good behaviour anyway.

agent-shell.el
```diff
@@ -1125,7 +1125,8 @@ COMMAND, when present, may be a shell command string or an argv vector."
                               (cons :command (agent-shell--tool-call-command-to-string
                                               (map-nested-elt update '(rawInput command))))
                               (cons :description (map-nested-elt update '(rawInput description)))
-                              (cons :content (map-elt update 'content)))
+                              (cons :content (map-elt update 'content))
+                              (cons :rawInput (map-elt update 'rawInput)))
```
xenodium: nit: `raw-input` please.  I'm trying to keep all internal keywords non-camel so it's easier to spot what is part of the protocol vs internal.

agent-shell.el
```diff
@@ -5585,10 +5600,42 @@ Returns the file path, or nil if disabled."
       (error
        (message "Error writing to transcript: %S" err)))))
 
-(cl-defun agent-shell--make-transcript-tool-call-entry (&key status title kind description command output)
+(defun agent-shell--extract-tool-parameters (raw-input)
+  "Extract and format tool parameters from RAW-INPUT.
```
xenodium, line 5594: Could you please include an example in docstring with input and outoput?

agent-shell.el
```diff
@@ -5585,10 +5600,42 @@ Returns the file path, or nil if disabled."
...
+Returns a formatted string of key parameters, or nil if no relevant
+parameters found.  Excludes `command' and `description' as these are
+already shown separately in transcript entries."
+  (when raw-input
+    (let ((params '())
```
xenodium: Can't we combine this let and when below using when-let?

xenodium (commented):
Thanks for the PR. Left some minor comments. Could you take a look please?

agent-shell.el
systemfreund: I'm not that used to writing elisp code. Do you mean
```
(when-let ((raw-input raw-input)
           (params '())
            ...)
```

As `'() == nil` wouldn't this lead to the body not being evaluated?

xenodium: I think we can flatten further with somthing like (please check it):

```emacs-lisp
(when-let* ((raw-input)
            (excluded-keys '(command description plan))
            (params (seq-remove
                     (lambda (pair)
                       (let ((key (car pair))
                             (value (cdr pair)))
                         (or (memq key excluded-keys)
                             (null value)
                             (and (stringp value) (string-empty-p value)))))
                     raw-input)))
  (mapconcat (lambda (pair)
               (format "%s: %s"
                       (symbol-name (car pair))
                       (cond
                        ((stringp (cdr pair)) (cdr pair))
                        ((numberp (cdr pair)) (number-to-string (cdr pair)))
                        ((eq (cdr pair) t) "true")
                        (t (prin1-to-string (cdr pair))))))
             params
             "\n"))
```

xenodium:

PR #319, opened 2026-02-20
Fix "Wrong type argument: font, unspecified" in terminal Emacs

eddof13:
## Problem

Running `agent-shell-anthropic-start-claude-code` (or any agent start command) in terminal Emacs (`emacs -nw`) raises:

```
Wrong type argument: font, unspecified
```

The error originates in `agent-shell--make-header-model`:

```elisp
(:font-size . ,(font-get (face-attribute 'default :font) :size))
```

In a terminal frame, `(face-attribute 'default :font)` returns the symbol `unspecified` — terminal frames have no GUI font object. Passing `unspecified` to `font-get`, which requires a font as its first argument, raises the type error and aborts the whole startup path.

## Fix

Guard the call with `display-graphic-p` and `fontp` so `:font-size` is `nil` when no GUI font is available:

```elisp
(:font-size . ,(let ((font (and (display-graphic-p) (face-attribute 'default :font))))
                 (when (fontp font) (font-get font :size))))
```

The downstream SVG rendering code already reads `:font-size` via `(map-elt header-model :font-size)` and uses it in arithmetic — a `nil` value here is the appropriate sentinel for "no font size available" in a terminal context.


eddof13:
hey this fixed an issue I had earlier, I switched terminal and updated the package so one of the two might have exposed it

xenodium:
Thank you for the fix

xenodium:
Heads up, minor tweak https://github.com/xenodium/agent-shell/commit/abe76a657dee27ba12c3018d44ca428fa52567ef

PR #323, opened 2026-02-21
Allow shell's to start outside of project root's via defcustom

timvisher-dd:
- rename `agent-shell-cwd` to `agent-shell-project-root` to clarify that this root anchors shell grouping, path shortening, and artifact placement
- introduce `agent-shell-project-root-function` as a defcustom override, so you can return `default-directory` (or any directory) to start shells outside the project root and have grouping/path resolution follow that root
- fix a few tests that were failing on trunk and align plan-formatting expectations with short status labels

### DIFF
--- a/README.org
+++ b/README.org
@@ -223,6 +223,21 @@ You can load environment variables from .env files using the `:load-env` paramet
 
 The .env files should contain variables in the format `KEY=value`, with one variable per line. Comments (lines starting with `#`) and empty lines are ignored.
 
+*** Project root
+
+By default, agent-shell uses the current project root (Projectile or project.el)
+and falls back to `default-directory`. See `agent-shell-project-root` for
+details on how this root is used. To override that behavior, customize
+`agent-shell-project-root-function` with a function that returns a directory.
+
+#+begin_src emacs-lisp
+;; Use the current buffer directory.
+(setq agent-shell-project-root-function (lambda () default-directory))
+
+;; Use a fixed directory.
+(setq agent-shell-project-root-function (lambda () (expand-file-name "~/work/")))
+#+end_src
+
 *** Anthropic Claude
 
 For login-based authentication (default):
--- a/agent-shell-project.el
+++ b/agent-shell-project.el
@@ -31,6 +31,23 @@
 
 (defvar projectile-mode)
 
+(defcustom agent-shell-project-root-function nil
+  "Optional function override for `agent-shell-project-root'.
+
+See `agent-shell-project-root' for details on how the root is used.
+
+When non-nil, the value must be a function called with no arguments
+that returns a directory satisfying `file-accessible-directory-p'.
+
+Examples:
+- Use the current buffer directory via a function:
+  (setq agent-shell-project-root-function (lambda () default-directory))
+- Use a custom function:
+  (setq agent-shell-project-root-function (lambda () \"~/work/\"))"
+  :type '(choice (const :tag "Default" nil)
+                 (function :tag "Function (no args)"))
+  :group 'agent-shell)
+
 (declare-function projectile-project-p "projectile")
 (declare-function projectile-project-root "projectile")
 (declare-function projectile-project-name "projectile")
@@ -54,20 +71,35 @@
               (project-files proj))))
    (t nil)))
 
-(defun agent-shell-cwd ()
-  "Return the CWD for this shell.
+(defun agent-shell-project-root ()
+  "Return the project root for this shell.
 
-If in a project, use project root."
-  (expand-file-name
-   (or (when (and (boundp 'projectile-mode)
-                  projectile-mode
-                  (fboundp 'projectile-project-root))
-         (projectile-project-root))
-       (when (fboundp 'project-root)
-         (when-let ((proj (project-current)))
-           (project-root proj)))
-       default-directory
-       (error "No CWD available"))))
+This root anchors project-scoped behavior:
+
+- Shell grouping: used to find or prompt for a shell when the current
+  buffer's project changes.
+- Path shortening and file mentions: file paths are shortened relative
+  to this root, and file mentions expand against it.
+- Project artifacts: screenshots and transcripts are stored under the
+  `.agent-shell/' directory in this root.
+
+If `agent-shell-project-root-function' is nil, use the project root when
+available, falling back to `default-directory'.  When
+`agent-shell-project-root-function' is non-nil, use that override."
+  (let ((project-root (if agent-shell-project-root-function
+                          (funcall agent-shell-project-root-function)
+               (or (when (and (boundp 'projectile-mode)
+                              projectile-mode
+                              (fboundp 'projectile-project-root))
+                     (projectile-project-root))
+                   (when (fboundp 'project-root)
+                     (when-let ((proj (project-current)))
+                       (project-root proj)))
+                   default-directory
+                   (error "No project root available")))))
+    (unless (file-accessible-directory-p project-root)
+      (error "Project root is not an accessible directory: %S" project-root))
+    (expand-file-name project-root)))
 
 (defun agent-shell--project-name ()
   "Return the project name for this shell.
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -532,8 +532,8 @@ Example configuration with multiple servers:
 
 Lambdas can be used anywhere in the configuration hierarchy for dynamic
 evaluation at session startup time.  This is useful for values that
-depend on runtime context like the current working directory
-\(`agent-shell-cwd').  Note: only lambdas are evaluated, not named
+depend on runtime context like the current project root
+\(`agent-shell-project-root').  Note: only lambdas are evaluated, not named
 functions, to avoid accidentally calling external symbols.
 
 For example, using the `claude-code-ide' package (see its documentation
@@ -546,7 +546,7 @@ the session and returns the appropriate endpoint:
            (headers . ())
            (url . (lambda ()
                     (require \='claude-code-ide-mcp-server)
-                    (let* ((project-dir (agent-shell-cwd))
+                    (let* ((project-dir (agent-shell-project-root))
                            (session-id (format \"agent-shell-%s-%s\"
                                          (file-name-nondirectory
                                            (directory-file-name project-dir))
@@ -886,10 +886,10 @@ When FORCE is non-nil, skip confirmation prompt."
       :shell-buffer (map-elt shell :buffer)))))
 
 (defun agent-shell--filter-buffer-substring (start end &optional delete)
-  "Return the buffer substring between BEG and END, after filtering.
+  "Return the buffer substring between START and END, after filtering.
 Strip the text properties `line-prefix' and `wrap-prefix' from the
-copied substring.  If DELETE is non-nil, delete the text between BEG and
-END from the buffer."
+copied substring.  If DELETE is non-nil, delete the text between START
+and END from the buffer."
   (let ((text (if delete
                   (prog1 (buffer-substring start end)
                     (delete-region start end))
@@ -1620,7 +1620,7 @@ For example:
 
 - /workspace/README.md => /home/xenodium/projects/kitchen-sink/README.md
 - /home/xenodium/projects/kitchen-sink/README.md => /workspace/README.md"
-  (let* ((cwd (agent-shell-cwd))
+  (let* ((cwd (agent-shell-project-root))
          (devcontainer-path (agent-shell--get-devcontainer-workspace-path cwd)))
     (if (string-prefix-p cwd path)
         (string-replace cwd devcontainer-path path)
@@ -1984,7 +1984,7 @@ DESTINATION-DIR is required and must be provided."
 Returns the full path to the saved image file on success.
 When NO-ERROR is non-nil, return nil instead of signaling errors.
 
-Needs external utilities. See `agent-shell-clipboard-image-handlers'
+Needs external utilities.  See `agent-shell-clipboard-image-handlers'
 for details."
   (unless destination-dir
     (error "Destination-dir is required"))
@@ -2052,7 +2052,6 @@ Returns a propertized string or nil."
 (defcustom agent-shell-status-kind-label-function
   #'agent-shell--default-status-kind-label
   "Function to render status and kind labels.
-
 Called with two arguments: STATUS (string or nil) and KIND (string or nil).
 Should return a propertized string or nil.
 
@@ -2077,7 +2076,7 @@ With INCLUDE-PROJECT
 
 \"/path/to/project/file.txt\" -> \"project/file.txt\""
   (when text
-    (let ((cwd (string-remove-suffix "/" (agent-shell-cwd))))
+    (let ((cwd (string-remove-suffix "/" (agent-shell-project-root))))
       (replace-regexp-in-string (concat (regexp-quote
                                          (if include-project
                                              (string-remove-suffix
@@ -2223,7 +2222,7 @@ variable (see makunbound)"))
                               :prompt (map-elt config :shell-prompt)
                               :prompt-regexp (map-elt config :shell-prompt-regexp)))
          (agent-shell--shell-maker-config shell-maker-config)
-         (default-directory (agent-shell-cwd))
+         (default-directory (agent-shell-project-root))
          (shell-buffer
           (shell-maker-start agent-shell--shell-maker-config
                              t  ;; Always use no-focus, handle display below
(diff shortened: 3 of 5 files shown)


xenodium:
Thanks for the proposal.

Can't we simplify (and also make more versatile) by introducing defcustom agent-shell-cwd-function (nil by default)? When non-nil, agent-shell-cwd returns funcall of agent-shell-cwd-function. Does this work for your use case?

timvisher-dd:
> Thanks for the proposal.
> 
> Can't we simplify (and also make more versatile) by introducing defcustom agent-shell-cwd-function (nil by default)? When non-nil, agent-shell-cwd returns funcall of agent-shell-cwd-function. Does this work for your use case?

It would work but it would be less versatile not more, I think. In this case I can just specify

```
(agent-shell-cwd-resolver 'default-directory)
```

in my customizations and it DTRT. If it could only take a function then I'd need to wrap that in a function somehow which could be as simple as a little lambda but still why not just allow a variable to be specified.

To be clear this works as well:

```
(defun my-resolver …)

(agent-shell-cwd-resolver 'my-resolver)
```

as does

```
(agent-shell-cwd-resolver (lambda () default-directory))
```

so it's quite flexible.

xenodium:
> It would work but it would be less versatile not more

By versatile, I mean its much more configurable/bendable by users beyond the options that we are providing in the PR proposal. It also has lower maintenance burden (new cases do not have to be extended in agent-shell). Users can introduce whatever logic they prefer, without requesting additional changes in package.

timvisher-dd:
Before responding I'm dropping this back to draft because I need to check something about how `default-directory` behaves. I'm concerned that `agent-shell-cwd` is dynamic in `agent-shell` which could mean that active shells are sticky to it rather than the emacs instance which would not be what I meant to do. xD

Now on with the response:

> > It would work but it would be less versatile not more
> 
> By versatile, I mean its much more configurable/bendable by users beyond the options that we are providing in the PR proposal. It also has lower maintenance burden (new cases do not have to be extended in agent-shell). Users can introduce whatever logic they prefer, without requesting additional changes in package.

I think I'm misunderstanding what you're proposing.

> When non-nil, agent-shell-cwd returns funcall of agent-shell-cwd-function

So you're suggesting we make the defcustom a function only, right?

> By versatile, I mean its much more configurable/bendable by users beyond the options that we are providing in the PR proposal

This PR supports `funcall` so I'm not sure how an option that _only_ supports `funcall` is 'much more configurable/bendable… [than] the PR proposal'.

> It also has lower maintenance burden (new cases do not have to be extended in agent-shell)

Do you have any particular cases in mind? The PR allows a variable or a function to be passed. It doesn't allow literals to be passed but a user can support that by defining a variable or a function that resolves to that literal. There isn't a built-in function I'm aware of that returns `default-directory` because the expectation is that if you need that you'll just reference the variable.

OTOH if we _only_ support functions and someone later on comes along and wishes to directly supply `default-directory` (that's what I want to do) then they can't and are forced to wrap it in a function.

This support for variables or functions adds 2 lines to the function which I think is a small price to pay maintenance wise to support all of elisp's variable types.

> Users can introduce whatever logic they prefer, without requesting additional changes in package.

I think any user who wishes to introduce specialized logic would use a function, right? They'd either create a named function and set resolver defcustom to it or they'd use a lambda. So again I feel like I must be missing something. :)

xenodium:
> This PR supports funcall so I'm not sure how an option that only supports funcall is 'much more configurable/bendable… [than] the PR proposal'.

The function-only enables you to do everything you want to achieve, even if it means wrapping in a lambda which is not hard to do. The benefit for the maintainer (me) is that I can offer you what you want with a smaller API surface. The public contract is simpler and I also don't have to maintain different cases/logic internally.

timvisher-dd:
> > This PR supports funcall so I'm not sure how an option that only supports funcall is 'much more configurable/bendable… [than] the PR proposal'.
> 
> The function-only enables you to do everything you want to achieve, even if it means wrapping in a lambda which is not hard to do. The benefit for the maintainer (me) is that I can offer you what you want with a smaller API surface. The public contract is simpler and I also don't have to maintain different cases/logic internally.

Understood. I'll respect that if I choose to take this further. :)

timvisher-dd:
@xenodium OK this is now more of an intensive change than I originally meant to do only because the `agent-shell-cwd` name really threw me (and the agents) off. I think `agent-shell-project-root` is much more descriptive of how it's used and it's a pretty safe internal change even though the function is technically 'publicly' named. WDYT?

xenodium:
> I think `agent-shell-project-root` is much more descriptive of how it's used

Ah yes, I can see why that may be confusing but it's named cwd because that's the agent's working directory for the session, as per protocol https://agentclientprotocol.com/protocol/schema#newsessionrequest

xenodium:
The tests should be fixed in main now. Thanks for looking into them.

xenodium:
> introduce agent-shell-project-root-function as a defcustom override, so you can return default-directory (or any directory) to start shells outside the project root and have grouping/path resolution follow that root

Could you share a little more about your flow please? What's currently preventing "start shells outside the project root"? Shells can be started anywhere (even on non-project directories).

timvisher-dd:
> > I think `agent-shell-project-root` is much more descriptive of how it's used
> 
> Ah yes, I can see why that may be confusing but it's named cwd because that's the agent's working directory for the session, as per protocol https://agentclientprotocol.com/protocol/schema#newsessionrequest

That makes sense but I think it's better to name the internal function according to how it's used and then pass it to the protocol, perhaps with a comment about why it makes sense to pass as CWD, rather than to name it `cwd` but use it for other functions like shell discovery.

timvisher-dd:
> > introduce agent-shell-project-root-function as a defcustom override, so you can return default-directory (or any directory) to start shells outside the project root and have grouping/path resolution follow that root
> 
> Could you share a little more about your flow please? What's currently preventing "start shells outside the project root"? Shells can be started anywhere (even on non-project directories).

So I flip the way most emacs users operate on its head.

Most try to start as few emacs instances as possible and then rely on 'project' detection from various contexts to only surface what's relevant to them at the time (i.e. switching buffers, compiling, fuzzy finding files, whatever).

I start an emacs instance per work stream and expect that instance to only exist as long as a workstream exists. I may visit many different projects, non-project locations, start many subprocesses, whatever, but they're all related to that workstream.

So my expectation for agent-shell is that once I've started a shell it should be considered the 'active shell' for that entire emacs instance no matter what my current context is. That's the _wrong_ default behavior since it's a decidedly anachronistic usage pattern for emacs but [it works very well for me](https://xkcd.com/1782/). In order to support it I plan to customize the new option and implement whatever I need to to make it work.

Does that all make sense?

xenodium:
If I'm understanding correctly, you run a single agent-shell instance per Emacs session, correct? And the challenge that you are facing is that every time you invoke M-x agent-shell, you expect it to take you back to the existing instance (no matter where you are) but instead it offers to create a new instance when outside the cwd of an existing shell?

edit: I'm assuming M-x agent-shell is what's getting in the way, or is it a different interactive command?

timvisher-dd:
> If I'm understanding correctly, you run a single agent-shell instance per Emacs session, correct?

Partially. I often actually run many agent-shell's per Emacs session as well. But the core mismatch is the same regardless. I don't expect to be tied to a given project root or default-directory. I expect to only have to tell agent-shell where it should operate from once per emacs Session.

> And the challenge that you are facing is that every time you invoke M-x agent-shell, you expect it to take you back to the existing instance (no matter where you are) but instead it offers to create a new instance when outside the cwd of an existing shell?

Yes that's the core where I feel the problem. But I'll also probably be extending my cwd handling function to handle other entrypoints like the various`-start` functions so that once I have the project-root set they all default to using it.

timvisher-dd:
> The tests should be fixed in main now. Thanks for looking into them.

I'm still seeing errors on trunk.

My test script is the following so I'm not sure if I'm just doing something wrong or not.

<details>

```bash
#!/usr/bin/env bash
set -euo pipefail

repo_root="$(git rev-parse --show-toplevel)"
acp_root="/Users/tim.visher/git.experiment.worktrees/xenodium/acp.el/main"
shell_maker_root="/Users/tim.visher/git.experiment.worktrees/xenodium/shell-maker/main"
tests_dir="${repo_root}/tests"

shopt -s nullglob
compile_files=("${repo_root}"/*.el)
test_files=("${tests_dir}"/*-tests.el)
shopt -u nullglob

if (( ${#compile_files[@]} < 1 )); then
  echo "No compile targets found in ${repo_root}" >&2
  exit 1
fi

if (( ${#test_files[@]} < 1 )); then
  echo "No test files found in ${tests_dir}" >&2
  exit 1
fi

test_args=()
for file in "${test_files[@]}"; do
  test_args+=( -l "$file" )
done

rm -f "${repo_root}"/*.elc

emacs -Q --batch \
  -L "${repo_root}" \
  -L "${acp_root}" \
  -L "${shell_maker_root}" \
  -f batch-byte-compile \
  "${compile_files[@]}"

emacs -Q --batch \
  -L "${repo_root}" \
  -L "${acp_root}" \
  -L "${shell_maker_root}" \
  "${test_args[@]}" \
  -f ert-run-tests-batch-and-exit
```

</details>

xenodium:
>  In order to support it I plan to customize the new option and implement whatever I need to to make it work.

> Yes that's the core where I feel the problem. But I'll also probably be extending my cwd handling function to handle other entrypoints like the various-start functions so that once I have the project-root set they all default to using it.

> That's the wrong default behavior since it's a decidedly anachronistic usage pattern for emacs but [it works very well for me](https://xkcd.com/1782/).

I'd prefer if we get a better understanding of the desired flow and validate the added customisation point does indeed solve your use-case.

Having said that and this being an unusual use-case, have you considered advising functions to get the behaviour you'd like? Sure they are not ideal, but sometimes that's just what we need to do to get the exact behaviour we want.

Also for creating new shells with a different cwd strategy, can't this be achieved with a custom `my-agent-shell-start-shell` which let binds default-directory during shell creation?

It may be worth playing with these options until you get the behaviour you'd like and perhaps then we can look into customisation points. Having said that, sometimes our own custom code (let bind, advise, etc) in our init.el's is ok too.

timvisher-dd:
> > In order to support it I plan to customize the new option and implement whatever I need to to make it work.
> 
> > Yes that's the core where I feel the problem. But I'll also probably be extending my cwd handling function to handle other entrypoints like the various-start functions so that once I have the project-root set they all default to using it.
> 
> > That's the wrong default behavior since it's a decidedly anachronistic usage pattern for emacs but [it works very well for me](https://xkcd.com/1782/).
> 
> I'd prefer if we get a better understanding of the desired flow and validate the added customisation point does indeed solve your use-case.
> 
> Having said that and this being an unusual use-case, have you considered advising functions to get the behaviour you'd like? Sure they are not ideal, but sometimes that's just what we need to do to get the exact behaviour we want.
> 
> Also for creating new shells with a different cwd strategy, can't this be achieved with a custom `my-agent-shell-start-shell` which let binds default-directory during shell creation?
> 
> It may be worth playing with these options until you get the behaviour you'd like and perhaps then we can look into customisation points. Having said that, sometimes our own custom code (let bind, advise, etc) in our init.el's is ok too.

Of course. There's no limit to what we can hack together in Emacs. But this does seem like a generically valuable addition to agent-shell so I'd like to offer it.

If you disagree that's fine and I'll hack around it myself. ¯\\_(ツ)\_/¯

xenodium:
Doesn't something like this do what you're after?

```emacs-lisp
(defun my/agent-shell-new-shell ()
  "Start a new agent shell in ~/."
  (interactive)
  (let ((default-directory "~/"))
    (agent-shell-new-shell)))
```

timvisher-dd:
> Doesn't something like this do what you're after?
> 
> ```emacs-lisp
> (defun my/agent-shell-new-shell ()
>   "Start a new agent shell in ~/."
>   (interactive)
>   (let ((default-directory "~/"))
>     (agent-shell-new-shell)))
> ```

No. My implementation will need to be substantially more complex than that.

That said I can see how another user might want to do something like that. Hence it would be nice if they could with a simple defcustom.

xenodium:
> But this does seem like a generically valuable addition to agent-shell so I'd like to offer it.

Technically, we haven't validated that this addition gives you the flow that you're after. Until then, I'd shy away from adding customisation points we haven't found concrete uses for.

> That said I can see how another user might want to do something like that. Hence it would be nice if they could with a simple defcustom.

They might, or they might not. I'd prefer adding customisations when we're sure they are going to be used. Any code that we add is code I have to maintain, specially public APIs and customisations.

timvisher-dd:
> > But this does seem like a generically valuable addition to agent-shell so I'd like to offer it.
> 
> Technically, we haven't validated that this addition gives you the flow that you're after. Until then, I'd shy away from adding customisation points we haven't found concrete uses for.
> 
> > That said I can see how another user might want to do something like that. Hence it would be nice if they could with a simple defcustom.
> 
> They might, or they might not. I'd prefer adding customisations when we're sure they are going to be used. Any code that we add is code I have to maintain, specially public APIs and customisations.

How can I show you that I'm able to use the defcustom to achieve my goal? We both agreed earlier in the thread that having a user-provided function that returns an accessible directory would provide users all the flexibility they could ever need so I'm unsure now why you'd be unsure the customization could work.

I'll drop this back to draft for now and proceed with it in my `dev` trunk and come back to you when I know it works for my use case. How's that sound?

xenodium:
> How can I show you that I'm able to use the defcustom to achieve my goal? We both agreed earlier in the thread that having a user-provided function that returns an accessible directory would provide users all the flexibility they could ever need so I'm unsure now why you'd be unsure the customization could work.

I took a step back to figure out why we actually needed this and it's not really clear.

> > Doesn't something like this do what you're after?
> > ```emacs-lisp
> > (defun my/agent-shell-new-shell ()
> >   "Start a new agent shell in ~/."
> >   (interactive)
> >   (let ((default-directory "~/"))
> >     (agent-shell-new-shell)))
> > ```
> 
> No. My implementation will need to be substantially more complex than that.

Sure that was just an illustration. You can make your version as complex as you'd like it to be.

timvisher-dd:
> > How can I show you that I'm able to use the defcustom to achieve my goal? We both agreed earlier in the thread that having a user-provided function that returns an accessible directory would provide users all the flexibility they could ever need so I'm unsure now why you'd be unsure the customization could work.
> 
> I took a step back to figure out why we actually needed this and it's not really clear.

I'm not sure what's unclear about 'the default way that agent-shell determines the cwd doesn't work for me, so I'd like a customization point to change it'? I recognize of course that having infinite customization points is not possible/advisable so I wouldn't argue that every little bit of behavior needs to be behind a defcustom but this core 'how agent-shell determines it's context' seems obviously generally useful to my eyes and core to the functioning of agent-shell as a package.

xenodium:
> I'm not sure what's unclear about 'the default way that agent-shell determines the cwd doesn't work for me, so I'd like a customization point to change it'?

https://github.com/xenodium/agent-shell/commit/89fe2eba671a08bc5fa56bbb4a5bdc1ffb91d46d

In the future, please save us both some effort https://github.com/xenodium/agent-shell/blob/main/CONTRIBUTING.org

> Before implementing new features, please file a feature request first to discuss the proposal. This helps ensure alignment with the project’s direction and prevents unnecessary work.

timvisher-dd:
> > I'm not sure what's unclear about 'the default way that agent-shell determines the cwd doesn't work for me, so I'd like a customization point to change it'?
> 
> [89fe2eb](https://github.com/xenodium/agent-shell/commit/89fe2eba671a08bc5fa56bbb4a5bdc1ffb91d46d)

🚀 

I'm surprised because that commit doesn't seem to include any documentation or test additions which this changeset did. I could've kept the `agent-shell-cwd` name and we could've had better test coverage and documentation but I get it.
 
> In the future, please save us both some effort https://github.com/xenodium/agent-shell/blob/main/CONTRIBUTING.org
> 
> > Before implementing new features, please file a feature request first to discuss the proposal. This helps ensure alignment with the project’s direction and prevents unnecessary work.

Will certainly try to. I didn't think the original small addition warranted an pre-discussion via an issue since it didn't seem like it would be controversial.

timvisher-dd:
Actually ironically it could've been a direct variable all along. Not sure if this will meet all my needs but at least I have the extension point now.

```elisp
(lambda () command-line-default-directory)
```

-----

For the record at least for now this is close to what I was looking for:

```elisp
(defvar timvisher--agent-shell-cwd nil
  "Sticky CWD for agent-shell, set on first use per Emacs session.
Once set, all agent-shell instances share this CWD.")

(defun timvisher--agent-shell-cwd ()
  "Return a sticky CWD for agent-shell.
On first call, captures `command-line-default-directory' and caches it.
All subsequent calls return the cached value."
  (or timvisher--agent-shell-cwd
      (setq timvisher--agent-shell-cwd command-line-default-directory)))
```

so my earlier 'substantially more complicated' statement turned out to be false. :)

PR #328, opened 2026-02-23
Fix agent-shell-diff-mode

wmedrano:

## Checklist

### DIFF
--- a/agent-shell-diff.el
+++ b/agent-shell-diff.el
@@ -109,6 +109,7 @@ Arguments:
                   (diff-mode-read-only nil))
               (erase-buffer)
               (agent-shell-diff--insert-diff old new file diff-buffer)
+              (agent-shell-diff-mode) ;; Must happen after inserting diff
               ;; Add overlays to hide scary text.
               (save-excursion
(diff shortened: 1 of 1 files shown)


wmedrano:
- Sets `agent-shell-diff-mode` after the diff is generated or else the mode gets reset.
- Sets more bindings in the `agent-shell-diff-mode` keymap and updates headerline to display the available keybindings.
- Interrupt key is only overwritten if it is not already bound.

Demo:

image

xenodium (approved):
Thanks for the PR.

Had been meaning to make similar changes. agent-shell-diff.el initially started as quick-diff.el, which was meant to be more generic. Keeping generic hasn't made much sense for some time since we'd like to optimize its experience for agent-shell itself.

xenodium:
Heads-up on two follow-ups https://github.com/xenodium/agent-shell/commit/46fb6c1228910560e44b9b85dc01b47de99aa1d4 and https://github.com/xenodium/agent-shell/commit/7b16d6a619ba6f9c078fe83d898b1556f5960a62

Issue #309, opened 2026-02-19
Question: Tools verbosity


JuanPabloMF:
<img width="1818" height="742" alt="Image" src="https://github.com/user-attachments/assets/66fae391-0b27-4f05-bee0-80871cefe43e" />

I really like that with ACP everything is more "readable" by having collapsable sections and very clear design for different sections, but sometimes I'm missing some information.

Many tool uses show the full command that was run (for example with find command), but there are some tools that are very concise, like in the image you can see that **grep** and **read file** don't show the rest of the content.

As I understand agent-shell receives from ACP the following: title, description, rawInput and content. Is there in one of these strings the info that is lacking from grep (the rest of the grep command) and read file (the file that is being read)?

Would it be worth to have an option to display this info in some way if it is available via ACP?

xenodium:
It depends if its possible. Agents kinda include different things in the ACP. Some include commands in titles, some don't. To be able to help, I'll need traffic details along with a screenshot like the one you just supplied, so we can correlate display with protocol data.

ps. Please try to use the [new issue template](https://github.com/xenodium/agent-shell/blob/main/.github/ISSUE_TEMPLATE/issue.md), which runs you through all the things I'd need from you to be able to help you.

JuanPabloMF:
---
name: Feature Request
about: Can we see the full command that was called (like for a find or a grep)
title: ''
labels: ''
assignees: ''
---

<img width="954" height="926" alt="Image" src="https://github.com/user-attachments/assets/eeabbb41-2bb2-4354-95c8-058258be916a" />

Here is a screenshot of one of the find tool calls. 

<img width="1882" height="1028" alt="Image" src="https://github.com/user-attachments/assets/d82f1deb-b787-4495-8d2f-32ff9a5272cf" />

[acp-traffic.log](https://github.com/user-attachments/files/25443718/acp-traffic.log)

xenodium:
In https://github.com/xenodium/agent-shell/commit/4cbf1e83c077d64c1256f166037ea47b85cb1f74 I'm now trying to upgrade the title across incoming notifications more generically. See if that works for you. You'll need to update to 0.40 or newer.

JuanPabloMF:
It seems that the following is happening:

```
There it is. In the `tool_call_update` handler (line 1221), when saving the tool call, agent-shell only saves:

1. `:status` (line 1227)
2. `:content` (line 1228)
[...]

PR #320, opened 2026-02-20
Trying out more reliable title upgrade (trying thing out for #309)

xenodium:
Looking to see if tool title upgrade works better for #309

PR #333, opened 2026-02-24
Trying out more reliable title upgrade (trying thing out for #309)

xenodium:
Looking to see if tool title upgrade works better for https://github.com/xenodium/agent-shell/issues/309


xenodium:
Patch confirmed in https://github.com/xenodium/agent-shell/pull/320#issuecomment-3952884133

PR #344, opened 2026-02-26
Add regression tests for shell buffer selection ordering

timvisher-dd:
I thought I'd hit a bug where `C-u C-c A` to create a new shell followed by navigating away and `C-c A` would route to the older shell instead of the new one. After investigating `(buffer-list)` ordering empirically with `emacs --batch -Q` and tracing the code paths, we proved that the current `(buffer-list)`-based ordering in `agent-shell-buffers` works correctly for this scenario: `select-window (display-buffer ...)` (the path used by `agent-shell--display-buffer`) does promote the buffer.

These tests document and lock down the actual ordering behavior as regression protection:

- `switch-to-buffer` and `select-window` promote a shell to the front
- `with-current-buffer` (used by `agent-shell--handle`) does not promote
- `bury-buffer` demotes a shell to the end
- Buffers created without display stay at the end
- `agent-shell-project-buffers` filters by CWD and preserves ordering

## Checklist

- [ ] ~I've filed a feature request/discussion for a new feature.~ N/A since this just tests documented behavior.

### DIFF
--- a/tests/agent-shell-buffer-ordering-tests.el
+++ b/tests/agent-shell-buffer-ordering-tests.el
@@ -1,0 +1,158 @@
+;;; agent-shell-buffer-ordering-tests.el --- Tests for shell buffer ordering -*- lexical-binding: t; -*-
+
+(require 'ert)
+(require 'agent-shell)
+
+;;; Code:
+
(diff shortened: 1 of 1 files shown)


timvisher-dd:
Ping on this. Would be nice to have these regression tests locked in. :)

Issue #345, opened 2026-02-27
Stale session/update notifications leak into next prompt turn


timvisher-dd:
## Summary

After a turn completes (`on-success` fires), late-arriving `session/update` notifications carrying `agent_message_chunk` or `agent_thought_chunk` are rendered as if they belong to the *next* turn. The user sees stale output from the prior turn displayed as the response to their new prompt.

## How to reproduce

1. Use agent-shell with Claude Code (via claude-agent-acp) as the backend.
2. Have the agent perform several tool calls in a single turn.
3. When the turn completes and the prompt reappears, quickly type and submit a new question.
4. The response to the new question may show text from the *previous* turn's output before eventually showing the real answer.

The symptom is most visible when the "Notices" drawer shows errors like:

```
No onPostToolUseHook found for tool use ID: toolu_XXXXX
```

These errors and the stale output tend to appear together.

## Root cause

There are two layers to this bug:

### 1. Server-side: claude-agent-acp sends notifications after the RPC response

In `claude-agent-acp` (the ACP server wrapping the Claude Agent SDK), the `prompt()` method in `acp-agent.ts` returns as soon as the SDK yields a `"result"` message (line 520: `return { stopReason: "end_turn", usage }`). It does **not** await pending `PostToolUse` hook callbacks.

These hooks are registered during streaming via `registerHookCallback()` (`tools.ts` line 731-746) when `tool_use` content blocks are encountered. The Claude Agent SDK fires them asynchronously after tool execution completes. The registered callback (`acp-agent.ts` line 1482-1508) calls `await client.sessionUpdate(notification)` to send a `tool_call_update` notification to the pipe.

When a hook fires after `prompt()` has already returned, the SDK's `Connection#processMessage` (`@agentclientprotocol/sdk`, `acp.js` line 762) has already sent the RPC response. The hook's `sessionUpdate()` write goes through `#sendMessage` -> `#writeQueue`, but the response was already enqueued first. The notification arrives on the pipe **after** the response.

The global `toolUseCallbacks` map (`tools.ts` line 720-728) is the mechanism:
- Registered: during streaming, on first tool_use encounter (line 1480-1481)
- Fired: by SDK's PostToolUse hook callback (`tools.ts` line 756-775)
- Cleaned up: `delete toolUseCallbacks[toolUseID]` after firing (line 767)
- Error case: "No onPostToolUseHook found" logged when the ID was never registered or already consumed (line 769)

[...]

PR #346, opened 2026-02-27
Drop stale session/update notifications that arrive after turn completion

timvisher-dd:
Fixes https://github.com/xenodium/agent-shell/issues/345

See the linked issue for the full problem description and server-side analysis.

## What changed

- New `:turn-completed` state field in `agent-shell--make-state`, initialized to `nil`.
- `on-success` callback in `agent-shell--send-command` sets `:turn-completed` to `t` when the turn's RPC response arrives.
- `agent-shell--send-command` resets `:turn-completed` to `nil` alongside `:last-entry-type` when sending a new prompt.
- `agent_message_chunk` and `agent_thought_chunk` handlers in `agent-shell--on-notification` are wrapped with `(unless (map-elt state :turn-completed) ...)` to drop notifications that arrive after the turn has completed.

Even if the server fixes its ordering, the client should be resilient to out-of-order notifications rather than blindly trusting that all notifications arrive before the response.

## Test plan

- New test `agent-shell--stale-notification-after-turn-completion` verifies that a stale `agent_message_chunk` arriving after `:turn-completed` is set is silently dropped (no fragment created, `:chunked-group-count` unchanged), and that a subsequent turn's chunks are processed normally.
- Existing `agent-shell--send-command-integration-test` and `agent-shell--send-command-error-fallback-test` updated to include `:turn-completed` in their test state to avoid `map-not-inplace` errors.
- Full test suite: 36/36 pass.

## Checklist

- [x] I've reviewed all code in PR myself and will vouch for its quality.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -586,6 +586,7 @@ OUTGOING-REQUEST-DECORATOR (passed through to `acp-make-client')."
                              (cons :mode-id nil)
                              (cons :modes nil)))
         (cons :last-entry-type nil)
+        (cons :turn-completed nil)
         (cons :chunked-group-count 0)
         (cons :request-count 0)
         (cons :tool-calls nil)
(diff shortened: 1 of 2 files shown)


ElleNajt:
Hm --

It seems like the stale turn alert solution breaks my [idiosyncratic multiplexing workflow]( https://github.com/ElleNajt/acp-multiplex), because messages sent in response to messages from a secondary client are marked stale.

Something like multiplexing is coming to ACP https://github.com/agentclientprotocol/agent-client-protocol/pull/533 , so there's reason to not just throw away messages that are considered stale in this way. 

In my fork I now have reverted this commit, I'd like it if I didn't have to patch this change out. Maybe we can relegate this stale message handling to some flag that I can disable?

xenodium:
Uh oh. [Are we here right now](https://xkcd.com/1172) ;)?

While offering a flag to re-enable a behaviour breaking offifical features doesn't feel quite right, I'm happy to offer a compromise by introducing agent-shell--active-requests-p in https://github.com/xenodium/agent-shell/commit/45f0107a87d2790411ba141f1cd221464c27135a which can be easily adviced.

timvisher-dd:
LOL this space is moving fast… xD

PR #350, opened 2026-02-27
Prepare Devcontainer package extraction

fritzgrabo:
As discussed, this prepares the extraction of a separate package for Devcontainer support.

Highlights:

- Rename `-container-command-runner` to `-command-prefix`. This is one of the two extension points that will remain in agent-shell (the other being `agent-shell-path-resolver-function` fwiw). I wanted to get rid of the "container" term in the variable name, so came up with a more generic "command prefix". There's an alias for the old name so we don't break existing setups (just yet).

- Extracts all Devcontainer-related code and tests from `agent-shell.el` into a separate file `agent-shell-devcontainer.el` (and `-test.el`) for now, and renames everything accordingly. Again, there's an alias for the old name of `agent-shell-devcontainer-resolve-path` so existing setups don't break.

- Updates to README.md as needed.

Please let me know if this is what you expected.

If this looks good, I'd pull `agent-shell-devcontainer*.el` and the related bits from the README into a separate repo and put the new package on MELPA next. Once the package is available, we can remove everything from here and mention the new package in the README.

Thanks!

### DIFF
--- a/README.org
+++ b/README.org
@@ -453,10 +453,10 @@ You can use different containers for different shell sessions, even of the same 
 Note that any =:environment-variables= you may have passed to =acp-make-client= will not apply to the agent process running inside the container. It's expected to inject environment variables by means of your devcontainer configuration / Dockerfile.
 
 Next, set an =agent-shell-path-resolver-function= that resolves container paths in the local working directory, and vice versa.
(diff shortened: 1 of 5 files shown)


xenodium (approved):

xenodium:
> Please let me know if this is what you expected.

Yes. This looks great. Thank you for doing this.

Are you ok with me merging?

fritzgrabo:
Yes, absolutely. Thanks!

PR #351, opened 2026-02-27
feat: Add Kiro CLI ACP support

zmjones:
Add agent-shell-kiro module with ACP integration for Kiro CLI. Includes configuration, client setup, and ASCII art welcome message. Kiro is the main agentic development tool developed by (and used internally by) Amazon. 

The kiro-cli send some custom metadata which is used to update the prompt with the % context used. I had some trouble getting the prompt to dynamically update using these notifications and didn't see any similar logic for other agents, so for now they are silently ignored by the notification handler.

Installation:
  curl -fsSL https://cli.kiro.dev/install | bash

Registered in default agent configs and preferred agent choices. Updated README with installation instructions and command reference.

## Checklist


agent-shell.el
```diff
@@ -1122,7 +1125,10 @@ COMMAND, when present, may be a shell command string or an argv vector."
 
 (cl-defun agent-shell--on-notification (&key state acp-notification)
   "Handle incoming ACP-NOTIFICATION using STATE."
-  (cond ((equal (map-elt acp-notification 'method) "session/update")
+  (cond ((string-prefix-p "_kiro.dev/" (map-elt acp-notification 'method))
```
xenodium: If I'm understanding correctly, this was added to prevent the fallback case of displaying unhandled notifications.

I've just added https://github.com/xenodium/agent-shell/commit/ed36fbec502b618ecd5fe2cb3c9daa353e804ccd which should hide unhandled display by default, but enables it when logging is on.

Does that work for your use-case?

zmjones: The behavior after adding the initial support was to print an unhandled notification message which printed this custom metadata which included information about context use. Kiro-CLI's default behavior is print a prompt as x%> which I tried to replicate, but basically I don't understand enough to do that now. It also just passes the % not the tokens used, so it didn't seem to fit with agent-shell-usage, and I didn't want to do any janky detecting of available tokens based on model name. So I thought just to filter it out but of course the general solution is better. 

Yes your change does work for me. I rebased and pushed again so that part should be removed.

xenodium:
Thanks for the follow up. Merged.

Issue #352, opened 2026-02-28
On X11, xclip blocks when pasting text


chemtov:
Hi,

when using `agent-shell-yank-dwim` and having text in the `CLIPBOARD` using `X11` session, `xclip` just hangs.  I need to `C-g` to have the text pasted.

I'm surprised to be alone here, but maybe everyone on Linux is using Wayland and there it might works better.

I have a patch for this gated on the usage of X11 as I didn't test it on Wayland.

Regards,

rpoisel:
> when using agent-shell-yank-dwim and having text in the CLIPBOARD using X11 session, xclip just hangs. I need to C-g to have the text pasted.

I observed similar behavior on my Debian GNU/Linux (unstable) box, also running X11.

The only difference: `C-g` cancels "the paste" completely.

Workaround to get text pasted: `[mouse-2]` (that's pressing the mouse-wheel in my case).

chemtov:
Yes, that's right `C-g` cancel the paste entirely in normal usage, only in `edebug` session was it able to just get paste the blocking call and paste.  Sorry for the description error.

Pressing `[mouse-2]` must call `yank` directly which works and it's what I was using before the patch.

xenodium:
Thank you both.

I've merged the change and made a small tweak in https://github.com/xenodium/agent-shell/commit/e5c8cd4a358b428d3810d197522a7483fc98585f. Could you please verify it?

I'm on macOS myself, but the flow uses C-y/yank remapped to agent-shell-yank-dwim, so you can paste images from the clipboard like you'd typically paste/yank. See screencast:

![Image](https://github.com/user-attachments/assets/4a992958-b815-44ac-9a2e-c8b263a318c9)

rpoisel:
> I've merged the change and made a small tweak in [e5c8cd4](https://github.com/xenodium/agent-shell/commit/e5c8cd4a358b428d3810d197522a7483fc98585f). Could you please verify it?

Can confirm. Works on my machine - both with text and PNG files.

Thanks again. Great that got fixed so quickly!

xenodium:
> Thanks again. Great that got fixed so quickly!

Glad to hear. If you're getting value out of agent-shell, please consider [sponsoring](https://github.com/sponsors/xenodium) this project (I am now exclusively an indie dev), so I can sustainably continue to maintain, improve, and provide support.

PR #353, opened 2026-02-28
Fix xclip handler blocking on X11 when clipboard has no image

chemtov:
On X11, calling xclip from a subprocess while Emacs owns the clipboard
deadlocks: Emacs can't service the X11 selection conversion request
while blocked in call-process.

Fix this by checking clipboard targets in-process via gui-get-selection
before invoking xclip, and erroring early if image/png is not available.
The check is gated on (window-system) 'x so Wayland/pgtk users keep the
original behavior.

Closes: #352

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -320,6 +320,11 @@ Assume screenshot file path will be appended to this list."
                            (error "Command pngpaste failed with exit code %d" exit-code))))))
    (list (cons :command "xclip")
          (cons :save (lambda (file-path)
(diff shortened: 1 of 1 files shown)


xenodium (approved):
Thanks for the fix

rpoisel:
Thanks, @chemtov, @xenodium, and @claude 😁

PR #362, opened 2026-03-03
Fix: Track and delete permission fragments using request ID to support multiple concurrent requests

tempdragon:
### Description

**What does this PR do?**
This PR fixes an issue where handling multiple tool permission requests in a single turn could result in mismanaged or stale UI fragments (orphaned permission dialogs) remaining in the shell buffer. 

**Why is this needed?**
Previously, permission dialog fragments were tracked, updated, and deleted using the `toolCallId`. However, the Agent Client Protocol (ACP) issues a `session/request_permission` with its own unique JSON-RPC request `id`. When an agent makes multiple simultaneous tool calls requiring permission, relying solely on `toolCallId` to manage the dialog's lifecycle causes inconsistencies when trying to clear the dialogs after a user responds.

**How was this solved?**
*   Modified `agent-shell--on-request` to save the `permission-request-id` into the tool call's local state when a `session/request_permission` is received.
*   Updated the block ID format for permission dialogs to use the JSON-RPC request ID (i.e., `(format "permission-%s" request-id)`) instead of the `tool-call-id`.
*   Updated the cleanup routines in `agent-shell--send-permission-response` and `agent-shell--on-notification` to correctly lookup and delete the fragment using the stored `req-id`.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1353,7 +1353,7 @@ COMMAND, when present, may be a shell command string or an argv vector."
                ;; likely selected one of: accepted/rejected/always.
                ;; Remove stale permission dialog.
                (when (and (map-nested-elt acp-notification '(params update status))
-                          (member (map-nested-elt acp-notification '(params update status)) '("completed" "failed" "cancelled" "rejected")))
+                          (not (equal (map-nested-elt acp-notification '(params update status)) "pending")))
                  ;; block-id must be the same as the one used as
                  ;; agent-shell--update-fragment param by "session/request_permission".
                  (when-let ((req-id (map-nested-elt state `(:tool-calls ,(map-nested-elt acp-notification '(params update toolCallId)) :permission-request-id))))


tempdragon:
A way to test: In agent-shell-github-start-copilot

```
Perform up to 6 concurrent local writes for these queries: ['OpenAI GPT-5 release date','Python 3.12 release notes','SpaceX Starship first orbital attempt March 2026','How to tie a Half Windsor knot','Top cybersecurity threats 2026 Gartner'] (or use a provided list).  
For each query, atomically create a UTF‑8 text file named test_query_<slug>.txt in a session directory (create it if missing) with line 1: "Query: <original query>" and line 2: a one-line snippet "Local test file simulating a search result for '<query>'"; perform the writes in parallel (max concurrency 6), on write failure retry up to 2 times (log each retry with attempt number and reason and verify file contents after write), and ask for permission before doing more than 2 retries.  YOU MUST NOT READ UNTIL THE WRITES ARE FINISHED.
Return a machine-readable JSON array of objects [{query, results:[{title,url,domain:'local',snippet}], retries:[{attempt,reason}], status:'written'|'failed'}] plus a 2–3 sentence human-readable summary noting any failures and suggested next steps; do not perform any web searches or use a `functions.bash`.
```

The 4ad3dc9 will have:

2026-03-03-202005_1920x1080_scrot

Blocked after you choose the first yes.

For the fixed version:

2026-03-03-205240_1920x1080_scrot

agent-shell.el
```diff
@@ -1353,10 +1353,11 @@ COMMAND, when present, may be a shell command string or an argv vector."
                ;; likely selected one of: accepted/rejected/always.
                ;; Remove stale permission dialog.
                (when (and (map-nested-elt acp-notification '(params update status))
-                          (not (equal (map-nested-elt acp-notification '(params update status)) "pending")))
+                          (member (map-nested-elt acp-notification '(params update status)) '("completed" "failed" "cancelled" "rejected")))
```
xenodium: Is this change needed to fix the issue the PR is looking to fix?

tempdragon: I think mostly the answer is yes. I updated this condition to match the ACP spec for [`ToolCallStatus`](https://agentclientprotocol.com/protocol/schema#toolcallstatus). The old logic `(not (equal status "pending"))` was a little bit fragile because `in_progress` is also a valid, non-terminal state in the protocol. If an agent sends an `in_progress` update, the old logic would prematurely remove the dialog/status. Checking explicitly against the four terminal states (`completed`, `failed`, `cancelled`, `rejected`) ensures we only clean up the fragment when the tool call is completely finished.

xenodium: Doesn't in_progress imply the permission dialog is no longer needed though? The tool call is already running. Why keep the dialog open and make it seem like the user needs to grant permission still?

tempdragon: Yes. You are right. I have pushed a new commit to revert it. Thanks for the observation.

xenodium:
Looking further into it, the new approach breaks permission fragment removal based on tool call updates (`tool_call_update` notifications). Since `toolCallId` in those notifications doesn't reliably map back to the stored `request-id`, the notification cleanup path can't find the right fragment to delete for all of the permission dialogs.

If we move to keying fragments by `request-id`, we'd also need to consider removing the update-based removal entirely rather than trying to look up the `request-id` from saved state. That also means we wouldn't need to store `:permission-request-id` in the tool call data at all since the button closures already capture `request-id` directly at creation time.

Having said all of this, the root cause here seems to be Copilot reusing `toolCallId` values (`"read-permission"`, `"write-permission"`) across distinct tool calls. The ACP spec requires toolCallId to be a unique identifier per tool call within a session (https://agentclientprotocol.com/protocol/schema#toolcallid). I'm thinking we should consider filing an isssue with Copilot if this behavior isn't inline with ACP spec.

cc @ewilderj

tempdragon:
Yes. Sadly, the fact that GitHub copilot-cli is a commercial software makes it unreliable to count on them to fix the bug. Maybe we should somehow find a workaround before it is changed.

xenodium:
We should give them a chance to fix it. The challenge with adding a workaround in agent-shell is that we need to downgrade our approach (we lose the ability to hide dialogs from agent updates).

Btw, the issue's been reported by a Microsoft employee at https://github.com/github/copilot-cli/issues/989. I've also [commented](https://github.com/github/copilot-cli/issues/989#issuecomment-3996220466).

Issue #369, opened 2026-03-05
Permission ui disappears and tool hangs if the tool status is updated before permission is granted


xar7:
Hey hi,
First thanks for this nice package and the work you put into it :)

I used it quite a lot the last weeks (mainly with opencode) and notice that I often face the following issue: the agent asks for a permission but I don't see the permission-ui which allows me to grant/deny, for me it looks like the agent is hanging, so I'm forced to interrupt and retry and try to avoid any permission prompt ...

I debugged this a little bit myself using your acp package to analyze the traffic and logs and found that the backend sometimes sends us a session update that changes the tool call execution from `pending` to `in_progress` EVEN IF the tool is not yet executing as it's waiting for the permission ... This status update makes the permission ui disappears as agent-shell thinks that the permission ui should be there only if the status is currently `pending`.

Maybe this is an issue with OpenCode but I'm pretty sure I got the same situation with Codex also.

agent-shell: 20260303.1631
acp: 20260302.1640
emacs: 30.2
opencode: 1.2.15

I reproduced the issue with an minimal emacs config loading only agent-shell and its dependencies:
<img width="1692" height="1264" alt="Image" src="https://github.com/user-attachments/assets/fe5dbb00-ac0e-44cf-bf06-589a035e7322" />
Here are the [agent-shell acp logs](https://github.com/user-attachments/files/25771863/agent-shell-acp-log-traffic-permission-ui-issue.log) and [acp traffic](https://github.com/user-attachments/files/25771897/acp-traffic-permission-ui-issue.log) from the issue as it happened in the screenshot(I asked the agent to execute `ls` to reproduce the issue).

I also already implemented a fix where I simply introduce a local `permission_state` variable and make agent-shell reacts on this instead of only on the `pending` status, will open a MR shortly after.

## Checklist

- [X] *I agree to communicate with the author myself* (not AI-generated).
- [X] I've read the README's [Filing issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#filing-issues) section.
- [X] I'm running the latest versions (fill in below).
  - agent-shell: 20260303.1631
  - acp.el: 20260302.1640
  - ACP package (e.g. claude-code-acp): opencode 1.2.15
  - Agent CLI (e.g. claude, gemini): opencode
- [ ] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
[...]

PR #370, opened 2026-03-05
Make permission-ui stays visible even if status_update comes

xar7:
This MR is my attempt of fixing the problem described in this issue : https://github.com/xenodium/agent-shell/issues/369

This introduces a `permission-state` that tracks if the user actually answered or not a permission ask of the agent. This solves the issue where the agent would update the tool call to `in_progress` even if the permission wasn't granted yet (and the agent is still waiting for it).

I'm currently running this and it fixes my issue with OpenCode/Codex.

I'm not that good in elisp so please don't hesitate to re-write it better or provide another solution. :)

## Checklist

- [X] I've filed a feature request/discussion for a new feature.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1369,11 +1369,11 @@ COMMAND, when present, may be a shell command string or an argv vector."
                          :output body-text)
                   :file-path agent-shell--transcript-file))
                ;; Hide permission after sending response.
-               ;; Status and permission are no longer pending. User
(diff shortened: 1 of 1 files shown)


xar7:
This now makes the permission ui now reacts on status changing to `failed` or `completed` thus ignoring out of order `in_progress`.

xenodium:
Thanks for the PR. Merged.

PR #371, opened 2026-03-05
Add minimum transient version (0.9.0) to Package-Requires

eddof13:
## Problem

Users loading `agent-shell` may encounter this error at startup:

```
Error (use-package): agent-shell/:catch: Symbol's function definition is void: transient--set-layout
```

## Root Cause

`transient--set-layout` was introduced in **transient 0.9.0** (released 2025-06-01) as an internal function used by the `transient-define-prefix` macro. When the package is compiled against transient 0.9.0+, calls to `transient--set-layout` are baked into the `.elc` bytecode via macro expansion.

If Emacs loads the built-in `transient` (shipped with Emacs 29.x, which is pre-0.9.0) instead of the MELPA version, the function is missing at runtime.

## Fix

Add `(transient "0.9.0")` to `Package-Requires` so that package managers (ELPA/MELPA) automatically enforce the minimum version and install the correct transient from MELPA when needed.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -5,7 +5,7 @@
 ;; Author: Alvaro Ramirez https://xenodium.com
 ;; URL: https://github.com/xenodium/agent-shell
 ;; Version: 0.44.1
-;; Package-Requires: ((emacs "29.1") (shell-maker "0.86.1") (acp "0.11.1"))
+;; Package-Requires: ((emacs "29.1") (shell-maker "0.86.1") (acp "0.11.1") (transient "0.9.0"))
 
 (defconst agent-shell--version "0.44.1")
 


eddof13:
I have been getting this after updating packages- it looks like it is because my default transient installed is older than the one you're expecting- this may fix that

eddof13:
also just FYI it appears in 30.2 the version is still old https://github.com/emacs-mirror/emacs/blob/emacs-30.2/lisp/transient.el 0.7.2.2

eddof13:
this might not be the only solution but it works for me anyway- otherwise I manually install transient- also could be the order in which I upgraded my packages

xenodium:
Technically, we should be ok with the built-in version.

Something's off. It doesn't look like we're using anything in agent-shell that specifically requires transient--set-layout. I'm thinking that perhaps you had a rogue .elc (agent-shell's?) and perhaps switched versions of Emacs or transient? Could you try deleting agent-shell*.elc and any transient.elc and see if you still reproduce the issue?

eddof13:
I figure as much- will try to clear out my cache, thanks

PR #377, opened 2026-03-07
Make dot-subdir storage location configurable

zackattackz:
Hello! Thank you for this awesome project. I wanted to open this PR because I dislike how the current `.agent-shell` directory functionality will pollute the repo and `.gitignore`. I wanted to provide an alternative that would instead store this data in the user emacs directory. Please let me know if anything is needed, thank you.

##
Extract path resolution from `agent-shell--dot-subdir` into a pluggable `agent-shell-dot-subdir-function` defcustom, with two built-in implementations:

- `agent-shell--dot-subdir-in-repo` (default): existing behaviour, stores data under `.agent-shell/` in the project root.
- `agent-shell--dot-subdir-user-emacs-dir`: stores data under `user-emacs-directory` instead, keeping the project tree clean.  The project path is sanitized into a flat directory name by stripping the leading slash and replacing remaining slashes with dashes.

Two further defcustoms are added: `agent-shell-dot-subdir-gitignore` (boolean, default t) to control whether `.agent-shell/` is added to `.gitignore`, and `agent-shell-user-emacs-dir-subdir` (string, default "agent-shell") to control the top-level directory name used by the user-emacs-dir implementation.

The gitignore update is only performed when the resolved path is under the project root, preserving the original behaviour of triggering only on first directory creation.

Tests are added for all three functions covering path computation, directory creation, delegation to the configured function, and the gitignore guard conditions.

## Checklist

### DIFF
--- a/README.org
+++ b/README.org
@@ -479,6 +479,35 @@ Optional: to prevent the agent running inside the container to access your local
 (setq agent-shell-text-file-capabilities nil)
 #+end_src
 
+*** Data storage location
+
+By default, agent-shell stores per-project data (transcripts, screenshots, etc.) under a =.agent-shell/= directory in the project root, and automatically adds that directory to =.gitignore=.
+
+You can change where this data is stored by customising =agent-shell-dot-subdir-function=.  Two built-in implementations are provided:
+
+- =agent-shell--dot-subdir-in-repo= (default) — stores data under =.agent-shell/= inside the project root.
+- =agent-shell--dot-subdir-user-emacs-dir= — stores data under =user-emacs-directory= instead, keeping the project tree clean.  The project path is flattened into a directory name by stripping the leading slash and replacing remaining slashes with dashes.
+
+To keep all agent-shell data out of your project trees:
+
+#+begin_src emacs-lisp
+(setopt agent-shell-dot-subdir-function #'agent-shell--dot-subdir-user-emacs-dir)
+#+end_src
+
+This will store data at a path like =~/.emacs.d/agent-shell/home-user-src-myproject/screenshots/=.
+
+The top-level directory name under =user-emacs-directory= defaults to ="agent-shell"= and can be changed:
+
+#+begin_src emacs-lisp
+(setopt agent-shell-user-emacs-dir-subdir "my-agent-shell-data")
+#+end_src
+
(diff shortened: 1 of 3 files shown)


xenodium:
Thank you for the PR.

I'm a little worried about adding 3 new defcustoms for a feature that hasn't been heavily requested. The number of defcustoms are quickly piling up from all the feature requests. Keep in mind the more public APIs I offer, the more I have to maintain and ensure I don't break user's configurations. It also increases the number of edge cases I need to consider. Having said that, I'm happy to find a middle-ground of sorts.

How about consolidting configurability into one defcustom `agent-shell-dot-subdir-function` with a single default implementation (the current behaviour). Users wanting further customisation are free to implement that function as preferred. Gitignore ebhaviour stays as is. This is a one-time deal anyway. Users can revert the change once if they prefer. Hope that makes sense.

zackattackz:
That all makes sense, apologies for cramming all these defcustoms all in one PR.

Just having `agent-shell-dot-subdir-function` for this PR sounds good, I will force push back with those changes. Do you think it'd make sense to include the user-emacs-directory example in the README where it would show how to define a custom function?

I do still think a `agent-shell-dot-subdir-gitignore` custom would be nice to have. I just do not like having packages touching my repos when it is not necessary (especially if you do not control the repo), and I think others might agree. But I can open a separate issue for this for discussion!

xenodium:
> Just having agent-shell-dot-subdir-function for this PR sounds good, I will force push back with those changes. Do you think it'd make sense to include the user-emacs-directory example in the README where it would show how to define a custom function?

Sounds good.

> I do still think a agent-shell-dot-subdir-gitignore custom would be nice to have. I just do not like having packages touching my repos when it is not necessary (especially if you do not control the repo), and I think others might agree. But I can open a separate issue for this for discussion!

It's Emacs world so I'm doomed either way. I added this feature because a user requested .gitignore support. In this case, I think keeping current behaviour is likely a good compromise for both camps. It's not a destructive operation and you can always revert it and never see the change again.

zackattackz:
That is totally fair. In the updated PR I will just add some extra comments in README + the gitignore elisp function to mention that "The gitignore entry for `.agent-shell/` is a one time operation, removing it once will cause it to not be created again" (Only because this wasn't clear to me when I was reading that recent commit that added that behavior, hence why I made all this in the first place 🤣)

zackattackz:
Closing in favor of #378

PR #378, opened 2026-03-07
Make dot-subdir storage location configurable via custom function

zackattackz:
Extract path resolution from `agent-shell--dot-subdir` into a pluggable
`agent-shell-dot-subdir-function` defcustom, with `agent-shell--dot-subdir-in-repo`
as the default implementation (existing behaviour: stores data under `.agent-shell/`
in the project root).

Tests are added for `agent-shell--dot-subdir-in-repo` and `agent-shell--dot-subdir`
covering path computation, directory creation, and delegation to the configured
function.

The README documents the new defcustom and shows a user-emacs-directory-based
custom function as an example.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -2009,7 +2009,9 @@ For example:
   "Return path to SUBDIR for agent-shell data, creating it if needed.
 Calls `agent-shell-dot-subdir-function' to resolve the path.
 When the directory is first created inside a git repo and
-.agent-shell/ is not yet ignored, automatically add it to .gitignore."
+.agent-shell/ is not yet ignored, automatically add it to .gitignore.
+This gitignore update is a one-time operation: if the entry is later
+removed from .gitignore it will not be re-added."
   (unless (functionp agent-shell-dot-subdir-function)
     (error "agent-shell-dot-subdir-function must be set to a function"))
   (let ((dir (funcall agent-shell-dot-subdir-function subdir)))
--- a/README.org
+++ b/README.org
@@ -481,7 +481,7 @@ Optional: to prevent the agent running inside the container to access your local
 
 *** Data storage location
 
-By default, agent-shell stores per-project data (transcripts, screenshots, etc.) under a =.agent-shell/= directory in the project root, and automatically adds that directory to =.gitignore=.
+By default, agent-shell stores per-project data (transcripts, screenshots, etc.) under a =.agent-shell/= directory in the project root, and automatically adds that directory to =.gitignore= the first time it is created.  This is a one-time operation: removing the entry from =.gitignore= will not cause it to be re-added.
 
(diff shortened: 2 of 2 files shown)


agent-shell.el
```diff
@@ -1985,17 +1985,34 @@ For example, shut down ACP client."
...
+  (expand-file-name (file-name-concat ".agent-shell" subdir)
+                    (agent-shell-cwd)))
+
+(defun agent-shell--dot-subdir (subdir)
+  "Return path to SUBDIR for agent-shell data, creating it if needed.
+Calls `agent-shell-dot-subdir-function' to resolve the path.
+When the directory is first created inside a git repo and
+.agent-shell/ is not yet ignored, automatically add it to .gitignore.
+This gitignore update is a one-time operation: if the entry is later
+removed from .gitignore it will not be re-added."
```
xenodium, line 2014: We should consider guarding with something like:

```emacs-lisp
(unless (functionp agent-shell-dot-subdir-function)
  (error "agent-shell-dot-subdir-function must be set to a function"))
```

agent-shell.el
```diff
@@ -1985,17 +1985,34 @@ For example, shut down ACP client."
...
+  (let ((dir (funcall agent-shell-dot-subdir-function subdir)))
```
xenodium, line 2017: Similarly another guard like:

```emacs-lisp
(let ((dir (funcall agent-shell-dot-subdir-function subdir)))
  (unless dir
    (error "Failed to resolve agent-shell data directory for %s" subdir))
  (unless (file-directory-p dir)
    (make-directory dir t)
    (agent-shell--ensure-gitignore (agent-shell-cwd)))
  dir)
```

agent-shell.el
zackattackz: Addressed in latest force-push

agent-shell.el
zackattackz: Addressed in latest force-push (I also checked if dir is a string)
Also, added tests for these errors

zackattackz: Actually may also want to fail if dir is an empty string "" ?

zackattackz: e.g `(unless (and dir (stringp dir) (not (string-empty-p dir)))`

xenodium: > Actually may also want to fail if dir is an empty string "" ?

SGTM! Maybe do:

(string-empty-p (string-trim...))

zackattackz: OK just pushed up with `(unless (and (stringp dir) (not (string-empty-p (string-trim dir))))` (removed the plain `dir` since `(stringp dir)` implies the same if dir is nil)
Also added test case for if dir is `"  "`

xenodium:
Thanks for the PR. Merged.

PR #395, opened 2026-03-11
If the response is not called fall back to asking the user

smagnuso:

## Checklist


xenodium:
Thanks for the PR!

Unfortunately, we cannot rely on the response invocation as the function can be invoke async. Having said that, I've tweaked the current implementation to consider agent-shell-permission-responder-function's return value in https://github.com/xenodium/agent-shell/commit/4561f5b63dcd99ff7517157ffe786b0c25021f38

PR #399, opened 2026-03-12
Make prompt images deletable (without selection)

cxa:
Problem:
- In `agent-shell-mode`, when the prompt contains an inline image, `C-d` with point on the image and `DEL` with point just after the image do not remove it.

Fix:
- Add explicit `agent-shell-mode` bindings for `C-d`, `DEL`, `<backspace>`, and `<delete>` so the expected delete keys are available in the mode.
- Route image deletion through `delete-field`.
- Mark only inline image attachments with a private field symbol: `agent-shell--image-field`.
- Leave non-image file attachments unchanged.

Tests:
- Added a test that inline image context is marked with the image field.
- Added a test that non-image file context does not get that field.
- Added a forward-delete test for deleting an inline image.
- Added a backward-delete test for deleting an inline image.
- Added a keybinding test for the delete keys in `agent-shell-mode`.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -998,6 +998,10 @@ END from the buffer."
 (defvar-keymap agent-shell-mode-map
   :parent shell-maker-mode-map
   :doc "Keymap for `agent-shell-mode'."
+  "C-d" #'agent-shell-delete-char
+  "DEL" #'agent-shell-backward-delete-char
+  "<backspace>" #'agent-shell-backward-delete-char
+  "<delete>" #'agent-shell-delete-char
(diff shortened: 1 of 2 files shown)


xenodium:
Thank you for the PR!

I've submitted https://github.com/xenodium/agent-shell/commit/92f99ade5d7c58fb982e12fda25bf4ec24141250 to make deletion more generic. Should work with any other binding users are used to and also cover both the shell and compose buffers. Can you give that a try and see if it works for your use case?

cxa:
Works, thanks!!!

Issue #416, opened 2026-03-16
Question: Clear Session/Start new


Makesesama:
## Checklist

- [X] *I agree to communicate with the author myself* (not AI-generated).
- [ ] I've read the README's [Filing issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#filing-issues) section.
- [X] I'm running the latest versions (fill in below).
  - agent-shell: 0.48.1
  - acp.el:
  - ACP package (e.g. claude-code-acp):
  - Agent CLI (e.g. claude, gemini):
- [ ] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [ ] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README).

## Start New Session from buffer
Just a quick question, is there currently a function like `/clear` like in claude code? There it starts a new session in the current claude-code so context is cleared. This is really helpful when working on small changes to keep the Context small and focused.
On my search i did not see anything like this in agent-shell. My current solution is to kill the buffer and open a new one, but for my workflow this adds unneeded complexity.

Thank you for the great work, i'm trying agent-shell every few weeks and it really has come a long way!

xenodium:
If `/clear` isn't exposed by the agent via ACP, then we can't use it from agent-shell. You can typically tell by inspecting:

<img width="490" height="116" alt="Image" src="https://github.com/user-attachments/assets/8d05c2c4-3dac-45ea-a804-5d000f82f18a" />

We could try adding an interactive command that kills the current shell and starts a new one in the same project. Would that do?

> Thank you for the great work, i'm trying agent-shell every few weeks and it really has come a long way!

Nice to hear. Thank you.

Makesesama:
Hey, i've checked the ACP and it is not exposed, purely client side. Opened a PR that kills the shell and starts a new one here: #421

PR #421, opened 2026-03-16
Add agent-shell-clear

Makesesama:
This adds a function that kills the current shell buffer and starts a new one with the same agent configuration.

This resolves issue #416

## Checklist

- [X] I've filed a feature request/discussion for a new feature.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -853,7 +853,7 @@ Always prompts for agent selection, even if existing shells are available."
   (agent-shell '(4)))
 
 ;;;###autoload
-(defun agent-shell-clear ()
+(defun agent-shell-restart ()
   "Clear conversation by restarting the agent shell in the same project.
 
 Kills the current shell buffer (shutting down the ACP client) and


agent-shell.el
```diff
@@ -852,6 +852,23 @@ Always prompts for agent selection, even if existing shells are available."
   (interactive)
   (agent-shell '(4)))
 
+;;;###autoload
+(defun agent-shell-clear ()
```
xenodium: Given it's not a straight up clear that folks are typically used to, we may need to name the command either restart or reset.

Makesesama: I have renamed this to `agent-shell-restart`. I was also not happy with `clear`, because there already is `agent-shell-clear-buffer`, so i think this is much better for ux aswell.

xenodium:
Thanks for renaming. Merging.

Heads-up... I'll follow up with changes, as we need to take strategy into consideration.

xenodium:
Submitted https://github.com/xenodium/agent-shell/commit/89aaccea221c79b7c0268e15574443860c4d1a6a Please check it to see if it still works for your use-case.

Makesesama:
Yep, this still works for my use-case. Thank you for finishing this up properly :)

xenodium:
Great. Thanks. Will mark #416 closed.

xenodium:
ps. Relatedly, there's agent-shell-reload too now. I've had instances where agent gets itself in a weird state. Killing and resuming session works around the issue. Reload can do both automatically now.

PR #446, opened 2026-03-23
Droid: use native acp client and support default model and mode

kohnish:

## Checklist

### DIFF
--- a/README.org
+++ b/README.org
@@ -157,13 +157,13 @@ See https://github.com/mistralai/mistral-vibe for details.
 
 *** Factory Droid
 
-For Factory Droid, install the =droid-acp= client:
+For Factory Droid, install Droid:
 
 #+begin_src bash
-npm install -g droid-acp
+curl -fsSL https://app.factory.ai/cli | sh
 #+end_src
 
-See https://github.com/yaonyan/droid-acp for details.
+See https://factory.ai for details.
 
 *** Pi
 
--- a/agent-shell-droid.el
+++ b/agent-shell-droid.el
@@ -21,7 +21,7 @@
 ;;; Commentary:
 ;;
 ;; This file includes Factory Droid specific configurations relying on the
-;; droid-acp client: https://github.com/yaonyan/droid-acp
+;; droid: https://docs.factory.ai/cli/getting-started/quickstart
 ;;
 
 ;;; Code:
@@ -73,8 +73,35 @@ For no authentication (e.g., using `droid-acp` built-in login):
   :type 'alist
   :group 'agent-shell)
 
+(defcustom agent-shell-droid-default-model-id
+  nil
+  "Default Droid AI model ID.
+
+Must be one of the model ID's displayed under \"Available models\"
+when starting a new shell."
+  :type '(choice (const nil) string)
+  :group 'agent-shell)
+
+(defcustom agent-shell-droid-default-session-mode-id
+  nil
+  "Default Droid AI session mode ID.
+
+Must be one of the mode ID's displayed under \"Available modes\"
+when starting a new shell."
+  :type '(choice (const nil) string)
+  :group 'agent-shell)
+
+(defcustom agent-shell-droid-default-reasoning-effort
+  nil
+  "Default reasoning effort.
+
+It can be one of the followings.
+none, dynamic, off, minimal, low, medium, high, xhigh, max"
+  :type '(choice (const nil) string)
+  :group 'agent-shell)
+
 (defcustom agent-shell-droid-acp-command
-  '("droid-acp")
+  '("droid" "exec" "--output-format" "acp")
   "Command and parameters for the Factory Droid ACP client.
 
 The first element is the command name, and the rest are command parameters."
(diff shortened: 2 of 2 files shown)


agent-shell-droid.el
```diff
@@ -73,8 +73,35 @@ For no authentication (e.g., using `droid-acp` built-in login):
...
+  "Default reasoning effort.
+
+It can be one of the followings.
+none, dynamic, off, minimal, low, medium, high, xhigh, max"
+  :type '(choice (const nil) string)
+  :group 'agent-shell)
+
 (defcustom agent-shell-droid-acp-command
-  '("droid-acp")
+  (list "droid" "exec" "--reasoning-effort" agent-shell-droid-default-reasoning-effort "--output-format" "acp")
```
xenodium: Using agent-shell-droid-default-reasoning-effort here will be problematic. Users may change agent-shell-droid-default-reasoning-effort but the new effort does not become active even after creating a new agent-shell session. Also setting to nil will break the command.

Let's leave this defcustom as it was and instead modify agent-shell-droid-make-client with something like:

```emacs-lisp
(agent-shell--make-acp-client :command (seq-first agent-shell-droid-acp-command)
                              :command-params (append (seq-rest agent-shell-droid-acp-command)
                                                      (when agent-shell-droid-default-reasoning-effort
                                                        (list "--reasoning-effort"
                                                              agent-shell-droid-default-reasoning-effort)))
                              :environment-variables (append ...)
                              :context-buffer buffer)
```

agent-shell-droid.el
```diff
@@ -73,8 +73,35 @@ For no authentication (e.g., using `droid-acp` built-in login):
...
+  :group 'agent-shell)
+
+(defcustom agent-shell-droid-default-reasoning-effort
+  "none"
+  "Default reasoning effort.
+
+It can be one of the followings.
+none, dynamic, off, minimal, low, medium, high, xhigh, max"
+  :type '(choice (const nil) string)
+  :group 'agent-shell)
```
xenodium, line 101: We should drop none and accept nil (default value), which signals not to include the setting at all (and document what nil does).

README.org
```diff
@@ -133,13 +133,13 @@ See https://github.com/mistralai/mistral-vibe for details.
 
 *** Factory Droid
```
xenodium, line 158: We need to update install instructions also

xenodium (commented):
Thank you for the PR. Left you some comments to address. Could you please take a look?

Also, more importantly, have you been using this setup for some time? Is it working better for you? I'm not a droid user, so I wouldn't be a good judge of it. I'm justr trying to be mindful of anyone who's been using the current server and may experience a regression in functionality. Could you please share your experience?

agent-shell-droid.el
kohnish: done!

agent-shell-droid.el
kohnish: done!

README.org
kohnish: done!

xenodium:
Thank you for the changes!

> Also, more importantly, have you been using this setup for some time? Is it working better for you? I'm not a droid user, so I wouldn't be a good judge of it. I'm justr trying to be mindful of anyone who's been using the current server and may experience a regression in functionality. Could you please share your experience?

Could you comment on this please?

kohnish:
> Thank you for the changes!
> 
> > Also, more importantly, have you been using this setup for some time? Is it working better for you? I'm not a droid user, so I wouldn't be a good judge of it. I'm justr trying to be mindful of anyone who's been using the current server and may experience a regression in functionality. Could you please share your experience?
> 
> Could you comment on this please?

Sorry, I missed your original comment.  
I've only used the original commit for a few days and ran better than droid-acp which seems to get stuck for whatever reasons.  
I'll test drive with reasoning-effort set tomorrow and let you know.

xenodium:
Thank you! I'll wait for your signal to merge.

kohnish:
> Thank you! I'll wait for your signal to merge.

All good. I used max and nil for a day. Thanks!

xenodium (approved):

xenodium:
Thank you for the PR. Merged. 

Apologies for the delay. #500 really made it tricky to spend many cycles on agent-shell in the last month. I have been slowly getting back to it as time permits. Pre #500 I was working on agent-shell daily, which enabled me to keep up with most pull requests and issues, but as you can see they are currently piling up. I started chipping at them, but it does present a challenge.

As an indie dev, I have to be super focused now on activies that help pay the bills. While I'm super appreciative of pull requests and contributions, they also take up time away from those activities. I do read/review pull requests myself and need them to be in some form that I feel comfortable accepting/maintaining. If you are in a position to get your employers to [sponsor](https://github.com/sponsors/xenodium) this project, it'd enable me to dedicate much more time to further maintain and evolve the project, which as I understand many of you use it for your work. I hear AI is now required at the workplace and agent-shell is enabling folks to continue using Emacs at work.

kohnish:
Thanks. It might be also an idea to find co-maintainers, in general it's much healthier for  projects.   
It's also very unlikely any organisation would support an indie project. I heard some companies have some legal issue or so even if they want to support it.  
I know some projects went as far as making it NPO.

PR #450, opened 2026-03-24
Fix restart using wrong default-directory

zackattackz:
`agent-shell-restart` kills the buffer and calls `agent-shell--start` directly after.
Since `kill-buffer` will have the effect of changing the default directory to the "other buffer", `agent-shell--start` can possibly end up creating the new shell an unexpected project (e.g if the user's last buffer was in a different project, the new shell will be made for this project instead of the expected behavior of being for the same project the original shell was for).

The fix is to bind `default-directory` around calling `agent-shell--start`, since this will retain the default directory across the `kill-buffer` call.

`agent-shell-reload` was also affected, since it delegates to `agent-shell-restart`.

A new test is added to ensure that, in a frame whose last buffer was another project than the current shell's, calling restart will still use the shell's project as expected. Without the fix, the test fails.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -940,7 +940,8 @@ Works from both shell and viewport buffers."
                  (not (y-or-n-p "Agent is busy.  Restart anyway?")))
         (user-error "Cancelled")))
     (kill-buffer shell-buffer)
-    (let ((new-shell-buffer (agent-shell--start
+    (let* ((default-directory (buffer-local-value 'default-directory shell-buffer))
+           (new-shell-buffer (agent-shell--start
                              :config config
                              :session-strategy strategy
                              :session-id session-id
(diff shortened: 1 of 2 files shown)


agent-shell.el
```diff
@@ -934,13 +934,15 @@ Works from both shell and viewport buffers."
                        'new-deferred
                      'new))
          (config (map-elt (buffer-local-value 'agent-shell--state shell-buffer)
-                          :agent-config)))
+                          :agent-config))
+         (shell-dir (buffer-local-value 'default-directory shell-buffer)))
```
xenodium: We can likely just do `(default-directory (buffer-local-value 'default-directory shell-buffer))` here and drop the change below.

zackattackz: Makes sense, yeah I think there was a part in the contributing guide about excluding redundant variables - my bad! I'll push that up sometime later :)

zackattackz: addressed in latest force push, also rebased onto current main!

xenodium:
Thanks for the fix. Merged.

PR #451, opened 2026-03-24
Extract agent-shell--get-available-models

zackattackz:
This is not urgent or a real issue, just a minor improvement for the code quality.

I wanted to make this simple PR for a little code cleanup since I noticed this repeated expression a lot when I was looking into model ID stuff.

Again it isn't major so I understand if you'd rather not merge, but I thought it'd be helpful to submit.

----

Consolidate repeated `(map-nested-elt state '(:session :models))` expressions into a single function, mirroring the existing `agent-shell--get-available-modes'.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -3106,7 +3106,7 @@ The model contains all inputs needed to render the graphical header."
   (let* ((model-name (or (map-elt (seq-find (lambda (model)
                                               (string= (map-elt model :model-id)
                                                        (map-nested-elt state '(:session :model-id))))
-                                            (map-nested-elt state '(:session :models)))
+                                            (agent-shell--get-available-models state))
                                   :name)
                          (map-nested-elt state '(:session :model-id))))
          (mode-id (map-nested-elt state '(:session :mode-id)))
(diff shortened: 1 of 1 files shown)


xenodium:
Thanks for thr PR! However, the current approach is intentional ;)

Knowing where things live in the state explicitly makes the codebase easier to reason with. Adding a function to access `(map-nested-elt state '(:session :models))` is a little unnecessary. To frame it a little differently, if you had a struct in another programming language, it is highly unlikely that we'd add a function that merely returns state.session.models.

zackattackz:
Fair enough it does make sense when you put it-like that with the struct comparison. You control that state anyways so its not like there would be unexpected changes due to acp stuff changing or whatnot. (If a third party package is referring to things from the state, but if those fields get changed around, is what I was thinking could be an issue. But realistically a change like that would be easy to trace and update anyways).

I will close this though, lmk if PRs like this (minor refactoring without any "real" changes) are ultimately helpful. Would not want to spam you with them when you already have a lot on your plate 😆. (Not that I had any others in mind, but just for future reference)

PR #453, opened 2026-03-24
Ensure that viewport compiles

martenlienen:
The help functions require transient at compile time for the =transient-define-prefix= macro and the transient definitions need to be at the top level to satisfy the byte compiler.

Before this fix, the help functions fail if you byte or native compile agent-shell before use.

## Checklist


xenodium (approved):
Thank you. Had been meaning to look into it.

PR #459, opened 2026-03-25
Make the limitation of region context lines optional

martenlienen:
This change lets the user decide to how many lines the embedded region context should be limited. I have set the default nil, because I personally expect the context to send the complete region when I explicitly select one, but I can of course change it to 5 as it was hard-coded before.

## Checklist

### DIFF
--- a/README.org
+++ b/README.org
@@ -650,6 +650,7 @@ always go to Evil modes if you need to with ~C-z~).
 | agent-shell-qwen-acp-command                  | Command and parameters for the Qwen Code client.                                |
 | agent-shell-qwen-authentication               | Configuration for Qwen Code authentication.                                     |
 | agent-shell-qwen-environment                  | Environment variables for the Qwen Code client.                                 |
+| agent-shell-region-context-max-lines          | Limit region context to this many lines                                         |
 | agent-shell-screenshot-command                | The program to use for capturing screenshots.                                   |
 | agent-shell-section-functions                 | Abnormal hook run after overlays are applied (experimental).                    |
 | agent-shell-session-strategy                  | How to handle sessions when starting a new shell.                               |
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -291,6 +291,11 @@ passed through to `acp-make-client'."
                                                    (map-elt (buffer-local-value 'agent-shell--state context-buffer)
                                                             :outgoing-request-decorator)))))
 
+(defcustom agent-shell-region-context-max-lines nil
+  "Limit region context to at most this many lines if non-nil."
+  :type '(choice (const nil) integer)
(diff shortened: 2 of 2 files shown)


xenodium:
Thank you for the PR.

> This change lets the user decide to how many lines the embedded region context should be limited. I have set the default nil, because I personally expect the context to send the complete region when I explicitly select one,

It's capped as a preview so you get a peek at what you're sending, but the agent still knows that you are inerested in the full range context, which is specified in the file path (along with line range).

> but I can of course change it to 5 as it was hard-coded before.

Yes please. The cap was introduced because it was too noisy if selection was large.

martenlienen:
Yes, I noticed that too after using it a bit now and I can see that there are cases where I just want to give the agent a pointer. My feeling at the moment is that when I select a region, I also want to send the full region, but that is noisy in the UI. What do you think about sending the full region, but hiding some of it in the agent-shell UI with an overlay that hides lines 6+?

xenodium:
> What do you think about sending the full region, but hiding some of it in the agent-shell UI with an overlay that hides lines 6+?

Curious, for your use-case does it matter if it's sent inline or not? The agent already knows the lines that it needs to look at encoded in the file path (eg. path/to/agent-shell.el:5771-5775). If anything, I'm wondering if relying on the existing behaviour is more efficient as it enables the agent to decide what's the best way to look at those lines (maybe they are too long and thus adjusts its streategy).

martenlienen:
So my situation is that I have a piece of text in an org file and I have a custom context function to send the current subtree. I want to refine the text with feedback from the agent. If I rely on the file link, I have to rely on the agent to deduce to read the file (which it sometimes does not and then gives me the same feedback again) and then wait for the tool use. Just sending the subtree directly as text to the agent is quicker in this situation and more reliable.

xenodium:
Ok, I've submitted https://github.com/xenodium/agent-shell/commit/5fd9794c7feb3030d9e8352ff74c64e3a1ddb1ad which likely caters for both worlds. Default behaviour remains as is, but you now have an expand button (actually replaces the preview with full text) to get the entire text if needed (and edit if needed).

2026-03-25-16:13:05-Emacs_x0 8_optimized

martenlienen:
That looks amazing! I will close this PR for now and report back if I have another idea for how to improve it.

xenodium:
> That looks amazing!

Glad you like it. If you or your employer are in a position to do so, please consider [sponsoring](https://github.com/sponsors/xenodium). I regularly get requests like this one and I work with submitter to find a suitable solution. I'm working daily on agent-shell to keep up, but it's not quite sustainable as indie dev (this is time I could have spent on more sustainable activities).

PR #463, opened 2026-03-25
Fix header text invisible when font-get :size returns 0

### LINKED ISSUE #462: Header text invisible when font-get :size returns 0
## Description

The graphical header SVG renders the agent name, model, mode, and project directory text at `font-size="0"`, making it invisible. The Anthropic logo icon renders fine, but all text is missing.

## Root cause

In `agent-shell--make-header-model`, the font size is determined by:

```elisp
(:font-size . ,(when-let* (((display-graphic-p))
                           (font (face-attribute 'default :font))
                           ((fontp font)))
                 (font-get font :size)))
```

`(font-get font :size)` returns `0` when the font is configured via `:height` (in 1/10pt units) rather than an explicit pixel size. This is common on macOS — for example, Menlo at `:height 120` (12pt) reports `:size 0` even though `frame-char-height` correctly returns `14`.

The `0` propagates into the SVG `<text>` elements as `font-size="0"`, rendering all header text invisible.

## Reproduction

```elisp
;; If your default font was set via :height rather than pixel size:
(font-get (face-attribute 'default :font) :size) ;; => 0
(frame-char-height) ;; => 14 (correct)
```

## Environment

- macOS (Darwin 24.6.0)
- Emacs 30.2 (GUI)
- agent-shell 0.50.1
- Font: Menlo, set via `:height 120`

## Suggested fix

Fall back to `frame-char-height` when `font-get :size` returns 0:

```elisp
(:font-size . ,(or (when-let* (((display-graphic-p))
                               (font (face-attribute 'default :font))
                               ((fontp font))
                               (size (font-get font :size))
                               ((> size 0)))
                     size)
                   (frame-char-height)))
```

## Workaround

```elisp
(with-eval-after-load 'agent-shell
  (advice-add 'agent-shell--make-header-model :filter-return
              (lambda (model)
                (when (and model (zerop (or (map-elt model :font-size) 0)))
                  (map-put! model :font-size (frame-char-height)))
                model)))
```

ftlio:
## Summary

- `font-get :size` returns `0` when the default font is configured via `:height` (1/10pt units) rather than an explicit pixel size — common on macOS (e.g., Menlo at `:height 120`)
- The `0` propagates into the SVG `<text>` elements as `font-size="0"`, rendering all header text (agent name, model, mode, project directory) invisible while the icon still displays
- Falls back to `frame-char-height` when `font-get :size` returns 0

Fixes #462


xenodium (approved):
Thanks for the fix

xenodium:
FYI: Tiny follow-up https://github.com/xenodium/agent-shell/commit/68ce6b5cd51cb0f07c3849767cf6455edb8fb992

ftlio:
> Thanks for the fix

Thanks for the plugin!

PR #469, opened 2026-03-26
Do not create a file if no image in Wayland clipboard

### LINKED ISSUE #468: wl-paste-image-handler breaks pasting text for me
## Checklist

- [ ] I'm running the latest versions (fill in below).
  - agent-shell version: 68ce6b5cd51cb0f07c3849767cf6455edb8fb992
  - acp.el version: c32fbf8df34ed0095853a8cf55dc783e68b67d90
  - ACP package (e.g. claude-agent-acp) version: 0.24.1
  - Agent CLI (e.g. claude, gemini) version: 2.1.84
- [ ] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README).

The changes in https://github.com/xenodium/agent-shell/commit/b7b27b7ab3eb8117269b1872b2ad128cb2d96e2b seem to break pasting any text in the agent shell buffer for me.

Yanking any text results in some square appearing, like so:

Image

The actual image it's trying to insert is a file containing the following raw text:
```
Clipboard content is not available as inferred output type "image/png"
Use "wl-paste --list-types" to view available types. Use "--type" to explicitly specify a type.
```

Removing the new wl-paste entry from agent-shell-clipboard-image-handlers fixes the problem

martenlienen:
The old call to ~call-process~ would merge stdout and stderr into a single stream that would always create ~file-path~ even if no image was in the clipboard. This could then lead downstream code to erroneously assume that an image had been found even though an error was signaled.

Fixes #468.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -395,9 +395,12 @@ Assume screenshot file path will be appended to this list."
   (list
    (list (cons :command "wl-paste")
          (cons :save (lambda (file-path)
-                       (let ((exit-code (call-process "wl-paste" nil `(:file ,file-path))))
-                         (unless (zerop exit-code)
-                           (error "Command wl-paste failed with exit code %d" exit-code))))))
+                       (with-temp-buffer
+                         (let* ((coding-system-for-read 'binary)
+                                (exit-code (call-process "wl-paste" nil (list t nil) nil "--type" "image/png")))
+                           (if (zerop exit-code)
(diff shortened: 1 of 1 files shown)


xenodium (approved):
Thank you for the fix

PR #474, opened 2026-03-27
Display qualifier key binding help in text style header

bcc32:
Previously, the qualifier (e.g., "[1/5][Edit]") and key binding hints
(e.g., "C-c C-c Submit") were only rendered in graphical-style headers.
This meant that users with agent-shell-header-style set to 'text would
not see any key documentation in the header line.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -3217,44 +3217,58 @@ When provided, included in help-echo tooltips."
   (unless state
     (error "STATE is required"))
   (let* ((header-model (agent-shell--make-header-model state :qualifier qualifier :bindings bindings))
-         (text-header (format " %s%s%s @ %s%s%s%s"
-                              (propertize (map-elt header-model :buffer-name)
-                                          'font-lock-face 'font-lock-variable-name-face)
-                              (if (map-elt header-model :model-name)
-                                  (concat " ➤ " (propertize (map-elt header-model :model-name)
-                                                            'font-lock-face 'font-lock-negation-char-face
-                                                            'help-echo (concat "Click to open LLM model menu "
-                                                                               (when model-binding
-                                                                                 (propertize model-binding 'face 'help-key-binding)))
-                                                            'mouse-face 'mode-line-highlight
-                                                            'local-map (let ((map (make-sparse-keymap)))
-                                                                         (define-key map [header-line mouse-1]
-                                                                                     (agent-shell--mode-line-model-menu))
-                                                                         map)))
-                                "")
-                              (if (map-elt header-model :mode-name)
-                                  (concat " ➤ " (propertize (map-elt header-model :mode-name)
-                                                            'font-lock-face 'font-lock-type-face
-                                                            'help-echo (concat "Click to open session mode menu "
-                                                                               (when mode-binding
-                                                                                 (propertize mode-binding 'face 'help-key-binding)))
-                                                            'mouse-face 'mode-line-highlight
-                                                            'local-map (let ((map (make-sparse-keymap)))
-                                                                         (define-key map [header-line mouse-1]
-                                                                                     (agent-shell--mode-line-mode-menu))
-                                                                         map)))
-                                "")
-                              (propertize (map-elt header-model :project-name) 'font-lock-face 'font-lock-string-face)
-                              (if (map-elt header-model :session-id)
-                                  (concat " ➤ " (map-elt header-model :session-id))
-                                "")
-                              (if (map-elt header-model :context-indicator)
-                                  (concat (if (> (length (map-elt header-model :context-indicator)) 1) " ➤ " " ")
-                                          (map-elt header-model :context-indicator))
-                                "")
-                              (if (map-elt header-model :busy-indicator-frame)
-                                  (map-elt header-model :busy-indicator-frame)
-                                ""))))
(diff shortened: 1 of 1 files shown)


xenodium:
Thanks for the PR!

I'm a viewport and graphical header user myself. Been trialing the text header to see if I'd swap more permanently and also noticed it needs to catch up with graphical. Thanks for kicking this off.

AFAIK, the text header has a single-line rendering limitation, or do you have additional settings? In most instances, a large chunk of the single-line text header is likely going to be hidden unless the window is large enough. The proposal currently looks as follows for me:

header-text

I'm thinking we may need to get a little creative with the text header line to compact things in. Here are some ideas.

## 1. States

For example:

`[2/2][View] Claude ➤ Default (recommended) ➤ Default @ project ➤ 23k/1m (2%) ➤ [?] Help`

pressing [h] switches header to:

`[2/2][View] Claude ➤ [n] Next [p] Previous [r] Reply [h] Next [h] Next  `

## 2. Transient

`[2/2][View] Claude ➤ Default (recommended) ➤ Default @ project ➤ 23k/1m (2%) ➤ [?] Help`

pressing [h] opens `agent-shell-viewport-help-menu`.

## 3. Split between header and mode line

See if we can potentially split information and render some of it in the header and remaining in mode line.

It's hard to tell which one of these would work out best in practice unless we try them out. I understand this may be more than you had originally signed up for in the PR, so I'm happy to play with the alternatives if you prefer to defer the work.

ps. Kinda glad you ran into this bug ;) It tells me there are folks out there using the viewport interaction. I have a feeling there are only a few of us.

bcc32:
Thanks for these suggestions!

> 
> AFAIK, the text header has a single-line rendering limitation, or do you have additional settings?

No, indeed, just a single line.  I agree it can get pretty crowded.

> In most instances, a large chunk of the single-line text header is likely going to be hidden unless the window is large enough. The proposal currently looks as follows for me:
> 
> header-text
> I'm thinking we may need to get a little creative with the text header line to compact things in. Here are some ideas.
> 
> ## 1. States
> For example:
> 
> `[2/2][View] Claude ➤ Default (recommended) ➤ Default @ project ➤ 23k/1m (2%) ➤ [?] Help`
> 
> pressing [h] switches header to:
> 
> `[2/2][View] Claude ➤ [n] Next [p] Previous [r] Reply [h] Next [h] Next `

I think it might be a bit too precedent-breaking to have the header line be stateful in this way.
I guess it's somewhat analogous to how which-key-mode works, but in any case I think it's not
the ideal way for this to work.

> ## 2. Transient
> `[2/2][View] Claude ➤ Default (recommended) ➤ Default @ project ➤ 23k/1m (2%) ➤ [?] Help`
> 
> pressing [h] opens `agent-shell-viewport-help-menu`.

This would definitely be an improvement, but I think isn't quite optimal for my use case.

Let me add a bit more context:

At my site, we've implemented a command which uses agent-shell-viewport-edit-mode in a completely fire-and-forget sort of way.  The implementation rebinds `C-c C-c` and `C-c C-k` to commands that do their normal function, but also delete the window displaying the viewport buffer.  So, the workflow looks like:

1. Invoke fire-and-forget prompt command, which creates a new viewport buffer and calls `pop-to-buffer` to select it in a new window.
2. Type prompt in the viewport buffer
3. Press `C-c C-c`; we make this call `agent-shell-queue-request` so that the user does not have to know whether the agent is currently working or not, or wait for it to be done.
4. After `C-c C-c` or `C-c C-k`, restore the window configuration from before step 1.

So, in some sense, this command hijacks the viewport implementation in order to have:
- Header line information
- Easy access to the correct "place" to send the request
- `agent-shell--context`

But in this case, we're optimizing for the workflow where users don't really care to read the actual output from the agent, and instead just prompt it to make changes to files on disk and read the diffs there.

> ## 3. Split between header and mode line
> See if we can potentially split information and render some of it in the header and remaining in mode line.

Alternatively, one could imagine putting some minimal help info in a comment (which would need to be stripped out before sending to the agent), or in an overlay with a `before-string` property.

> It's hard to tell which one of these would work out best in practice unless we try them out. I understand this may be more than you had originally signed up for in the PR, so I'm happy to play with the alternatives if you prefer to defer the work.
> 
> ps. Kinda glad you ran into this bug ;) It tells me there are folks out there using the viewport interaction. I have a feeling there are only a few of us.

xenodium:
Finally had a chance to look more into this... 

The queuing flow you're describing is effectively a feature we're missing in the viewport flow `(setq agent-shell-prefer-viewport-interaction t)`. Now added in https://github.com/xenodium/agent-shell/commit/76459c263014d8f7e96366e3125131f8121ab574

I've also tweaked the header a bit in https://github.com/xenodium/agent-shell/commit/8757b5cb50df0777c6d9b8fbd7ace91c10bb29f7

Screenshot 2026-06-04 at 09 52 28

`?` shows:

Screenshot 2026-06-04 at 09 53 30

Screenshot 2026-06-04 at 09 53 05

`C-c C-h` shows:

image

Give the viewport flow a try. IMO you'll get the flow that you're after but also benefit from the single-key bindings that enable quicker interaction as well as navigation.

Try with `(setq agent-shell-prefer-viewport-interaction t)`

PR #478, opened 2026-03-29
viewport pin page

zackattackz:
Introduce the following:

1. New command `agent-shell-viewport-pin-page` for use in `agent-shell-viewport-view-mode`. This command will open the currently viewed page in a read-only mode in some other window (in `special-mode` so you can quit with "q"). This is useful for referencing previous turns without the constant need to switch between pages or manually copy/pasting output.

2. Custom variable `agent-shell-viewport-pin-page-switch-buffer-function`. The function used to display the pinned page. Defaults to `#'display-buffer`.

3. Map `agent-shell-viewport-pin-page` to `"d"` in `agent-shell-viewport-view-mode-map`.

4. Display this binding in the help page of viewport view mode.
---

I understand if you don't wish to merge this, it could be implemented in another package. But it is a really small amount of code for something I think is essential for any serious usage of viewport mode (Unless there is some alternative workflow I am missing). Thank you for considering!

## Demo
[pin-demo.webm](https://github.com/user-attachments/assets/4ac52b42-a4c3-4f8b-8d46-364b39689f29)

## Checklist

### DIFF
--- a/README.org
+++ b/README.org
@@ -607,86 +607,115 @@ always go to Evil modes if you need to with ~C-z~).
 #+END_SRC
 
 #+RESULTS:
-| Custom variable                               | Description                                                                     |
-|-----------------------------------------------+---------------------------------------------------------------------------------|
-| agent-shell-agent-configs                     | The list of known agent configurations.                                         |
-| agent-shell-anthropic-authentication          | Configuration for Anthropic authentication.                                     |
-| agent-shell-anthropic-claude-acp-command      | Command and parameters for the Anthropic Claude client.                         |
-| agent-shell-anthropic-claude-command          | Command and parameters for the Anthropic Claude client.                         |
-| agent-shell-anthropic-claude-environment      | Environment variables for the Anthropic Claude client.                          |
-| agent-shell-anthropic-default-model-id        | Default Anthropic model ID.                                                     |
-| agent-shell-anthropic-default-session-mode-id | Default Anthropic session mode ID.                                              |
-| agent-shell-auggie-acp-command                | Command and parameters for the Auggie client.                                   |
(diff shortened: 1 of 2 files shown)


zackattackz:
Also to note: I regenerated the README's "Customizations" and "Commands" by running the source blocks in there. I glanced at the output to make sure it looks good (it does) but I did not do a deep comparison of those changes - so there may be more commands/variables added in those blocks than just the new ones if the README was stale.

xenodium:
Thanks for the PR.

We have something somewhat similar by pressing `C-c C-p` to go to the previous page. You can select any region and press `r` to reply and that region is automatically copied over. Does that cover your use case?

zackattackz:
Heyo, thank you for checking this out.

> Does that cover your use case?

Is the text copied this way included in the actual reply to the agent or is it automatically excluded? If not excluded it seems inefficient if my purpose is just to reference prior pages.

Also, the fonts + markdown rendering is not done through this method. Collapse-able sections are expanded.

Either way I still prefer it to be copied to another buffer in another window for ease of reference. The UX is a bit nicer if you are pinning multiple windows though, it allows the user to quickly pin several pages and organize them to their liking (through display buffer alist or through the custom variable). Perhaps this could be better integrated with the method you suggest though?

zackattackz:
Honestly after using this for some more days I think while it is nice to have, there's no need to merge to agent shell. Reason is there is lot of room for opinions on how the pages should be pinned, and it could be implemented better with indirect buffers rather than just copying everything over like this does. Could also arguably get it's own mode defined, and then you have a lot of extra code to maintain...

So I don't think it really should be included, so I will close this. (Though it is totally up to you in the end if you want to re-open this!)

That said I think it'd be cool if we had an active github discussions page or something like it. For people to share their more opinionated configs, like this, around to get inspiration and tips :)

PR #503, opened 2026-04-06
Add agent-shell-macext to related projects

cxa:
Adds <https://github.com/cxa/agent-shell-macext> to the related projects list in README.org.


xenodium (approved):
Lovely. Thank you!

Issue #500, opened 2026-04-03
📣 Expect temporary delays responding to PR/Issues


xenodium:
https://xenodium.com/and-then-there-were-three

zackattackz:
Congrats! :)

tarushnagpal:
Congratulations! 🎊

lima-limon-inc:
Congratulations!!!!!! Wish you the best ❤️

timfel:
Congratulations 🎉 You're in for a ride 🙂

robjgray:
Wonderful!

HIRANO-Satoshi:
Congrats!! Have a great time!

xenodium:
Thank you all for the wishes! It's been wild and winderful at the same time. I love this little guy so much. Exciting times.

Gonna close this announcement now. I've been finding pockets of time here and there to chip at the project. Hope you're enjoying the latest changes.

PR #513, opened 2026-04-09
Quote reply to the complete response with R

martenlienen:
Add a keybinding `R` in view-mode to quote reply to the complete response without having to mark it first. This is useful when one wants to respond to many points by the agent in detail.

This is a follow-up to #512, so includes those changes as well. It makes sure not to show the response as context at the top, because it is all quoted anyway.

## Checklist

### DIFF
--- a/agent-shell-viewport.el
+++ b/agent-shell-viewport.el
@@ -662,6 +662,34 @@ With EXISTING-ONLY, only return existing buffers without creating."
               (agent-shell-viewport-edit-mode)
               (current-buffer))))))))
 
+(defun agent-shell-viewport--block-quote (text)
+  "Prefix each line of TEXT with \"> \"."
+  (concat "> " (replace-regexp-in-string "\n" "\n> " text)))
+
+(cl-defun agent-shell-viewport--setup-reply (&key reply-context quoted-text)
+  "Set up the buffer to compose a reply.
+
+REPLY-CONTEXT is shown as read-only context at the top.  QUOTED-TEXT is
+inserted as a block quote as part of the reply."
+  (with-current-buffer (agent-shell-viewport--shell-buffer)
+    (goto-char (point-max)))
+  (let ((had-snapshot agent-shell-viewport--compose-snapshot))
+    (agent-shell-viewport-edit-mode)
+    (agent-shell-viewport--initialize)
+    (agent-shell-viewport--restore-compose-snapshot :reply-context reply-context)
+    (when quoted-text
+      (goto-char (point-max))
+      (insert (if had-snapshot "\n\n" "")
+              (agent-shell-viewport--block-quote quoted-text) "\n\n"))
+    ;; Skip past any cursor-intangible layout text (e.g. the
+    ;; newline inserted by `agent-shell-viewport--initialize')
+    ;; so callers like `agent-shell-viewport-reply-1' can insert.
+    (goto-char (if (or had-snapshot quoted-text
+                       (and reply-context (not (string-empty-p reply-context))))
+                   (point-max)
+                 (or (next-single-property-change (point-min) 'cursor-intangible)
+                     (point-max))))))
+
 (defun agent-shell-viewport-reply ()
   "Reply as a follow-up and compose another prompt/query."
   (declare (modes agent-shell-viewport-view-mode))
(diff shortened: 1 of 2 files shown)


agent-shell-viewport.el
```diff
@@ -1048,6 +1114,9 @@ VIEWPORT-BUFFER is the viewport buffer to check."
                       '(((:function . agent-shell-viewport-reply)
                          (:description . "Reply…")
                          (:if-not . agent-shell-viewport--busy-p))
+                        ((:function . agent-shell-viewport-quote-reply)
```
xenodium, line 1068: I don't seem to find agent-shell-viewport-quote-reply. Am I missing a piece from the PR?

xenodium (commented):
Thanks for the PR.

Looks like a nice addition. Went to try it out, but I'm missing agent-shell-viewport-quote-reply. Am I missing a piece from PR?

martenlienen:
Sorry for that oversight. I lost the function while juggling all of these PRs in my local repository and moving a patch between them. Now everything should be here.

martenlienen:
Hey @xenodium did you have a chance to look at this again? This PR and the included changes from #512 are features that I rely on every day and they improve the viewport flow significantly for me, so I think it would be great for the community to have them merged.

xenodium:
Thanks for the ping and apologies for the delay. I'm gradually getting back to agent-shell work, but still finding my footing due to #500.

I had a play with it. I really like `R` and  `agent-shell-viewport-quote-reply` but found that #512 makes my reply buffers a little busy by default. I'm gonna use it a little more to see if I can get used to it, but gut feeling at the moment is to keep #513 and possibly drop #512. At present, the compose buffers can easily access previous pages and even remember drafts if you happen to want to go back and forth.

martenlienen:
I agree that it the look is not as nice as before where the reply buffer is completely clean. However, I reference the last agent message so often to write my reply that it is worth it to me to have it always visible without having to peek.

My mental model is as follows. If we denote user messages as `U` and agent messages as `A`, viewport lets us focus on `[U A]` pairs of messages in a long chat `U A U A U A [U A]`. With #512, we are only moving forward a half step `[A U']` while drafting message `U'` instead of having the viewport at `[U' _]` with a non-existing agent response. I have also thought about advising f and b to allow to navigate through the history by half steps in general, i.e. make `U A U [A U] A U A` a legal viewport state, but I have not tried that yet.

I think what makes sense for the user depends on how they think about the agent interaction. If the mental model is that the user sends a request and the model a response, then it makes sense to only allow to look at `[U A]` pairs of messages. But if the user message can also be a response to what the agent said, then it also makes sense for the user to want to look at `[A U]` pairs in the viewport.

xenodium:
Thanks for the additional details. I've given it a change and used it for a few days.

While the quick access to previous response is nice, it comes with the busy response buffer tradeoff I mentioned earlier. The majority of the time, the previous response text is not needed, so it feels like we'd be optimising for a secondary flow instead of the primary one.

I'm happy to merge `R/agent-shell-viewport-quote-reply` without the reply at top change if you're still keen to get that change in.  Sorry, I gave it a good try but it just didn't feel right.

martenlienen:
I have isolated the `R` change in this PR with a tiny fix that trims whitespace from the front and end of a quoted reply when you `r`. Often, when I select a paragraph, I also select the final newline of the paragraph, which means that the response quote ends with `\n> `, which does nothing. Trimming whitespace before quoting cleans this up.

I will continue to maintain the other patch privately then, because in my workflow this is useful basically all the time :) Would you be open to a PR that makes this configurable? Maybe also more generally allows you to switch between navigating in full steps vs half steps in the terminilogy of my previous comment?

xenodium (approved):

xenodium:
> I have isolated the R change in this PR

Thank you. Merged it.

>  tiny fix that trims whitespace from the front and end of a quoted reply when you r. Often, when I select a paragraph, I also select the final newline of the paragraph, which means that the response quote ends with \n> , which does nothing. 

Good one!

> Would you be open to a PR that makes this configurable?

My concern here is additional complexity. If we are able to keep the logic fairly isolated/simple, then I'd be open to it, but would have to have a look at some code to make a call. Sorry, I know this places some work on you without guarantees of being merged. Also, there was a small bug. If you reply and immediately cancel, you're prompted if you'd like to discard the buffer content (but nothing was entered by user).

martenlienen:
No worries, I will keep mulling it over a bit in my head and if I find a good solution without too much additional complexity, I will send a PR.

PR #514, opened 2026-04-09
Generate repetitive viewport bindings with a macro

martenlienen:
A minor refactoring to eliminate this repetitive code.

## Checklist

### DIFF
--- a/agent-shell-viewport.el
+++ b/agent-shell-viewport.el
@@ -731,8 +731,8 @@ buffer from the snapshot and switch to edit mode."
   (let* ((shell-buffer (or (agent-shell--current-shell)
                            (user-error "Not in an agent-shell buffer")))
          (viewport-buffer (agent-shell-viewport--buffer
-                          :shell-buffer shell-buffer
-                          :existing-only t)))
+                           :shell-buffer shell-buffer
+                           :existing-only t)))
     (with-current-buffer shell-buffer
       (agent-shell-set-session-model
        (lambda ()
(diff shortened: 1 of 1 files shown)


xenodium:
Thank you Marten.

I really appreciate the desire to remove repetitive code. Over time though however, I've grown to have some tolerance for a little repetition, when considering the tradeoffs to consolidate.

In this particular instance, we're using a macro to generate a bunch of small functions. While we are indeed removing duplicate code, it's small/simple code. With a macro, we give up a couple of maintenance goodies:

1. We can no longer search the codebase for something like `agent-shell-viewport-reply-yes` and find the function. This is important for newcomers, and probably me, when I forget that these functions/commands are special.
2. We can no longer set breakpoints and step through the code at runtime using edebug (debugger).

With that in mind, I'm sorry to say I'm a little hesitant to merge this PR. Hope the above makes sense.

martenlienen:
I can totally see that point. At first, I tried a small utility `--response-and-send`, but removed barely any code. It is of course also fine to leave it as is.

By the way, I just tacked this small change onto this PR that fixes the indentation of a few lines. I use aphelia-mode to format buffers on save, which also ensures that elisp files are always correctly indented. That made me notice that several files have lines where the indentation is off, usually by a single space. Do you prefer fixing the indentation or keeping ill-formatted lines as they are for better blaming?

xenodium:
Thanks for this. If possible, I generally prefer fixing indent when touching the same lines. Having said that, I'm fairly flexible. Ideally maybe we apply formatting changes in a separate commit.

I'm generally indenting via select/tab using the elisp mode defaults. If aphelia produces the same result, SGTM.

martenlienen:
Yes, it does. I can recommend it. In the end, it [just calls `lisp-indent` on the buffer on save](https://github.com/radian-software/apheleia/blob/e6e5d5523d229735ab5f8ec83e10beefcfd00d76/apheleia-formatters.el#L349).

martenlienen:
I was too quick to speak. `lisp-indent` is not a function. It calls [`apheleia-indent-lisp-buffer`](https://github.com/radian-software/apheleia/blob/e6e5d5523d229735ab5f8ec83e10beefcfd00d76/apheleia-formatters.el#L1235), which also does the right thing by calling `indent-buffer` on the buffer after some setup.

PR #515, opened 2026-04-09
Make detection of binary files more robust

martenlienen:
The previous heuristic would try to embed non-image files as text, which would fail, for example, for small PDFs. For those files, which contain non-UTF-8 bytes, `json-serialize` would raise an error, because it would be an invalid UTF-8 string.

The new function has more robust MIME type detection based on Emacs' `mailcap` and ensures that only files with valid multi-byte content will not be base64 encoded.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -45,6 +45,7 @@
 (require 'dired)
 (require 'diff)
 (require 'json)
+(require 'mailcap)
 (require 'map)
 (unless (require 'markdown-overlays nil 'noerror)
   (error "Please update 'shell-maker' to v0.90.1 or newer"))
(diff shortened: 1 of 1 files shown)


agent-shell.el
```diff
@@ -4632,30 +4633,48 @@ Returns an alist with:
...
-                        (insert-file-contents file-path)
-                        (buffer-string))))))
+         (is-binary nil)
+         (content
+          (unless shallow
+            ;; If we have a known MIME type that is not explicitly a text type
+            (if (and mime-type
+                     (not (or (string-prefix-p "text/" mime-type)
+                              (string-match-p "\\(?:json\\|xml\\|yaml\\)$" mime-type))))
+                ;; Read contents as base-64 encoded binary
```
xenodium: Won't small non-image binary files (PDFs, .docx, etc.) that pass the size check will now be embedded as `"resource"` blocks with base64-encoded content in the `text` field? Before this PR, `json-serialize` would error on non-UTF-8 bytes and the `condition-case` would catch it, falling back to a plain text mention, somewhat accidentally the right behavior.

Maybe resource embed branch in the caller probably needs a `(not (map-elt file :base64-p))` guard so binary non-image files fall through to `resource_link` instead?

xenodium (changes requested):
Thanks for the PR and making this area more robust.

Had a quick look (apologies under tight time constraints at the moment due to #500).

Left a comment. Mind looking into it?

agent-shell.el
martenlienen: I worked on this, because I @-mentioned a small PDF file that was then embedded in the message but not base64 encoded. agent-shell then refused any further cooperation, when sending the request failed on `json-serialize` in acp.

agent-shell.el
```diff
@@ -4589,13 +4590,13 @@ Returns list of alists with :start, :end, and :path for each mention."
...
                       content-blocks))
-               ;; Text file, small enough, text file capabilities granted and embeddedContext supported
+               ;; Small enough, text file capabilities granted and embeddedContext supported
                ;; Use ContentBlock::Resource
                ((and agent-shell-text-file-capabilities supports-embedded-context (map-elt file :size)
                      (< (map-elt file :size) agent-shell-embed-file-size-limit))
                 (push `((type . "resource")
                         (resource . ((uri . ,(concat "file://" resolved-path))
-                                     (text . ,(map-elt file :content))
+                                     (,(if (map-elt file :base64-p) 'blob 'text) . ,(map-elt file :content))
```
martenlienen, line 5278: This embeds small non-text files as binary resources now as documented [here](https://agentclientprotocol.com/protocol/content#embedded-resource). And I think at least claude would also support embedded PDFs, if [these API docs](https://platform.claude.com/docs/en/build-with-claude/files) also apply to ACP.

Regarding `and agent-shell-text-file-capabilities` above, why is this necessary to work with embedded resources? From the naming and the pattern in the image case, I would think that `supports-embedded-context` would be the right test to check if the agent can deal with embedded context.

agent-shell.el
```diff
@@ -4923,30 +4924,48 @@ Returns an alist with:
...
-                      (if is-binary
-                          (progn
-                            (insert-file-contents-literally file-path)
-                            (base64-encode-string (buffer-string) t))
-                        (insert-file-contents file-path)
-                        (buffer-string))))))
+         (is-binary nil)
+         (content
+          (unless shallow
+            ;; If we have a known MIME type that is not explicitly a text type
```
xenodium: Won't we get false positives here for a lot of text cases?

```
file                           mailcap mime              classified as
/tmp/pr515-demo.el             application/emacs-lisp    BINARY (base64)
/tmp/pr515-demo.sh             application/x-sh          BINARY (base64)
/tmp/pr515-demo.py             nil                       text
/tmp/pr515-demo.json           nil                       text
/tmp/pr515-demo.md             nil                       text
```

How about we consider peeking at file to determine if binary and use mailcap for mime-type? Maybe something like:

```emacs-lisp
(defun agent-shell--file-binary-p (file-path)
  "Return non-nil if FILE-PATH appears to be binary.
Sniffs for a NUL byte in the first 4 KB — the same fast-path
heuristic used by file(1), git, and grep -I.  Reads at most 4 KB
regardless of file size."
  (with-temp-buffer
    (set-buffer-multibyte nil)
    (insert-file-contents-literally file-path nil 0 4096)
    (goto-char (point-min))
    (search-forward "\0" nil t)))
```

xenodium (commented):
Sorry for the delay. Just saw you made further changes. Thank you!

agent-shell.el
xenodium: > Regarding and agent-shell-text-file-capabilities above, why is this necessary to work with embedded resources?

IIRC, this was somewhat relevant to running agents in a container and preventing access to files outside the container, but this is a moot point since we're crafting a prompt so the user intent is indeed to send the file. We can likely remove it.

agent-shell.el
martenlienen: Good idea! I have implemented that and simplified the code in general.

xenodium (approved):
Thanks for the PR and follow-up changes.

xenodium:
Some small tweaks in https://github.com/xenodium/agent-shell/commit/314d82474591f0b3bc12bb61684b5a25414c9cea

martenlienen:
Thanks! Why did you drop `(set-buffer-multibyte nil)`? Isn't that required for correct handling of binary files?

xenodium:
That's my mistake. It's back in. Thanks for catching it.

PR #518, opened 2026-04-12
Add hook to customize @ file completion

nohzafk:
# Problem

@ completion is hardcoded to CAPF/projectile, preventing fuzzy or alternative backends. 

Without corfu/company it faills through to *Completions* buffer with prefix mathching.

# Solution

- add `agent-shell-file-completion-function` defcustom to handle @ completion; if it returns non-nil, CAPF is skipped.
- Default behavior remains unchanged (CAPF still used when the custom function returns nil).

# My Hack
Because I dont' use courfu/company,  currently I use this hack to make agent-shell completion to use with another backend, it would be nice that agent-shell provide an option for user to provide custom function

```elisp
;; Fuzzy file completion for @ references using consult-snapfile.
    ;; Replaces prefix-based CAPF with minibuffer fuzzy search via vertico.
    (defun backbone/agent-shell-fuzzy-insert-file ()
      "Insert a file reference using consult-snapfile fuzzy matching.
On cancel, inserts bare @ for manual typing."
      (interactive)
      (if (not (fboundp 'consult-snapfile--async-source))
          (progn (self-insert-command 1 ?@)
                 (completion-at-point))
        (let* ((root (consult-snapfile--project-root))
               (default-directory root)
               (prompt (format "@ File [%s]: " (abbreviate-file-name root)))
               (selected (condition-case nil
                             (consult--read
                              #'consult-snapfile--async-source
                              :prompt prompt
                              :sort nil
                              :require-match t
                              :category 'file
                              :history 'file-name-history)
                           (quit nil))))
          (if selected
              (insert "@" (substring-no-properties selected) " ")
            (insert "@")))))

    (keymap-set agent-shell-mode-map "@" #'backbone/agent-shell-fuzzy-insert-file)

    ;; Route / command CAPF completion through minibuffer for fuzzy matching.
    (defun backbone/agent-shell-setup-fuzzy-completion ()
      "Use consult-completion-in-region for CAPF in agent-shell."
      (setq-local completion-in-region-function
                  #'consult-completion-in-region))
    (add-hook 'agent-shell-mode-hook #'backbone/agent-shell-setup-fuzzy-completion)
```

## Checklist

### DIFF
--- a/agent-shell-completion.el
+++ b/agent-shell-completion.el
@@ -39,6 +39,15 @@
   :type 'boolean
   :group 'agent-shell)
 
+(defcustom agent-shell-file-completion-function
+  #'agent-shell--file-completion-at-point
+  "Function to call for @ file completion.
+Called with no arguments when @ is typed at a word boundary.
+If it returns non-nil, `completion-at-point' is not called.
+If it returns nil, falls through to the standard CAPF."
+  :type 'function
+  :group 'agent-shell)
+
 (defun agent-shell--completion-bounds (char-class trigger-char)
   "Find completion bounds for CHAR-CLASS, if TRIGGER-CHAR precedes them.
 Returns alist with :start and :end if TRIGGER-CHAR is found before
@@ -105,7 +114,8 @@ preventing spurious completions mid-word or in paths."
                  (memq (char-before (1- (point))) '(?\s ?\t ?\n))))
     (cond
      ((eq (char-before) ?@)
-      (completion-at-point))
+      (unless (funcall agent-shell-file-completion-function)
+        (completion-at-point)))
      ((and (eq (char-before) ?/)
            (agent-shell--command-completion-at-point))
       (completion-at-point)))))


xenodium:
Thanks for the PR. Sorry about the delay. Working my way through the backlog from #500.

How about we make this a little more versatile with something like?

```emacs-lisp
(defcustom agent-shell-file-completion-function
  #'completion-at-point
  "Function called when @ is typed at a word boundary.
The function takes no arguments and is responsible for performing
the completion (e.g. calling `completion-at-point' or invoking a
custom completion UI).  Set to nil to disable @ completion."
  :type '(choice (const :tag "Disabled" nil)
                 function)
  :group 'agent-shell)

(cond
 ((eq (char-before) ?@)
  (when agent-shell-file-completion-function
    (funcall agent-shell-file-completion-function)))
 ((and (eq (char-before) ?/)
       (agent-shell--command-completion-at-point))
  (completion-at-point)))
```

ps. Don't you need a similar thing for completing commands?

nohzafk:
Yes, I might need similar things for command completion. However, I suggest we add another custom variable for command completion that is symmetric to file completion.

xenodium:
> However, I suggest we add another custom variable for command completion that is symmetric to file completion.

SGTM!

PR #530, opened 2026-04-22
Add simple blinking circle as busy indicator

rudolf-adamkovic:
Use a simple (and elegant, IMO) busy indicator by default.

Needs to be experienced. 😄

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -373,7 +373,8 @@ Only appears when a session is active."
 Can be a symbol selecting a predefined style, or a list of frame strings.
 When providing custom frames, do not include leading spaces as padding
 is added automatically."
-  :type '(choice (const :tag "Wave (pulses up and down)" wave)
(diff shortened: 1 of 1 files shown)


xenodium:
Thank you for the PR.

> Use a simple (and elegant, IMO) busy indicator by default.

It's great you found a style you like. Preferences here vary significant across users. Some don't even want an indicator at all. I'm happy to merge the PR, but please don't change the default.

rudolf-adamkovic:
> Some don't even want an indicator at all. 

I agree.  In fact, the idea was to make the default indicator the least "aggressive" and/or opinionated.  Then, those who want more "circus" would be those who need to customize.  That said, I get your point about it being somewhat subjective.

> I'm happy to merge the PR, but please don't change the default.

Updated.  Now, the simple circle is not the default busy indicator anymore.

xenodium (approved):

xenodium:
Thanks for the changes. Merged.

PR #532, opened 2026-04-23
Fix removing queued messages

Gleek:
The fix replaces `when-let*` with `when` + `let*`. The problem was that `when-let*` short-circuits on any `nil` binding when removing the last item, `new-pending` is `nil` (empty list), so the `map-put!` never executed and the request stayed in the queue.

## Checklist


xenodium:
Thanks for the fix

PR #536, opened 2026-04-24
Add programmatic query API: agent-shell-query and agent-shell-shell-buffer

eddof13:
Hi @xenodium I was trying to get org-babel for AI prompts but I don't like the idea of configuring another client like https://github.com/karthink/gptel or https://github.com/rksm/org-ai so I have this https://github.com/eddof13/ob-agent-shell to get that support while reusing agent-shell- I had claude suggest some API hooks to clean up the code,  maybe this will be useful or you have some alternative suggestion?

## Summary

- `agent-shell-shell-buffer` — public wrapper around `agent-shell--shell-buffer`
- `agent-shell--extract-last-response` — extracts plain-text response from a buffer position
- `agent-shell-query` — programmatic async query function with cancel support

## Motivation

I built [ob-agent-shell](https://github.com/eddof13/ob-agent-shell), an org-babel backend that routes `#+begin_src agent-shell` blocks through a running agent-shell session. Rather than configure a second AI client for org, it reuses whatever agent-shell buffer you already have open.

To make this work I needed to:
1. Resolve a shell buffer from outside agent-shell (buffer resolution)
2. Send a query programmatically and capture the response

Both require reaching into private internals (`agent-shell--shell-buffer`, manual subscribe/unsubscribe/extract). That works today but is fragile across refactors.

These three additions provide a stable surface for that use case — and for any other package wanting to drive agent-shell programmatically.

## API

```elisp
;; Get the current shell buffer (same heuristic as internal code)
(agent-shell-shell-buffer)
(agent-shell-shell-buffer :no-error t)

;; Send a query and handle the response asynchronously
(agent-shell-query
 :text "What is 2+2?"
 :on-complete (lambda (response) (message "Got: %s" response))
 :on-error    (lambda (msg)      (message "Error: %s" msg)))

;; Cancel if needed
(let ((cancel (agent-shell-query :text "..." :on-complete ...)))
  (funcall cancel))
```

## Impact on ob-agent-shell

With these merged, ob-agent-shell's core [simplifies significantly](https://github.com/eddof13/ob-agent-shell/pull/1) — the subscribe/unsubscribe/extract/timeout logic all moves into `agent-shell-query` where it belongs.

## Notes

- `agent-shell-query` returns a cancel thunk so callers can abort in-flight requests
- Follows existing style: `cl-defun`/`&key`, `map.el`, `seq.el`, `when-let`
- `agent-shell--extract-last-response` is private (`--`) since it's an implementation detail of `agent-shell-query`; only `agent-shell-shell-buffer` and `agent-shell-query` are new public API
- Happy to split into two separate PRs if preferred

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -4930,6 +4930,98 @@ Returns a buffer object or nil."
                                          (current-buffer)))
                                 (agent-shell-buffers))))))
 
+;;;###autoload
+(cl-defun agent-shell-shell-buffer (&key viewport-buffer no-error no-create)
+  "Return an agent-shell buffer for the current context.
+
+A stable public API wrapping the internal resolver, intended for
+packages that integrate with agent-shell programmatically.
+
+Resolution order: viewport → current buffer → project buffers → prompt user.
+
+Example:
+  (agent-shell-shell-buffer)           ;; most recently used shell
+  (agent-shell-shell-buffer :no-error t) ;; nil instead of error when none found"
+  (agent-shell--shell-buffer :viewport-buffer viewport-buffer
+                              :no-error no-error
+                              :no-create no-create))
+
+(defun agent-shell--extract-last-response (buf start-pos)
+  "Extract plain-text agent response from BUF added after START-POS.
+
+Strips the echoed input and <shell-maker-end-of-prompt> separator,
+then strips the trailing shell prompt, returning only the response text.
+
+Example:
+  start-pos points to position before query was inserted
+  → returns \"Paris\" from a \"What is the capital of France?\" query"
+  (with-current-buffer buf
+    (let ((raw (buffer-substring-no-properties start-pos (point-max)))
+          (prompt-re (map-elt (agent-shell-get-config buf) :shell-prompt-regexp)))
+      (string-trim
+       (if (string-match (regexp-quote "<shell-maker-end-of-prompt>") raw)
+           (let ((after (substring raw (match-end 0))))
+             (if (and prompt-re (string-match (concat prompt-re ".*\\'") after))
+                 (substring after 0 (match-beginning 0))
+               after))
+         raw)))))
+
+(cl-defun agent-shell-query (&key text shell-buffer on-complete on-error timeout)
+  "Send TEXT to SHELL-BUFFER and call ON-COMPLETE with the response string.
+
+ON-COMPLETE is called with the plain-text response string.
+ON-ERROR is called with an error message string.
+TIMEOUT is seconds to wait before giving up (default: 60).
+SHELL-BUFFER defaults to the most recently used agent-shell buffer.
+
+Returns a cancel function — call it with no args to abort the in-flight
+query and clean up all subscriptions.
+
+Example:
+  (agent-shell-query
+   :text \"What is 2+2?\"
+   :on-complete (lambda (response) (message \"Got: %s\" response))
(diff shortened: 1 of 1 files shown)


eddof13:
One nice-to-have worth calling out: if `agent-shell-query` serializes requests internally (or tags each turn with an ID so `:on-complete` only fires for the matching turn), it would solve a subtle race condition in ob-agent-shell.

Currently, the package subscribes to `turn-complete` events which fire for *any* completed turn on the buffer. If the user is actively querying the same buffer while a babel block is executing, a concurrent turn completing first can cause the wrong response to be captured. The workaround today is to keep a dedicated idle buffer for org-babel use.

If `agent-shell-query` handles this internally — either by queueing or by correlating turns — callers get correct behavior for free and don't need to document the dedicated-buffer workaround. Not a blocker for the PR, just flagging it as a follow-on worth considering.

xenodium:
Thanks for the PR!

> * `agent-shell-shell-buffer` — public wrapper around `agent-shell--shell-buffer`

Happy to merge this. Could you please fix the indent?

> * `agent-shell--extract-last-response` — extracts plain-text response from a buffer position

I've reworked some existing logic to expose public APIs in https://github.com/xenodium/agent-shell/commit/9d9f22d2b24fb71ff2388ee8e0e8ad4dc0d8e572 which will enable you to achieve the same functionality using:

- agent-shell-goto-last-interaction
- agent-shell-interaction-at-point

> * `agent-shell-query` — programmatic async query function with cancel support

Sorry, but will push back on this one for the time being. Gut feeling is that we may run into some reliability edge cases in the current approach (some you rightfully spotted). Once I offer a public API, I'll need to support it and users will likely file issues.

I'm thinking there may be some issues with content integrity also? agent-shell buffers display a combination of collapsable UI along with markdown. How does this translate to display in RESULTS block?

For now, I'd prefer if `agent-shell-query` lives in `ob-agent-shell`. Sorry. Hope that's ok.

edit: At present, I also need to be super efficient with time dedicated to agent-shell due to #500 ;)

eddof13:
edited for humanization:

Thanks! And congrats! I've updated the PR to drop `agent-shell-query` and `agent-shell--extract-last-response`, and fixed the indentation on `agent-shell-shell-buffer`. ob-agent-shell now uses the new public API directly.

I have not dug into your question regarding the RESULTS display yet but this is what the machine said:

On content integrity: `agent-shell-interaction-at-point` returns the underlying plain text (`:response`), which is the raw markdown the agent emitted — not the rendered Emacs UI with overlays, collapsible sections, etc. So in a `#+RESULTS:` block you get the raw markdown string. That's actually fine for the ob-agent-shell use case: org-mode can render markdown-ish content reasonably well, and users who want to further process it can use `:results raw`. The collapsible UI rendering stays in the agent-shell buffer and doesn't leak into org results. Happy to add a note about this in the ob-agent-shell README if useful.

xenodium (approved):
Thanks for the changes

xenodium:
> Thanks! And congrats!

Thanks!

> yet but this is what the machine said

tell it to check agent-shell-ui--insert-fragment

Even then, it may be ok for your use case (including "▼ " "▶ ").

eddof13:
Thanks for the suggestion- seems unlikely that tool use will happen during a org-babel request but I updated the implementation to filter out those tool uses

PR #545, opened 2026-04-27
Keep buffer name on agent-shell-reload and -restart.

timfel:
I do this frequently to get out of situations where the ACP client is stuck or to clean out history, would be nice to keep the name then.

## Checklist


xenodium (approved):
Good call. Thank you.

PR #566, opened 2026-05-03
Preserve point and window-start across streaming rerenders

### LINKED ISSUE #454: When "Thinking" output is streaming, point constantly jumps up to top of section
## Checklist

  - agent-shell version: 82a7fff89094f509c22f89889401dce5134d459a
  - acp.el version: 863f2d62c4b4da8b229581be42d490a7403b2eb1
  - ACP package (e.g. claude-agent-acp) version: 0.22.2
  - Agent CLI (e.g. claude, gemini) version: 2.1.81
- [ ] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README). (I don't think it's relevant for this issue, but I can provide if needed)

If you attempt to follow the output as it streams in the "Thinking" section, your point will constantly jump back up to the top of the section (where the collapsable "Thinking" label is).

See this video for an example of what I mean (apologies for low quality but it gets the point across):

https://github.com/user-attachments/assets/7cff81cc-be7f-4f98-b44b-694f79556375

You can see I'm scrolling down but the point is jumping back up.

If I go to the end of the buffer (M->) it properly streams without jumping back up, but if I scroll up a little bit it will have the same issue.

It is kind of a pain to reproduce the issue, but usually if I just ask sonnet to think about something for a while because I am testing the thinking output it will work well (Huge waste of energy and tokens, I know 😬)

I had claude try solving the issue, but it did not end up really producing much meaningful information. (Following is from claude so take with huge grains of salt) It was saying that it seems like the issue may be caused by `shell-maker-with-auto-scroll-edit`. And also how if the point is inside the fragment (the thinking block) while it is being updated, since the fragment is deleted and recreated the point gets clamped to the start of the fragment (the thinking header).

Gleek:
Hello Alvaro,

This solves a bug where the point jumps to the top of the block that is currently rendering, if it's placed inside it.

#### Root cause
Each streaming chunk delete-and-reinserts the in-progress block, which displaces any markers inside that region to its start. `(point-marker)` in the viewport branch and `save-mark-and-excursion`/`save-excursion` (in `agent-shell-ui-update-fragment` and `shell-maker-with-auto-scroll-edit`) all rely on markers, so cursor and viewport jumped to the block header when reading mid-stream.

#### Fix
Switch to integer-based save/restore for point and window-start: inside `agent-shell-ui-update-fragment` (covers the viewport flow) and around `shell-maker-with-auto-scroll-edit` in the main shell branch (overrides shell-maker's broken `save-excursion`).

Fixes #454 
Also related (though different issue) is the #552

Please view the change with ?w=1 ([link](https://github.com/xenodium/agent-shell/pull/566/changes?w=1)). The change is not as big as the UI makes it to appear

## Checklist

### DIFF
--- a/agent-shell-ui.el
+++ b/agent-shell-ui.el
@@ -70,88 +70,100 @@ When EXPANDED is non-nil, body will be expanded by default.
 When NO-UNDO is non-nil, disable undo recording for this operation.
 
 For existing blocks, the current expansion state is preserved unless overridden."
-  (save-mark-and-excursion
-    (let* ((inhibit-read-only t)
-           (buffer-undo-list (if no-undo t buffer-undo-list))
-           (namespace-id (map-elt model :namespace-id))
-           (qualified-id (format "%s-%s" namespace-id (map-elt model :block-id)))
-           (new-label-left (map-elt model :label-left))
-           (new-label-right (map-elt model :label-right))
-           (new-body (map-elt model :body))
-           (block-start nil)
-           (padding-start nil)
-           (padding-end nil)
-           (match (save-mark-and-excursion
-                    (goto-char (point-max))
-                    (text-property-search-backward
-                     'agent-shell-ui-state nil
-                     (lambda (_ state)
-                       (equal (map-elt state :qualified-id) qualified-id))
-                     t))))
-      (when (or new-label-left new-label-right new-body)
-        (when match
-          (goto-char (prop-match-beginning match)))
-        (if (and match (not create-new))
-            ;; Found existing block - delete and regenerate
-            (let* ((existing-model (agent-shell-ui--read-fragment-at-point))
-                   (state (get-text-property (point) 'agent-shell-ui-state))
-                   (existing-body (map-elt existing-model :body))
-                   (block-end (prop-match-end match))
-                   (final-body (if new-body
-                                   (if (and append existing-body)
-                                       (concat existing-body new-body)
-                                     new-body)
-                                 existing-body))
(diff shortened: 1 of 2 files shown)


xenodium:
Thank you for the PR.

I've tried the patch briefly and ran into issues with point placed in unexpected places in viewport buffer when multiple permission dialogs came through. It could be unrelated. Will have to have another play with it later on.

ps. IIRC, the previous implementation remembered region too?

Gleek:
Thanks for trying it out.

Good point on region. I’d missed that `save-mark-and-excursion` also preserved mark state outside the rerendered block. I’ve amended the PR to preserve `mark` / `mark-active` using integer positions too.

One caveat: for a region inside the block being delete/reinserted, this is still positional rather than semantic. It restores the same offsets after rerender, not necessarily the same logical text if the replacement changes substantially. But it avoids the marker displacement to the block start, which was the original issue.


On the viewport permission-dialog issue: I couldn’t spot a direct cause in this patch. Permission dialogs still explicitly jump to the latest permission button after rendering, including in the viewport buffer. Happy to look more if you find a reproducible case.

xenodium:
Thanks for this!

> Each streaming chunk delete-and-reinserts the in-progress block

Since introducing the new markdown renderer, I've also changed fragment population so we append rather than reconstruct. Hopefully we no longer need this change. Closing, but please reopen if we still need work in the space.

PR #571, opened 2026-05-05
Fix unkillable buffer after major-mode change

Scott-Guest:
If a user manually changes the major mode of an agent-shell buffer (e.g. because they have fat fingers like I do :smile:), the `agent-shell--clean-up` entry on `kill-buffer-hook` survives, and its
```elisp
(unless (derived-mode-p 'agent-shell-mode)
  (error "Not in a shell"))
```
check makes the buffer unkillable.

To fix this, we
- Make `agent-shell--clean-up` no-op rather than error when the buffer isn't in agent-shell mode
- Explicitly register `agent-shell--clean-up` with `change-major-mode-hook` so we don't leak the process

## Checklist


xenodium (approved):
Thanks for the fix

Issue #584, opened 2026-05-17
Poor performance on long responses


AdamNiederer:
## Checklist

- [x] *I agree to communicate with the author myself* (not AI-generated).
- [x] I've read the README's [Filing issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#filing-issues) section.
- [x] I'm running the latest versions (fill in below).
  - agent-shell version: `20260515.1444`
  - acp.el version: `20260503.1851`
  - ACP package (e.g. claude-agent-acp) version: N/A
  - Agent CLI (e.g. claude, gemini) version: `opencode-1.5.0`
- [x] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README).

## Issue

Long reasoning traces can cause emacs to lock up with 100% cpu usage, seemingly mostly caused by markdown overlay reapplication on each token coming in. This happens even when the expander is collapsed and the user wouldn't see the overlays, so perhaps that should be either done at a much lower frequency or skipped entirely if the collapsible isn't expanded?

[agent-shell-profiler-cpu.txt](https://github.com/user-attachments/files/27864313/agent-shell-profiler-cpu.txt)
[agent-shell-profiler-mem.txt](https://github.com/user-attachments/files/27864314/agent-shell-profiler-mem.txt)
[agent-shell-traffic.txt](https://github.com/user-attachments/files/27864312/agent-shell-traffic.txt)

I repro'd this for the purposes of submitting this issue with a throwaway openai-compatible mocking script and a real reasoning trace, but it's from a real session and also occurred with the real provider.

[llm-mock.js](https://github.com/user-attachments/files/27864415/llm-mock.js)
[reasoning.md](https://github.com/user-attachments/files/27864416/reasoning.md)

xenodium:
Thanks for filing this issue. Left some comments in #585

xenodium:
I'd be interested to hear what kind of performance you get with this experimental markdown renderer in this branch/PR https://github.com/xenodium/agent-shell/pull/597

AdamNiederer:
The new renderer actually runs worse; I'm seeing redisplay get suppressed from 13% to just 5% of CPU time:

```
       33667  79% + timer-event-handler
        6006  14%   Automatic GC
        2492   5% + redisplay_internal (C function)
          54   0% + command-execute
          31   0% + #<byte-code-function 9E0>
[...]

PR #585, opened 2026-05-18
Skip setting overlays on collapsed fragments

AdamNiederer:
Partial resolution of #584

Before:
```
1,868,011,162  93% + timer-event-handler

         8541  48% + timer-event-handler
         6263  35%   Automatic GC
         2373  13% + redisplay_internal (C function)
```

After:
```
 979,703,965  88% + timer-event-handler

        7261  39% + redisplay_internal (C function)
        5444  29% + timer-event-handler
        5411  29%   Automatic GC
```

redisplay is no longer throttled and we save ~1GB of garbage on a 32k fragment length. CPU usage is still high but it's no longer causing major UI stutter. Expanded fragments still need help and there's probably some more optimization that could be done for both cases.

## Checklist

- N/A I've filed a feature request/discussion for a new feature.
- N/A I'm making visual changes, so I'm including screenshots so you can view and discuss.
- N/A I've added tests where applicable.

### DIFF
--- a/agent-shell-ui.el
+++ b/agent-shell-ui.el
@@ -529,7 +529,13 @@ When NO-UNDO is non-nil, disable undo recording."
                              (point) indicator-properties)
         (map-put! state :collapsed new-collapsed-state)
         (put-text-property (map-elt block :start)
-                           (map-elt block :end) 'agent-shell-ui-state state)))))
+                           (map-elt block :end) 'agent-shell-ui-state state)
+        ;; We skip markdown overlays on incoming msg when collapsed, so do it here.
+        (unless new-collapsed-state
+          (save-restriction
+            (narrow-to-region (map-elt body :start) (map-elt body :end))
+            (let ((markdown-overlays-highlight-blocks agent-shell-highlight-blocks))
+              (markdown-overlays-put))))))))
 
 (defun agent-shell-ui-collapse-fragment-by-id (namespace-id block-id)
   "Collapse fragment with NAMESPACE-ID and BLOCK-ID."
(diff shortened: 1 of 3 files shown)


agent-shell-ui.el
```diff
@@ -529,7 +529,13 @@ When NO-UNDO is non-nil, disable undo recording."
                              (point) indicator-properties)
         (map-put! state :collapsed new-collapsed-state)
         (put-text-property (map-elt block :start)
-                           (map-elt block :end) 'agent-shell-ui-state state)))))
+                           (map-elt block :end) 'agent-shell-ui-state state)
```
xenodium, line 539: agent-shell-ui.el is fairly generic in enabling buffer updates through collapsable fragements. With this change, we'd be introducing client-specific logic.

I'd prefer if we found a way to keep markdown-specific logic in agent-shell.el. Maybe through an agent-shell-ui-post-toggle-fragment-at-point-function (hook) that we funcall from here? Happy to consider alternatives.

xenodium:
Thanks a lot for the PR! Source block highlighting in markdown-overlays.el is highly inneficient. I'm working on a separate markdown rendering solution but that will take a little more time. I'll push it to a branch somewhat soon.

WRT to the PR itself, I'm happy to merge for the time being but could we please address the comment in review? Alternatively, isn't agent-shell-highlight-blocks set to nil enough to improve the immediate issue?

xenodium (approved):

xenodium:
Thank you for the follow-up changes. Could you resolve the conflicts so they can merged?

xenodium:
> WRT to the PR itself, I'm happy to merge for the time being but could we please address the comment in review? Alternatively, isn't agent-shell-highlight-blocks set to nil enough to improve the immediate issue?

I've rebased and made a few tweaks in https://github.com/xenodium/agent-shell/commit/d72f4b728d0d9bd4acee4539d484a395aa75873a. This is now merged.

Thanks for the PR

PR #589, opened 2026-05-19
Only trim response region when region is active

martenlienen:
Hey, sorry for the oversight. The response trimming in #513 fails when there is no response selected, because it applies `string-trim` to `nil`. This fixes that.


xenodium (approved):
Thanks!

PR #598, opened 2026-05-20
feat: support all ACP permission options including allow_session

yitang:
## Problem

Hermes Agent sends multiple distinct permission options (e.g. `allow_always`, `allow_session`) that may share the same `kind` value (`"allow_always"`). The current code deduplicates by kind, causing `allow_session` to be silently dropped from the UI. This means users get prompted for every single file in multi-file edits instead of being able to "Allow for session".

## Fix

- Remove the `acp-seen-kinds` parameter and deduplication logic from `agent-shell--make-permission-action`
- Add `"allow_session"` (`/`) keybinding to the char-map
- Allow all distinct ACP permission options to appear as separate buttons in the UI

## Changes

1. **`agent-shell--make-permission-actions`**: Removed `let (acp-seen-kinds)` tracking and passed-through no-arg calls to `agent-shell--make-permission-action`. Changed from `defun` to `cl-defun` for consistency.

2. **`agent-shell--make-permission-action`**: Removed `acp-seen-kinds` keyword argument. Added `("allow_session" . "/")` to the char-map. Simplified char lookup from `(unless (member kind acp-seen-kinds) (map-elt char-map kind))` to `(map-elt char-map kind)`.

3. Updated docstrings to document the no-deduplication behavior.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -5892,23 +5892,20 @@ CHAR and OPTION are used for cursor sensor messages."
                            button)))
     button))
 
-(defun agent-shell--make-permission-actions (acp-options)
+(cl-defun agent-shell--make-permission-actions (acp-options)
   "Make actions from ACP-OPTIONS for shell rendering.
 
-See `agent-shell--make-permission-action' for ACP-OPTION and return schema."
-  (let (acp-seen-kinds)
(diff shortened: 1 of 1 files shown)


yitang:
This PR was created by mistake, i am still testing this new feature locally.

PR #600, opened 2026-05-21
Do not emit permission-request for auto-handled permissions

Gleek:
When `agent-shell-permission-responder-function` handles a permission request, the prompt UI is skipped, but `agent-shell` still emitted `permission-request` and started the permission idle timer.

This made auto-approved permissions look pending to event subscribers.

This change treats a non-nil responder result as fully handled:
- skip rendering the permission prompt
- skip emitting `permission-request`
- skip starting the permission idle timer
- keep sending `permission-response` through the existing response path

The existing responder test now also verifies that auto-approved permissions do not emit a permission request event or leave an idle timer running.

Please view the change without whitespaces: https://github.com/xenodium/agent-shell/pull/600/changes?w=1

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1873,9 +1873,11 @@ COMMAND, when present, may be a shell command string or an argv vector."
                   (when-let ((diff (agent-shell--make-diff-info
                                     :acp-tool-call (map-nested-elt acp-request '(params toolCall)))))
                     (list (cons :diff diff)))))
-         (unless (and (functionp agent-shell-permission-responder-function)
+         (let* ((tool-call-id (map-nested-elt acp-request '(params toolCall toolCallId)))
+                (permission-handled
+                 (and (functionp agent-shell-permission-responder-function)
                       (funcall agent-shell-permission-responder-function
-                               (list (cons :tool-call (map-nested-elt state (list :tool-calls (map-nested-elt acp-request '(params toolCall toolCallId)))))
+                               (list (cons :tool-call (map-nested-elt state (list :tool-calls tool-call-id)))
                                      (cons :options (agent-shell--make-permission-actions
                                                      (map-nested-elt acp-request '(params options))))
                                      (cons :respond (lambda (option-id)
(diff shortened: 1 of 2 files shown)


xenodium:
Thanks for the improvement! I'm thinking that we'd still want to emit an event, but we can include additional information in the event data stating wether or not it was already handled programatically. This way, it's up to the client code to decide whether to ignore the event or potentially display something different. WDYT?

Gleek:
> I'm thinking that we'd still want to emit an event

How do you imagine this event being used?

To me, `agent-shell-permission-responder-function` is a convenient customization layer on top of what agents already provide through their session modes (`agent-shell-cycle-session-mode`). It lets users handle custom policies and provides a unified interface across agents.

It should behave consistently with agent/session-level permission policy: if a tool call is allowed by the active policy, whether that policy comes from the agent session or from `agent-shell-permission-responder-function`, it should not surface as a pending permission request.

Because of that, I don’t see much utility in emitting `permission-request` for already-handled permissions. For notification clients, `permission-request` means user action is needed. If the responder handled it, there is no pending user action, and emitting the same event forces every client to special-case handled permissions.

xenodium:
> How do you imagine this event being used?

You can think of it as reporting status. "We automatically accepted X", "We automatically rejected Y", events. If we keep them, but mark them handled, I'm thinking it'll enable your use-case, but also anyone else that want a different notification policy. In Emacs, it's only a matter of time until someone asks for a different way of handling it, thus I try to favor some generic integration points. Hope that makes sense?

Gleek:
That makes sense as a general integration goal, and I’m open to events that report tool/permission status.

The part I’m unsure about is emitting this through `permission-request`. If a tool call is allowed automatically, either by an agent/session mode or by `agent-shell-permission-responder-function`, I don’t think it should look like a pending permission request. Those two policy mechanisms should feel consistent to clients.

I can see two clearer event shapes:

1. `permission-request`: emitted only when user permission is actually required.
2. A separate tool/permission status event: emitted for broader reporting, including “auto-accepted”, “auto-rejected”, and normal tool calls that never required permission.

That would cover notification clients and audit/status clients without making every `permission-request` consumer check whether the request is actually pending.

The case I’m struggling to justify is an event only for responder-handled permissions, but not for permissions/tool calls handled by the agent’s own session policy. If the goal is reporting “what happened,” a separate status/tool event seems more complete and less surprising.

For custom user policies, the responder function itself is already the most flexible place to add logging/notifications/advice, because it directly knows why a request was accepted or rejected.

That said, I understand the preference for generic integration points. If you still prefer keeping the event and adding handled/status metadata, I’m okay adapting the PR in that direction.

xenodium:
> A separate tool/permission status event: emitted for broader reporting, including “auto-accepted”, “auto-rejected”, and normal tool calls that never required permission.

Good idea. This is much better. Do you fancy doing this too? No presure if not. Otherwise, I'll review PR as is.

Gleek:
Sure! I'll give it a shot in sometime. Do you think we can merge this now and raise a separate PR for the general tool event? Or hold this until that is ready and merge those together?

Gleek:
I see that we already have a `tool-call-update` event.

From what I understand, `tool-call-update` already reports the tool lifecycle from ACP `tool_call` / `tool_call_update` notifications, with data like `:tool-call-id` and `:tool-call` containing status, kind, title, command, raw input, content, etc.

Maybe the more general integration point is to enrich that event with permission/policy outcome data when available. For example, when a responder auto-handles a permission, `tool-call-update` could include something like:

```elisp
(:permission . ((:status . auto-approved)
                (:option-id . "...")
                (:handled-by . responder)))
```

That way:

- permission-request stays focused on "user action is required".
- Notification clients don’t need to special-case already-handled permissions.
- Audit/status clients can still observe that a tool call was auto-approved or auto-rejected.
- The same event family can cover normal tool lifecycle updates and permission policy outcomes.

This feels like a cleaner generic integration point to me. WDYT?

xenodium (approved):

xenodium:
> Do you think we can merge this now and raise a separate PR for the general tool event?

Yep. Merging now. Thank you.

xenodium:
> ```emacs-lisp
> (:permission . ((:status . auto-approved)
>                 (:option-id . "...")
>                 (:handled-by . responder)))
> ```

Thanks for this. LGTM, though I'm wondering what responder may be? The function itself? If so, how could clients use it?

PR #605, opened 2026-05-26
Add `agent-shell-restore-context` defcustom

nhojb:
Currently when restoring a session you can choose a minimal resume or a full load via `agent-shell-prefer-session-resume`.

This PR adds a `summary` option via a new agent-shell-restore-context `defcustom`. The options are:

  - `minimal`: Equivalent to `agent-shell-prefer-session-resume: t`
  - `full`: Equivalent to `agent-shell-prefer-session-resume: nil`
  - `summary`: Loads the full session (when available), but renders only the first prompt and last text response.

The idea is that summary gives you enough context about the session, without needing to restore the entire conversation and tool calls.

Here's an example:

8oar1X67TfB

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -637,26 +637,26 @@ When non-nil (and supported by agent), prefer ACP session resumes over loading."
   :group 'agent-shell)
 
 (make-obsolete-variable 'agent-shell-prefer-session-resume
-                        'agent-shell-restore-context
+                        'agent-shell-session-restore-strategy
                         "agent-shell 0.52")
 
-(defcustom agent-shell-restore-context 'minimal
+(defcustom agent-shell-session-restore-strategy 'minimal
   "How much prior context to show when restoring a session.
 
   `minimal': Show only the session title (default).  Uses
              `session/resume' when supported (no message replay),
              so restore is fast and quiet.
-  `summary': Use `session/load' and, when the replay completes,
-             render the initial user prompt and the last agent
+  `first-last': Use `session/load', when the replay completes,
+             render the first user prompt and the last agent
              text response.  Other notifications (tool calls,
              thoughts) are suppressed during restore.
   `full':    Use `session/load' and replay the entire conversation.
 
-`summary' and `full' both require the agent to advertise
+`first-last' and `full' both require the agent to advertise
 `session/load' support.  When unavailable, restore falls back
 to `minimal' behavior."
   :type '(choice (const :tag "Title only (minimal)" minimal)
-                 (const :tag "First prompt + last response (summary)" summary)
+                 (const :tag "First prompt + last response (first-last)" first-last)
                  (const :tag "Full replay" full))
   :group 'agent-shell)
 
@@ -4823,16 +4823,16 @@ Falls back to latest session in batch mode (e.g. tests)."
 (defun agent-shell--use-session-load-p (state)
   "Return non-nil when STATE should restore via `session/load'.
 
-`agent-shell-restore-context' decides the protocol:
+`agent-shell-session-restore-strategy' decides the protocol:
 
-  `summary' and `full' force `session/load' when the agent
+  `first-last' and `full' force `session/load' when the agent
   advertises it (so a replay is available to read from); they
   fall back to `session/resume' otherwise.
 
   `minimal' uses `session/resume' when available, falling back
   to `session/load' only if the agent doesn't support resume."
   (cond
-   ((and (memq agent-shell-restore-context '(summary full))
+   ((and (memq agent-shell-session-restore-strategy '(first-last full))
          (map-elt state :supports-session-load))
     t)
    ((map-elt state :supports-session-resume)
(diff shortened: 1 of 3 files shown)


agent-shell.el
```diff
@@ -573,6 +573,30 @@ When non-nil (and supported by agent), prefer ACP session resumes over loading."
   :type 'boolean
   :group 'agent-shell)
 
+(make-obsolete-variable 'agent-shell-prefer-session-resume
+                        'agent-shell-restore-context
+                        "agent-shell 0.52")
+
+(defcustom agent-shell-restore-context 'minimal
```
xenodium: How about naming this agent-shell-session-restore-strategy, it's its more about restoring sessions than context-centric?

agent-shell.el
```diff
@@ -573,6 +573,30 @@ When non-nil (and supported by agent), prefer ACP session resumes over loading."
...
+  "How much prior context to show when restoring a session.
+
+  `minimal': Show only the session title (default).  Uses
+             `session/resume' when supported (no message replay),
+             so restore is fast and quiet.
+  `summary': Use `session/load' and, when the replay completes,
```
xenodium: Can we reserve summary for a different feature please?

I've been contemplating pairing resume request with an automaric prompt sent to agent asking to summarize the resumed session.

Maybe name it `'first-last`?

xenodium (changes requested):
Thanks for the PR!

Left a couple of minor comments.

The more significant request I'd ask is that you commit to landing the feature, if that's ok with you? If there are related bugs raised, I'd cc ya for visibility.

agent-shell.el
```diff
@@ -4661,6 +4691,118 @@ Falls back to latest session in batch mode (e.g. tests)."
...
+a `Last response' fragment for the latest agent text reply, then
+clears the summary state.  Does nothing if neither was captured."
+  (when-let ((summary (map-elt state :restore-summary)))
+    (agent-shell--restore-summary-commit-in-flight summary)
+    (when-let ((text (map-elt summary :first-user)))
+      (agent-shell--update-fragment
+       :state state
+       :namespace-id "bootstrapping"
+       :block-id "restore_summary_first_user"
+       :label-left (propertize "Initial prompt" 'font-lock-face 'font-lock-doc-markup-face)
```
xenodium: nit: search "Initial prompt", replace with "First prompt".

nhojb:
Thanks for the review. Just pushed the suggested fixes.

> The more significant request I'd ask is that you commit to landing the feature, if that's ok with you? If there are related bugs raised, I'd cc ya for visibility.

Yes, I'm quite happy to tackle any related bugs and own the feature 😅

xenodium:
Thanks for the PR! This is now merged.

Heads-up, I may need to rethink/consolidate settings. We currently have agent-shell-session-strategy and agent-shell-session-restore-strategy. While they toggle different things, it may be time to shed some functionality. For example, `new-deferred` is the oldest flow before we were able to bootstrap shells before offering a prompt.

Initializing/bootstrapping has been evolving but also accumulating complexity. We may not need to support every configuration here. Need to think more about it.

nhojb:
> Heads-up, I may need to rethink/consolidate settings. We currently have agent-shell-session-strategy and agent-shell-session-restore-strategy. While they toggle different things, it may be time to shed some functionality. 

Yes, makes sense to me. It's quite confusing adding or updating defcustoms right now because of the interaction between the different options. Of course it's _nice_ to have options in emacs packages, but better to keep them sane.

> For example, `new-deferred` is the oldest flow before we were able to bootstrap shells before offering a prompt.

I'd never use this one now, always preferring to have my session history available. So I think it could safely be removed since you have to pay the startup cost to use the agent anyway, even if the user doesn't want session history.

xenodium:
Heads-up. For now I've renamed [agent-shell-session-restore-strategy to agent-shell-session-restore-verbosity and retired 'new-deferred usage](https://github.com/xenodium/agent-shell/commit/e58909d79ae8b57ca392560a9e66e653902fe60e).

Where I'm thinking about going longer term is to remove `agent-shell-session-restore-verbosity` altogether and keep `agent-shell-session-strategy` with possible values of: `always-new`, `restore-latest`, `restore-minimally`, `restore-first-last`, `restore-fully`. These don't necessarily map to the internals, but that's probably fine (internal detail anyway).

xenodium:
Heads-up https://github.com/xenodium/agent-shell/commit/183b3077814c83372084bbf70f41cfc86c6c2c61 renders first-last as regular shell interactions now. Forgot to mention in commit, `'last` is now an option too.

xenodium:
Also https://github.com/xenodium/agent-shell/commit/dffbcfc297f0943590e84ff3664b167310b60fd6 alerting users of degradation

```
╭────────────────────────────────────────────────────────╮
│ Warning: minimal unsupported. Using first-last loading │
╰────────────────────────────────────────────────────────╯

╭──────────────────────────────────────────────────╮
│ Warning: last unsupported. Using minimal loading │
╰──────────────────────────────────────────────────╯

╭────────────────────────────────────────────────────────╮
│ Warning: first-last unsupported. Using minimal loading │
╰────────────────────────────────────────────────────────╯

╭──────────────────────────────────────────────────╮
│ Warning: full unsupported. Using minimal loading │
╰──────────────────────────────────────────────────╯

╭────────────────────────────────────────────────────────╮
│ Warning: Could not load existing session. Starting new │
╰────────────────────────────────────────────────────────╯

╭─────────────────────────────────────────────────────────╮
│ Warning: Could not list existing sessions. Starting new │
╰─────────────────────────────────────────────────────────╯

```

PR #625, opened 2026-06-04
Fix source-block face background not rendering when theme loads after…

phairoh:
The previous fix sampled org-block's background at load time using face-attribute. If the package loads before the user's theme is applied (common with lazy-loading setups), face-attribute returns unspecified and no background is rendered.

Switching to :inherit org-block resolves this by deferring face attribute lookup to render time, so the face always reflects the active theme. :foreground unspecified is added explicitly to prevent org-block's foreground (if set by the theme) from overriding font-lock syntax highlighting colors inside the block.

With the fix:
screenshot-20260604-173045

Without the fix:
screenshot-20260604-174903

The difference is subtle, but it's there.

## Checklist

### DIFF
--- a/agent-shell-markdown.el
+++ b/agent-shell-markdown.el
@@ -144,13 +144,11 @@
   :group 'agent-shell-markdown)
 
 (defface agent-shell-markdown-source-block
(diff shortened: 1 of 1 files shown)


xenodium (approved):
Thank you!

Issue #630, opened 2026-06-06
add support for oh-my-pi (omp) agent


tychoish:
I have an implementation that I'll file in parallel. Been using it for a bit and it works pretty well. No complaints.

## Checklist

- [x] *I agree to communicate with the author myself* (not AI-generated).
- [x] I've read the README's [Filing issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#filing-issues) section.
- [x] I'm running the latest versions (fill in below).
  - agent-shell version: (melpa; `20260509.1712260509.1712`)
  - acp.el version: (melpa; `20260503.1851`)
  - ACP package (e.g. claude-agent-acp) version: 15.9.0
  - Agent CLI (e.g. claude, gemini) version: as above
- [x] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [-] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README).

PR #631, opened 2026-06-06
feat: add implementation/support for oh-my-pi (omp)

tychoish:
As mentioned in #630, support for omp agent. Been using it for a bit, and it seems to work. 🤷.

Didn't want to keep it to myself, and I hope it's useful to other people.

(agent shell is great!)

## Checklist

- [y] I've filed a feature request/discussion for a new feature. (#630)
- [-] I'm making visual changes, so I'm including screenshots so you can view and discuss.


agent-shell-omp.el
```diff
@@ -0,0 +1,134 @@
...
+  (agent-shell-make-agent-config
+   :identifier 'omp
+   :mode-line-name "OMP"
+   :buffer-name "OMP"
+   :shell-prompt "OMP> "
+   :shell-prompt-regexp "OMP> "
+   :welcome-function #'agent-shell-omp--welcome-message
+   :client-maker (lambda (buffer)
+                   (agent-shell-omp-make-client :buffer buffer))
+   :install-instructions "Install omp (Oh My Pi) and ensure it is available on your PATH."))
```
xenodium: We should follow the same patterns as others and include a URL

Something like (if that's the right location):

"See https://github.com/can1357/oh-my-pi for installation."

agent-shell-omp.el
```diff
@@ -0,0 +1,134 @@
...
+  "Environment variables for the Oh My Pi (omp) client.
+
+This should be a list of environment variables to be used when
+starting the omp client process.
+
+Example usage:
+
+  (setq agent-shell-omp-environment
+        (agent-shell-make-environment-variables
+         \"ANTHROPIC_API_KEY\" \"your-key\"))"
```
xenodium: Is ANTHROPIC_API_KEY relevant for this agent? If not, we should try to show a relevant variable if possible

xenodium (changes requested):
Thanks for the PR!

Left a few minor comments. Mind taking a look? Also, we should update the README.org file with similar information as the others.

agent-shell-omp.el
tychoish: It definitely can be, depending on what backend or backends you use with the agent.

To be less confusing I'll change this to just <KEY> and <VALUE>

agent-shell-omp.el
tychoish: done!

xenodium:
Thanks for the changes. Merging.

PR #632, opened 2026-06-07
Render tool call parameters for non-standard tools like MCP calls

martenlienen:
agent-shell already has the information and this change makes it visible to the user. Showing the parameters is possible for any tool call, but built-ins like read or the bash tool have their parameters baked into their titles by the agent (at least Claude Code), so only display parameters for non-standard tools.

The following screenshot shows the two new cases (single and multiple parameters). Standard tool calls like read and bash look unchanged.

grafik

## Checklist


xenodium:
Thanks for the PR. Merged.

PR #658, opened 2026-06-16
Fix viewport previous-page navigation and stale position label

### LINKED ISSUE #657: Switching to the viewport: first previous-page does nothing and position label can be stale
## Checklist

  - agent-shell version: 0.55.1 (main @ 3ffc362)
  - acp.el version: n/a (UI navigation bug, reproducible without a running agent)
  - ACP package (e.g. claude-agent-acp) version: n/a
  - Agent CLI (e.g. claude, gemini) version: n/a
- [ ] For issues, I'm including ACP traffic (n/a — no agent interaction is needed to reproduce).

## Steps to reproduce

1. In an `agent-shell` buffer with at least two interactions, put point in
   the middle of an interaction (not on its prompt line).
2. `M-x agent-shell-other-buffer` to switch to the viewport (view mode).
3. `M-x agent-shell-viewport-previous-page`.

## What happens

- The first `agent-shell-viewport-previous-page` appears to do nothing (it
  re-shows the current interaction); a second invocation is needed to move to
  the previous interaction.
- The position label in the header sometimes shows a count that doesn't match
  the displayed interaction (e.g. `1/2` when the latest interaction is shown).

## Cause

- `agent-shell-viewport-next-page` navigates from the raw shell-buffer point.
  `agent-shell-other-buffer` leaves that point wherever it was — mid-interaction
  — so the first backward step only reaches the current interaction's prompt
  rather than the previous interaction. (`view-last` doesn't hit this because it
  repositions the point to the last prompt first.)
- `agent-shell-viewport--initialize` refreshes the cached position but never
  re-renders the header, so `agent-shell-viewport-refresh` (used on switch)
  leaves the previously-rendered position label in place.

Fix and tests ready (PR to follow).

liaowang11:
Fixes #657.

Two issues when switching to the viewport with `agent-shell-other-buffer` and
then navigating:

- **First `previous-page` does nothing.** `agent-shell-viewport-next-page`
  navigated from the raw shell-buffer point, which `agent-shell-other-buffer`
  leaves mid-interaction. From there the first backward step only reaches the
  current interaction's prompt, so the page doesn't change until a second
  invocation. Fixed by normalizing the point to the current interaction's
  prompt start (`shell-maker--prompt-begin-position`) before stepping back.
  Scoped to the backward branch so forward navigation is unchanged.

- **Stale position label (e.g. `1/2`).** `agent-shell-viewport--initialize`
  refreshed the cached position but never re-rendered the header, so
  `agent-shell-viewport-refresh` (used on switch) left the previously-rendered
  label in place. Fixed by re-rendering the header at the end of
  `--initialize`; the now-redundant explicit call in `next-page` is dropped.

Adds regression tests for both.

## Checklist

- [x] I've filed an issue for this bug (#657).
- [ ] I'm making visual changes, so I'm including screenshots so you can view and discuss. (n/a — no visual change)
- [ ] I've updated documentation where necessary. (n/a)

### DIFF
--- a/agent-shell-viewport.el
+++ b/agent-shell-viewport.el
@@ -311,7 +311,12 @@ Optionally set its PROMPT and RESPONSE."
     (when response
       (insert response))
     (let ((inhibit-read-only t))
-      (markdown-overlays-put))))
+      (markdown-overlays-put))
+    ;; Re-render the header from the freshly-cached position.  Without
+    ;; this, callers that set content via this function (e.g.
+    ;; `agent-shell-viewport-refresh' on switch) leave the header showing
+    ;; the previously-rendered position.
+    (agent-shell-viewport--update-header)))
 
 (defun agent-shell-viewport--ensure-buffer ()
   "Ensure current buffer is a viewport and err otherwise."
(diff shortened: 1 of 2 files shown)


xenodium (approved):

xenodium:
Thanks for the improvement! Very familiar with this quirk. Glad you fixed it.

PR #659, opened 2026-06-17
CI: Use native ERT JUnit Reporting

phairoh:
I originally couldn't get native JUnit reporting to produce a file, so I fell back to `ert-junit`, which is unmaintained and crashes when a test errors, so `test-results.xml` was never generated on those runs. As it turns out native JUnit was fine all along, I just wasn't setting the output name via `ert-load-file-name` (the `EMACS_TEST_JUNIT_REPORT` env var enables it but doesn't control the filename). This PR drops `ert-junit` and uses the built-in reporter.

Also, in `test-report.yml`, `dorny/test-reporter@v3` defaults `use-actions-summary: true`, which doesn't attach the test report to the commit so you have to go digging for it. Setting it to false fixes this. **NOTE**: this will only take effect once it's on `main`.

What Passing tests will look like when you click on the checkbox next to the commit:
screenshot-20260617-152239

Failing tests:
screenshot-20260617-153349

## Checklist

### DIFF
--- a/.github/workflows/test-report.yml
+++ b/.github/workflows/test-report.yml
@@ -32,3 +32,4 @@ jobs:
           name: ERT Tests
           path: test-results.xml
           reporter: java-junit
+          use-actions-summary: false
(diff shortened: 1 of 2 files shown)


xenodium (approved):
Lovely! Thank you for looking into it.

PR #661, opened 2026-06-18
Don't delete temp dir on agent-shell-restart.

arthurgleckler:
This is a bug fix.  In shells created using agent-shell-new-temp-shell, running agent-shell-restart would trigger the hook that deleted the temporary directory, leaving no temporary directory.  This prevents that.

An alternative would be to create a new temporary directory under those circumstances.

## Checklist

- [x ] *I've reviewed all code in PR myself and will vouch for its quality*.
- [N/A ] I've filed a feature request/discussion for a new feature.
- [N/A] I'm making visual changes, so I'm including screenshots so you can view and discuss.
- [N/A] I've added tests where applicable.
- [N/A] I've updated documentation where necessary.
- [N/A] I've run `M-x checkdoc` and `M-x byte-compile-file`.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -887,6 +887,14 @@ OUTGOING-REQUEST-DECORATOR (passed through to `acp-make-client')."
 (defvar-local agent-shell--transcript-file nil
   "Path to the shell's transcript file.")
 
+(defun agent-shell--cleanup-default-directory ()
+  "Delete the current buffer's `default-directory'.
+Only attach this to shells created by `agent-shell-new-temp-shell',
+whose `default-directory' is a disposable temp dir.  Attaching it to
+an ordinary shell would delete its working directory."
+  (when (file-directory-p default-directory)
+    (delete-directory default-directory t t)))
+
 (defvar agent-shell--shell-maker-config nil)
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


xenodium:
Thanks for this!

I'm thinking we can do without the additional global var? The cleanup happens on the default directory and is triggered only when a cleanup hook is added, thus we can always just add/remove the hook when needed without keeping track of a separate variable?

So something like the following (I have not tested this):

```emacs-lisp
((defun agent-shell--cleanup-default-directory ()
  "Delete the current buffer's `default-directory'.
Only attach this to shells created by `agent-shell-new-temp-shell',
whose `default-directory' is a disposable temp dir.  Attaching it to
an ordinary shell would delete its working directory."
  (when (file-directory-p default-directory)
    ;; Third arg is TRASH: recoverable, matching the temp-shell contract.
    (delete-directory default-directory t t)))
```

arthurgleckler:
That's much cleaner.  Done.  (There is one failing test, `agent-shell--emit-event-isolates-throwing-subscriber-test`, but that seems unrelated.)

xenodium (approved):

xenodium:
Thanks for the fix and follow-up changes

xenodium:
FYI tiny change https://github.com/xenodium/agent-shell/commit/4c8db2c374fe90cf469ce239d1a4334db7ec940b

PR #664, opened 2026-06-21
Fix transcript section heading glued to interrupted agent message

liaowang11:
An interrupted `agent_message_chunk` leaves the transcript body without
a trailing newline, so the next prompt's heading lands on the same line
as the partial message, e.g.

    Actually, I should## User (2026-06-20 19:45:42)

Append a blank-line separator at turn end whenever the last streamed
entry was an `agent_message_chunk`. Centralize this in a helper called on
both the success and failure (interrupt) paths of the session/prompt
request.

## Checklist


xenodium (approved):
Thank you for the improvement!

PR #666, opened 2026-06-22
Preserve viewport edit draft when queued request executes

liaowang11:
## What

When a command is sent (e.g. a queued request executing) while the user is composing in the viewport edit buffer, the draft is now saved to the compose snapshot instead of being wiped by the switch to view mode. The draft is restored on the next return to edit mode.

## Why

`agent-shell--send-command` switches the viewport to view mode and re-initializes the buffer. If a queued request runs while the user is mid-composition in the viewport edit buffer, their in-progress draft was lost. Reusing the existing `agent-shell-viewport--compose-snapshot` mechanism preserves it.

## Checklist


xenodium (approved):
Thanks for the improvement!

PR #674, opened 2026-06-25
Display math rendering (LaTeX → SVG)

alberti42:
This is not a PR in the strict sense -- it's a demo submission of a new feature. I'll call it a PR for brevity.

The code is there to show how the feature can be done, and to share design decisions with a working implementation rather than having a generic abstract discussion. This gives us the opportunity to discuss the design concretely and iterate fast -- trying out how certain choices implement in practice. Once we converge on the desired concept, I'm happy if you take the code, but equally happy if you decide to throw it away and recode from scratch once we've identified how we want the whole thing to work.

## What it does

Have you ever tried to have a conversation about scientific subjects with an agent? It always returns mathematical expressions written in raw LaTeX, which is hard to read. In my academic field, I daily have to read walls of unformatted LaTeX, which makes the experience unnecessarily difficult. The code demoed here renders all equations live as they are streamed by the agent. The community of scientists using Emacs may not be as big as that of coders, but for those in academia or data science, this kind of feature makes a real difference.

Here's a short demo (~1 min, no audio): https://www.youtube.com/watch?v=wGM3xH06Wso

- Block-level `\[…\]`, `$$…$$`, and `` ```math ``/`` ```latex `` fences render as SVG images; inline `\(…\)` too
- SVGs render with perfect resolution — no pixelated effect
- Equation size is matched to the buffer font size
- Foreground and background colors sync automatically with the theme; no need to regenerate the SVGs
- When changing font size, all equations rescale correctly; no need to regenerate the SVGs
- Rendering is async and cached on disk, so it doesn't block streaming and repeats are instant
- No special instructions needed for the agent — the only thing worth telling it is to avoid rendering inline equations as markdown inline code, because there's no unambiguous way to distinguish an equation from actual code

The code lives in a dedicated `agent-shell-markdown-math.el`, with minimal changes to the rest of the codebase — I tried to localize everything into one file to keep the footprint small and make it easier to maintain going forward.

I can also share the `CLAUDE.md` I wrote to document the module architecture — it may help navigate the code faster than a PR description would.

## Screenshot

Here is a screenshot of the new feature. If you click on the YouTube link above, you will get a more precise impression.

Screenshot 2026-06-25 at 15 57 38

### DIFF
--- a/agent-shell-markdown-math.el
+++ b/agent-shell-markdown-math.el
@@ -135,6 +135,21 @@ Also used as the automatic fallback when the LaTeX toolchain
 \(`agent-shell-markdown-math-latex-program' /
 `agent-shell-markdown-math-dvisvgm-program') is unavailable.")
 
+(defvar agent-shell-markdown-math-render-on-non-graphic nil
+  "When non-nil, render equation images even on a non-graphical frame.
+
+By default equations are only compiled when the selected frame is
+graphical (`display-graphic-p').  In an Emacs daemon a buffer may
+be rendered while a TTY frame is selected, yet later viewed in a
+graphical frame; without this the equation would never have been
+produced and stays raw text in the GUI too.
+
+Set non-nil (typically in a daemon setup) to always compile the
+SVG when the build supports it: it is ignored on a TTY frame (the
+raw LaTeX shows) but appears as soon as a graphical frame views
+the buffer.  The trade-off is that a purely terminal session then
+spawns LaTeX compiles whose images it never displays.")
+
 (defvar agent-shell-markdown-math-latex-program "latex"
   "Program that compiles a LaTeX document to DVI.")
 
@@ -399,18 +414,31 @@ For example:
         (apply #'color-rgb-to-hex (append rgb '(2)))
       fallback)))
 
+(defun agent-shell-markdown--math-renderable-p ()
+  "Return non-nil when equation images should be produced.
+
+Requires SVG image support in this Emacs build, plus either a
+graphical selected frame or
+`agent-shell-markdown-math-render-on-non-graphic' (the daemon /
+mixed TTY+GUI case — the image is ignored on a TTY frame but shows
+once a graphical frame views the buffer)."
+  (and (image-type-available-p 'svg)
+       (or (display-graphic-p)
+           agent-shell-markdown-math-render-on-non-graphic)))
+
 (defun agent-shell-markdown--latex-placeholder-image (latex)
   "Return a placeholder SVG image boxing the raw LATEX, or nil.
 
 This does NOT typeset LATEX — it draws the source inside a
 bordered panel.  Used when `agent-shell-markdown-math-use-placeholder'
 is set or the LaTeX toolchain is unavailable, so math still has a
-visible (if un-typeset) rendering.  Returns nil when no graphical
-display is available, so callers fall back to the raw text.
+visible (if un-typeset) rendering.  Returns nil when equations
+aren't renderable (see `agent-shell-markdown--math-renderable-p'),
+so callers fall back to the raw text.
 
 LATEX is the equation source with the surrounding delimiters
 already stripped, e.g. \"E=mc^2\"."
-  (when (display-graphic-p)
+  (when (agent-shell-markdown--math-renderable-p)
     (let* ((lines (split-string latex "\n"))
            ;; `frame-char-width' / `-height' give per-char pixel
            ;; dimensions on a graphical frame and stay robust off it
@@ -520,14 +548,15 @@ is preserved."
 (defun agent-shell-markdown--math-render (buffer start end latex)
   "Render LATEX over BUFFER's START..END as an equation image.
 
-On a non-graphical display, does nothing (the raw faced text
-stands in).  With `agent-shell-markdown-math-use-placeholder' set
+Does nothing when equations aren't renderable (see
+`agent-shell-markdown--math-renderable-p') — the raw faced text
+stands in.  With `agent-shell-markdown-math-use-placeholder' set
 or no LaTeX toolchain, overlays the placeholder panel.  Otherwise
 overlays the cached SVG immediately when available, else schedules
 an async compile (see `agent-shell-markdown--math-compile') that
 overlays the result once ready — START / END are captured as
 markers so the overlay lands even after more output streams in."
-  (when (display-graphic-p)
+  (when (agent-shell-markdown--math-renderable-p)
     (cond
      ((or agent-shell-markdown-math-use-placeholder
           (not (agent-shell-markdown--math-tools-available-p)))
--- a/agent-shell-markdown-math.el
+++ b/agent-shell-markdown-math.el
@@ -188,8 +188,30 @@ spawns LaTeX compiles whose images it never displays.")
   "Program that converts DVI to SVG.")
 
 (defvar agent-shell-markdown-math-scale 1.4
-  "`dvisvgm' output scale for rendered equations.
-Larger values produce bigger images.")
+  "`dvisvgm' output scale for compiled equation SVGs.
+
+This is a render-quality / vector-precision knob, not a size knob:
+the displayed equation is rescaled to the buffer font at display
+time (see `agent-shell-markdown--math-display-scale'), and this
+factor cancels out of that computation.  To change how big
+equations appear on screen, use `agent-shell-markdown-math-font-scale'.")
+
+(defvar agent-shell-markdown-math-font-scale 1.0
+  "Size of rendered equations relative to the buffer font.
+
+Equation images are scaled so LaTeX's 10pt body font maps onto the
+buffer's font height; this multiplier rides on top of that match.
+1.0 makes equation text the same size as the surrounding text;
+greater than 1 enlarges, less than 1 shrinks.  Because the match is
+recomputed from the current font, equations track the buffer font
+across themes and faces (run `agent-shell-markdown-math-refresh'
+after a pure font-size change — see its docstring).")
+
+(defvar agent-shell-markdown-math--svg-px-per-pt nil
+  "Cached pixels-per-point Emacs uses to render SVG images.
+Measured once on a graphical frame by
+`agent-shell-markdown--math-svg-px-per-pt' (so HiDPI / image
+scaling is captured exactly); nil until then.")
 
 (defvar agent-shell-markdown-math-preamble
   "\\documentclass[border=2pt]{standalone}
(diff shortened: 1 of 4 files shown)


xenodium:
Thanks for this! Very cool.

> Once we converge on the desired concept, I'm happy if you take the code, but equally happy if you decide to throw it away and recode from scratch once we've identified how we want the whole thing to work.

Being a substantial subpackage, my preference would be for this to live as a separate third-party package. I'd be happy to link to it like others in https://github.com/xenodium/agent-shell#related-projects

Having said this, you likely need an entry point to agent-shell-markdown, which we can add as customisable behvior (maybe a list of functions to apply). Would this work for ya?

alberti42:
Thanks for the quick and positive reply. I'm happy to maintain it as a separate package, so you don't carry the burden of maintaing something you don't use yourself. Being listed under `related-projects` would be great. And if the extension proves stable, merging is always an option we consider down the line.

On the entry point, a list of hook functions would cover part of it, but the integration is a bit deeper than that. I squashed everything into a single commit (on branch [display-math-squashed](https://github.com/alberti42/fork-agent-shell/tree/display-math-squashed)) so you can see what we touched in agent-shell-markdown.el. The
changes are ~120 lines across 7 points:

1. Avoid-ranges: math blocks must be excluded from emphasis/link/header passes (and vice versa). I need to contribute ranges before the styling passes run, and read the source-block ranges.
2. Watermark back-off: an unclosed $$ or \[ at the buffer tail must hold the frontier, like an open fence does.
3. Fenced math: ```math / ```latex fences are structurally code blocks but need to route to the math renderer instead of the code panel. This hooks into --style-source-blocks.
4. Pass ordering: math runs after source blocks (so $$ inside code stays literal) and before tables (so a frozen equation isn't mis-parsed as rows).

A "list of functions to apply at the end" would work for (4) alone. For the rest, the cleanest approach might be:
- A pre-pass hook that receives and can extend the avoid-ranges
- A watermark hook so external packages can hold the streaming frontier back past their own open delimiters (like an unclosed $$)
- A way to register fence languages that get a custom handler instead of the code panel

I don't need all of these on day one; I am happy to discuss which are most practical to start with. The squashed diff should give you a concrete picture of what's needed.

Thanks again!

P.S.: I am super excited about this extension. I can discuss scientific papers downloaded from arXiv directly in Emacs; it displays equations, it shows SVG images from the papers, and it supports all statements with direct links to the LaTeX source. For this I use a simple SKILL.md I wrote to instruct the agent how to take advantage of agent shell for engaging in scientific discussions.

xenodium:
Thanks a lot for consolidating needs.

I had a go at seeing if we could do without multiple hooks, and it looks like we can. I've gone ahead and implemented in https://github.com/xenodium/agent-shell/commit/401009305c7c5605ce6d3ff4f3b64638f2d4111f and https://github.com/xenodium/agent-shell/commit/1a2e0d8af6931951df0f665052c6be3bc23ff556. Mind taking a look and seeing if those integration points are enough for what you need? agent-shell-markdown-render-functions-all-math-cases should give an idea of how to go about it.

> I am super excited about this extension.

Looks great!

> I can discuss scientific papers downloaded from arXiv directly in Emacs;

I'd love to see your workflow if you have a demo video or a blog post.

alberti42:
Thanks for the API! The new additions (hook + frozen tag + watermark
return) work nicely and handle $$...$$, \[...\], and inline \(...\)
with no issues.

One thing is still needed for fenced math blocks (```math / ```latex): my
renderer claims and freezes the block in the hook, but
--style-source-blocks runs after and still matches the fence regex; this
would delete the fence lines and re-styles the body as code, clobbering
the overlay with SVG-rendered LaTeX equation. The function
--style-source-blocks doesn't check agent-shell-markdown-frozen before
processing a block, unlike the other passes; the reason is that this
function normally produces the tag agent-shell-markdown-frozen

It would be very useful if the function could become idempotent, thus
honoring the flag if already set. This could be something like: if the
body is already frozen, skip the block. That way an external renderer
can claim a fenced block by language, freeze it, and the source-block
pass leaves it alone — consistent with how every other pass already
respects the frozen tag.

Without it, I'd have to delete the fences in the hook (like the current 
code actually does), altering the agent's original output. This is not
particularly. Keeping the text intact and overlaying via display properties
feels much cleaner.

xenodium:
> The new additions (hook + frozen tag + watermark
> return) work nicely and handle . . . , [...], and inline (...)
> with no issues.

Good to hear!

> --style-source-blocks doesn't check agent-shell-markdown-frozen before
processing a block

A yes. Added in https://github.com/xenodium/agent-shell/commit/ba0918f6d8e502ff4d8dc6aa871bc8be5e50808f. Please try that. Lemme know how that works out.

alberti42:
While testing the standalone package
(https://github.com/alberti42/agent-shell-math-renderer), a second
issue surfaced in addition to the frozen-tag clobbering in
--style-source-blocks.

The render functions registered on agent-shell-markdown-render-functions
run before the inline-code pass.  This means, the renderer has no way to
know what is inside backticks -- everything inside backticks (inline code) should be preserved untouched.

For example, when the agent outputs ``\(x\)`` (showing a LaTeX snippet `x` as
code), the renderer sees the raw `\(...\)` delimiters and incorrectly
claims the region as math.

The alist passed to each render function already includes :source-blocks
— the list of fenced code blocks in the buffer — so renderers can skip
delimiters inside blocks like
````
```python ... ```
````
Could the alist also include the inline-code spans (backtick ranges)? 
That way a renderer can skip backtick-enclosed content the same way
it skips fenced blocks.

Alternatively, running the render functions after inline-code
detection (but still before the emphasis passes) would also work:
the inline-code regions would already carry the frozen tag, and
renderers could simply check for it -- which my code already does.

**Comment by alberti42** [comment]
>> The new additions (hook + frozen tag + watermark
>> return) work nicely and handle . . . , [...], and inline (...)
>> with no issues.
>
> Good to hear!
>
>> --style-source-blocks doesn't check agent-shell-markdown-frozen before processing a block
>
> A yes. Added in https://github.com/xenodium/agent-shell/commit/ba0918f6d8e502ff4d8dc6aa871bc8be5e50808f. Please try that. Lemme know how that works out.

This was very fast. Awesome.

I can confirm end-to-end: my renderer now claims a fenced

````
```math / ```latex
````

block in the hook by freezing it, and `--style-source-blocks` correctly leaves it alone. The equation is overlaid via a display property with the raw fence lines kept intact in the buffer, so copy/save round-trips the original source! This is neat and clean; no more deleting fence lines to dodge the regex as I was doing before.

Much appreciated!

Screenshot 2026-07-02 at 21 12 15

**Comment by xenodium** [comment]
Looking good!

> This means, the renderer has no way to know what is inside backticks

context param now receives both `:source-blocks` and `:inline-code-ranges` as of https://github.com/xenodium/agent-shell/commit/a894020ddce9c0fc7db81555a8ba7cf0fb124449

> display property with the raw fence lines kept intact in the buffer, so copy/save round-trips the original source!

I would drop the fences altogether, but keep the LaTeX body. This way, when users select the math image and copy it, they copy the LaTeX text only (no need to deal with backticks or fences in clipboard).

**Comment by alberti42** [comment]
Thanks a lot for the review and the suggestion! I followed your advice and
dropped the backtick fences from the rendered math block, so a copy of the
equation no longer carries markdown fences into the clipboard.

One small tweak on top of your idea: instead of keeping the *bare* LaTeX body,
we now wrap it in `\[…\]` display delimiters. The reasoning is round-tripping: the
copied text then renders as math in other modern viewers (GitHub, Obsidian,
etc.) without any further parsing, and it self-identifies as display math rather
than being an anonymous fragment. It also keeps the fenced path uniform with the
`$$…$$`, `\[…\]`, and `\(…\)` delimiter paths, which all already keep their
LaTeX delimiters under the image.

For context, here's how the fenced-math handling has evolved:

| Version | Fences in buffer | Clipboard yields |
|---|---|---|
| Original hook | deleted | `E=mc^2` (bare body) |
| Keep-and-overlay | kept intact under image | ` ```math\nE=mc^2\n``` ` |
| Now (your suggestion) | deleted, replaced with `\[…\]` | `\[E=mc^2\]` |

So we're back to dropping the fences, but replacing them with renderable LaTeX
rather than leaving nothing behind.

Thanks again; this made the output noticeably cleaner.

P.S. Somewhat ironically, this change means we no longer need the
`agent-shell-markdown-frozen` check you added to `--style-source-blocks` for the
math renderer extension: since the region no longer contains a backtick fence, the regex
never matches it, so the `frozen-check` is never even reached on our path. That said, I
still think it's a good addition to keep; it's a sound idempotency property that
any other extension keeping fenced text under an overlay would rely on.

P.S.2 I also checked what happens to the transcript. There we preserve the
agent's original output, with the backtick fence intact:

````
```math
E=mc^2
```
````

I think this is the right choice: the transcript stays faithful to what the
agent actually produced. It's arguably less convenient if you open the file in a
markdown viewer that doesn't render `math` fences (e.g. `markdown-ts-mode` in
Emacs shows it as a plain code block), but that seems a small price for keeping
the transcript a faithful record. (And viewers that do support it — GitHub,
GitLab, … — render `` ```math `` as display math anyway.)

**Comment by xenodium** [comment]
> this change means we no longer need the agent-shell-markdown-frozen check you added to --style-source-blocks for the math renderer extension

agreed. It's good to have anyway.

> Thanks a lot for the review and the suggestion!

Glad it's all working now. I think we can close this PR now? Lemme know when your extension is ready for consumption and I'll link to it from README.

**Comment by alberti42** [comment]
Thank you for a894020. Much appreciated!

 exposing `:inline-code-ranges` in the render-hook
context is exactly what was needed to close the last gap on my side.

Now the hook just takes the `:inline-code-ranges` you hand it and folds those
(marker) ranges straight into its avoid-ranges, so inline math inside a code
span stays literal while real `\(…\)` beside it still renders. No more need of
`define-advice`.

The integration is now clean! The renderer now plugs into agent-shell purely
through the public hook surface, with no advice or any other workaround left.

I sketched a first README.md in

https://github.com/alberti42/agent-shell-math-renderer

In due time, I will add screenshots and other goodies like SKILLS to fetch
arXiv papers (all modern papers in physics/math/CS/stats are always submitted
outside of paywalls on https://arxiv.org/). This is perfect for an agent because it can
fetch the LaTeX source text, and you can start discussing science with direct
references to the paper itself. A great aid for beginners and experts to read papers
more effectively.

After a few weeks of testing it, I also plan to submit it to MELPA.

I would say we could close the PR. What do you think?

With your help, the migration to a standalone package was smooth and fast! Big thanks!

P.S. One small thing that can be helpful in the future: the renderer relies on
three of your internal (`--`) functions:

- `agent-shell-markdown--in-avoid-range-p` — avoid-range lookup
- `agent-shell-markdown--sort-ranges` — range merging
- `agent-shell--cache-dir` — the shared cache directory

They're exactly the utilities an external renderer needs, so I'm glad if I can
keep using them. Since the `--` names mark them private, though, a rename or a
signature change on your side would break the package. Would you mind leaving a
short note in the source (a comment, inside or beside the docstring—just for
you and your agent) that `agent-shell-math-renderer` depends on these? If you ever
do need to change or rename them, it would be a great courtesy if you could open
a quick issue at https://github.com/alberti42/agent-shell-math-renderer/issues
so I can keep up. (If you'd rather promote them to public API instead, that
works just as well; I have no strong preference as long as I get a heads-up on
changes.)

**Comment by xenodium** [comment]
> * `agent-shell-markdown--in-avoid-range-p` — avoid-range lookup
> 
> * `agent-shell-markdown--sort-ranges` — range merging
> 
> * `agent-shell--cache-dir` — the shared cache directory

All three renamed (public) in https://github.com/xenodium/agent-shell/commit/7ee6ff90bf7928d6774c1aab84865fd8362b9957

> I would say we could close the PR. What do you think?

Yup. We're good, unless you find something else needed. Closing.

> With your help, the migration to a standalone package was smooth and fast! Big thanks!

Welcome! Glad its working well for ya.

**Comment by xenodium** [comment]
btw, you can also do something like:

```emacs-lisp
(use-package agent-shell-math-renderer
  :vc (:url "alberti42/agent-shell-math-renderer" :rev :newest))
```

alberti42:
> btw, you can also do something like:
> 
> ```emacs-lisp
> (use-package agent-shell-math-renderer
>   :vc (:url "alberti42/agent-shell-math-renderer" :rev :newest))
> ```

Thanks! I now documented the :vc installation alternative in the README.

alberti42:
Hi Álvaro,

If your offer

> I'd be happy to link to it like others in https://github.com/xenodium/agent-shell#related-projects

is still valid, I would much appreciate if you could list the package there. Thanks!

P.S.: I have now removed (reverted) recent changes supporting the deprecated markdown-overlays-put (overlay renderer); I just kept a note in the README ([link](https://github.com/alberti42/agent-shell-math-renderer#history-experimental-overlay-renderer-support-removed)) should anyone still need it.

xenodium:
> is still valid, I would much appreciate if you could list the package there. Thanks!

Yup. https://github.com/xenodium/agent-shell/commit/ef4605e231edf6a7216b432f80f44682ac6da6a3

alberti42:
Hi Álvaro, following on from when you kindly made `agent-shell-markdown-sort-ranges`, `agent-shell-markdown-in-avoid-range-p`, and `agent-shell-cache-dir` public for the math renderer (7ee6ff9), I have one more request in the same spirit.

**Background.** My package renders math in the streamed agent output by attaching a function to `agent-shell-markdown-render-functions`.  Each such function is handed a `context` alist with `:source-blocks` and `:inline-code-ranges`, which it uses to avoid rendering math inside code.  That works great for streamed output.

I now also want to render math in a *static* region, i.e., in a user prompt after it is submitted.  That text is not streamed, so it never passes through the render-functions hook.  To handle it I need to build the *same* `context` alist myself, for an arbitrary narrowed region.

**The problem.** There is currently no public way to obtain that context.  The code that builds it lives inline inside `agent-shell-markdown-replace-markup` (right after the `narrow-to-region` near the top): it calls the private `agent-shell-markdown--source-blocks` and `agent-shell-markdown--inline-code-ranges`, derives the source ranges, wraps the inline-code ranges in markers, and packs everything into the alist it then passes to the render functions.  So to reproduce it I would have to call those private functions and re-implement those same lines on my side; this is the kind of duplication that may drift out of sync with your code over time.

**The request.** Would you consider extracting/refactoring that assembly into a single public function — say `agent-shell-markdown-context` — that runs on the current narrowed region and returns the `((:source-blocks . …) (:inline-code-ranges . …))` alist? And then have `replace-markup` call that new function in place of its current inline block.

That last part (routing `replace-markup` through the new function too) avoids duplicate code inside agent-shell and ensures the public function and `replace-markup` share the same code, so they cannot diverge in the future.

I'm happy to open a PR that does exactly this if it saves you time — just let me know.

Thanks again! Andrea

alberti42:
Hi Álvaro, I allowed myself to sketch a draft of what the required code can be. In this way, I could test end-to-end that the new public API is exactly what is needed for correctly rendering the math equations in the user prompt. You find the suggested code in PR #717. Cheers, Andrea

xenodium:
Thanks Andrea. Looks good to me. Merged it.

alberti42:
Thanks, Álvaro, for endorsing `agent-shell-math-renderer` in https://xenodium.com/agent-shell-0-63-updates ❤️

xenodium:
Thanks for building it :)

PR #676, opened 2026-06-26
Render agent image content blocks inline

melito:
`agent_message_chunk` handler extracts only `(content text)` from a content block.

When the agent returns an image content block that block has no `text` field and it gets silently dropped and never rendered.

I've added a new helper `agent-shell--content-block-to-markdown` that returns text for text blocks and a markdown image (``) for image blocks that carry a file `uri`

I've added a simple test at `agent-shell--content-block-to-markdown-test`

I did not file a feature request first.
I was hacking on something using agent-shell and noticed nothing was rendering so took a shot at alleviating my problem.
Open to discussing correct solution if this diverges from any inflight agent-shell work or steps on toes.

Screenshot 2026-06-26 at 1 09 59 PM

## Checklist

- [] I've filed a feature request/discussion for a new feature.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1911,20 +1911,21 @@ pretty-printed JSON inside a json fence."
            ;; below the transcript's ## section headers.  Applied
            ;; per-chunk: if a header is split across chunks it may
            ;; not be indented (graceful degradation).
-           (agent-shell--append-transcript
-            :text (agent-shell--indent-markdown-headers
-                   (map-nested-elt acp-notification '(params update content text)))
-            :file-path agent-shell--transcript-file)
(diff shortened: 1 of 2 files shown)


xenodium:
Thanks for the improvement!

xenodium:
Heads-up, we also needed base64 handling which is the actually required protocol field. Added in https://github.com/xenodium/agent-shell/commit/df1b2ea244d9f10339ec4cd63f4fe26e340a2800

melito:
Excellent!  Glad to help.  Thanks for covering the base64 data case.  Will follow up if I come up with anything else useful.

PR #686, opened 2026-07-02
Fix command to pull Qwen2.5 model in README

jcubic:
Corrected mistake in README. Replace the wrong command 'opencode pull' with 'ollama pull'.


xenodium (approved):
Thank you!

PR #687, opened 2026-07-02
Record session ID and model in transcript header

alberti42:
## Summary

Adds `**Session ID:**` and `**Model:**` lines to the transcript file header
when those values are known.

Today a transcript's only identity is its timestamped filename; nothing in
the file ties it back to the session that produced it. Recording the session
ID makes transcripts self-identifying — a file can be correlated with a
session returned by `session/list`, which is what external tooling (and
session-aware features built on top of transcripts) needs to locate the right
file. The model is recorded alongside it so the transcript captures which
model produced the conversation.

## Header, before → after

Before:

```
# Agent Shell Transcript

**Agent:** Claude
**Started:** 2026-07-02 14:19:52
**Working Directory:** /path/to/project
```

After:

```
# Agent Shell Transcript

**Agent:** Claude
**Started:** 2026-07-02 14:19:52
**Working Directory:** /path/to/project
**Session ID:** 8f3c1a20-...-abc
**Model:** claude-opus-4-6
```

## Notes

- Both fields are conditional: if the value is `nil` the line is omitted, so the header degrades gracefully.
- Timing is safe: the header is written lazily on the first transcript append, which happens after the ACP handshake; so the session ID and model are already in state by then.
- Backward compatible: only new transcripts gain the lines, and they follow the existing `**Key:** value` header convention, so existing files and any header parser are unaffected.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -8286,13 +8286,14 @@ Returns the file path, or nil if disabled."
           (let ((agent-name (or (map-nested-elt agent-shell--state '(:agent-config :mode-line-name))
                                 (map-nested-elt agent-shell--state '(:agent-config :buffer-name))
                                 "Unknown Agent"))
-                (session-id (map-nested-elt agent-shell--state '(:session :id))))
+                (session-id (map-nested-elt agent-shell--state '(:session :id)))
+                (model-id (map-nested-elt agent-shell--state '(:session :model-id))))
             (write-region
              (format "# Agent Shell Transcript
 
 **Agent:** %s
 **Started:** %s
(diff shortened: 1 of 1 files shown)


xenodium:
LGTM. Happy to merge if you are.

alberti42:
> LGTM. Happy to merge if you are.

Yes, I am happy with the PR. I first staged it because I wanted to have it tested in real situations. Everything works. So, I switched from Draft to Ready for review. Thanks!

xenodium:
Merged. Thanks for the PR!

PR #689, opened 2026-07-02
Support angle-bracketed link/image destinations

alberti42:
## Motivation

**Who emits these links?** The agent does. agent-shell renders the markdown the agent streams into the buffer, and an agent can be instructed — via a skill, system prompt, or tool — to **embed local images inline** by emitting image markup with an absolute path. Concretely, I drive a skill that has the agent embed figures/SVGs from local files directly in the transcript (``), so the rendered conversation shows the graphics next to the discussion.

Those absolute paths routinely contain **spaces** (cache dirs, `~/Google Drive/...`, figure filenames, etc.). Two things have to hold for that to be useful:

1. **It renders in agent-shell.** Today it *happens* to, because the renderer matches a destination as everything up to the first `)` — spaces included. But that leniency is **not** standard markdown: in a conformant CommonMark parser a bare `(url)` destination ends at the first space, so the path is truncated.
2. **It survives leaving agent-shell.** These buffers get **saved and copy-pasted into other markdown viewers** (docs, GitHub, editors, static-site generators). A transcript that relied on agent-shell's lenient bare-space handling silently breaks there — the image won't load, or only part of the path is used.

The standard, portable way to write a destination containing spaces is to wrap it in angle brackets. Per the [Markdown links reference](https://www.codecademy.com/resources/docs/markdown/links) — *"How do you Markdown a URL with spaces?"*:

> Either encode spaces as %20 or enclose the entire URL in angle brackets: `<URL with spaces>`. For file paths with spaces, use proper encoding or quotes.

This is also one of the two link-destination forms defined by the [CommonMark spec](https://spec.commonmark.org/current/#link-destination). This PR makes agent-shell **render** that form, so a skill can emit spec-compliant `alt` that both displays inline *and* stays valid markdown when the transcript is reused elsewhere.

## What

Match CommonMark's angle-bracketed link/image destinations:

- `[text](<url>)`
- `alt`

## How

- New shared helpers `agent-shell-markdown--link-markup-regexp` / `--link-markup-url` that match **both** the bare `(url)` and angle-bracketed `(<url>)` destination forms and return the URL from whichever matched.
- `--replace-links` and `--replace-images` both use them, replacing their duplicated inline `rx` destination patterns.
- The bare form is byte-for-byte unchanged (regression-covered); the angle form strips the brackets and permits spaces and parentheses in the URL.

## Scope

Bare image-path lines (`--replace-image-file-paths`) are a non-CommonMark extension and are intentionally left as-is.

## Tests

**ERT.** Added coverage (angle-bracket link, remote-image angle fallback, space/paren-containing URLs, empty-alt image, bare-form regression). Full suite: `102/102` passing.

\`\`\`
emacs -batch -l ert -l tests/agent-shell-markdown-tests.el -f ert-run-tests-batch-and-exit
\`\`\`

**End-to-end in a live Emacs session.** Loaded the branch and had an agent emit an angle-bracketed absolute path, ``, in its streamed response. agent-shell rendered it as an inline SVG — the markup collapses to the alt placeholder carrying a `display` image whose `:file` is the correct, un-truncated path. Bare `` and existing `[text](url)` links continue to render as before.

### DIFF
--- a/agent-shell-markdown.el
+++ b/agent-shell-markdown.el
@@ -257,9 +257,9 @@ For example:
   (agent-shell-markdown-convert \"_my_ **text**\")
   => #(\"my text\" 0 2 (face italic) 3 7 (face bold))"
   (agent-shell-with-work-buffer
-   (insert markdown)
-   (agent-shell-markdown-replace-markup)
-   (buffer-string)))
+    (insert markdown)
+    (agent-shell-markdown-replace-markup)
+    (buffer-string)))
 
 (cl-defun agent-shell-markdown-replace-markup (&key force
                                                     (render-images t)
@@ -702,8 +702,8 @@ face `agent-shell-markdown-inline-code' on \"code\"."
                 (put-text-property markup-start end
                                    'agent-shell-markdown-source source)))))))))
 
-(defun agent-shell-markdown--link-markup-regexp (imagep)
-  "Return a regexp matching link (or image, when IMAGEP) markup.
+(cl-defun agent-shell-markdown--link-markup-regexp (&key as-image?)
+  "Return a regexp matching link (or image, when AS-IMAGE?) markup.
 
 The destination accepts both the bare `(url)' form and the
 CommonMark angle-bracketed `(<url>)' form, the latter allowing the
@@ -714,9 +714,9 @@ group 2 is the angle-bracketed destination body; group 3 is the
 bare destination body.  Exactly one of groups 2 and 3 participates
 in any match — read the URL from whichever did."
   (rx-to-string
-   `(seq ,@(when imagep '("!"))
+   `(seq ,@(when as-image? '("!"))
          "["
-         (group (,(if imagep 'zero-or-more 'one-or-more) (not (any "]"))))
+         (group (,(if as-image? 'zero-or-more 'one-or-more) (not (any "]"))))
          "]"
          "("
          (or (seq "<" (group (zero-or-more (not (any "<" ">" "\n")))) ">")
@@ -748,7 +748,7 @@ For example, the buffer \"see [docs](https://example.com)\"
 becomes \"see docs\" with face `agent-shell-markdown-link' on \"docs\"
 and a keymap that opens the URL."
   (let ((case-fold-search nil)
-        (regexp (agent-shell-markdown--link-markup-regexp nil)))
+        (regexp (agent-shell-markdown--link-markup-regexp)))
     (goto-char (point-min))
     (while (re-search-forward regexp nil t)
       (let* ((markup-start (match-beginning 0))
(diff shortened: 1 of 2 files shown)


agent-shell-markdown.el
```diff
@@ -605,6 +607,35 @@ face `agent-shell-markdown-inline-code' on \"code\"."
                                    '(agent-shell-markdown-frozen t
                                      rear-nonsticky (agent-shell-markdown-frozen))))))))))
 
+(defun agent-shell-markdown--link-markup-regexp (imagep)
```
xenodium: nit: mind changing to cl-defun and make &key param :as-image? makes calling code more explicit.

alberti42:
I put this PR back to `draft`; the reason is that I noticed a wrong behavior while I was developing the math renderer. I think it is unrelated. I could not reproduce it. So, most likely the code of this PR is fine. I had it tested end to end and ran different tests. But just as a precaution, I prefer to run with it for another couple of days observing whether everything works fine.

xenodium:
Sounds good

alberti42:
This is a short message to give you a heads-up. I found the origin of the misbehavior. It had nothing to do with this PR. I filed a separate PR #713, which fixes two bugs related to embedding markdown images. When these are fixed, I will resolve conflicts and re-review this PR. Once then ready for review, I will switch from draft to ready-for-review.

alberti42:
I rebased it. Everything looks good to me. Indeed, the issues I had experienced before were related to #713. After the merge of #713 upstream, all issues are now gone. This PR is now ready for review.

agent-shell-markdown.el
```diff
@@ -2106,64 +2130,64 @@ CJK) so right borders align across rows.  Without it,
 measurement falls back to `string-width' — fine for ASCII but
 prone to a few-pixel drift on emoji-heavy tables."
   (agent-shell-with-work-buffer
```
xenodium, line 2175: Are the whitespace changes necessary? Does TAB indent in elisp mode yield the new indentation?

xenodium (approved):
Thank you. Looks good to me, with a couple of minor changes suggested in comments.

agent-shell-markdown.el
alberti42: Yes, I changed it to 4-space indentation instead of 3. It was my mistake: I applied TAB on symbols that were not evaluated yet, which produced 3 spaces instead of 4.

agent-shell-markdown.el
alberti42: Yes, I adjusted the code accordingly. I added a new commit for both requests. If you prefer a single commit, just squash the second into the first; or I can do it for you. Just let me know what you prefer.

xenodium:
Thank you for the PR and follow-up changes.

PR #693, opened 2026-07-03
Fixes for cline agent

nhojb:
Some fixes for the `cline` agent:

- Guard `agent-shell--refresh-session-title` by `session-list` capability. 
   Cline does not currently support `session-list`.
- Add default model and session-mode defcustoms for cline.
- Match config category options by `:id`. 
    Cline tags both its `provider` and `model` options with category `model`.  Matching on category alone would resolve `model` to the first match (`provider`); the lookup must prefer the option whose `:id' equals the category. This fixes the switch-model and switch-provider actions.

To test cline in agent-shell:

```
(setq agent-shell-cline-environment
   (agent-shell-make-environment-variables :inherit-env t))
(setq agent-shell-cline-default-model-id "gpt-5.5") ;; your default model here
```

Note: Setting the default model is required, otherwise cline acp will use a hard-coded model which fails silently.
  
## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 /.agent-shell/
+agent-shell-autoloads.el
 
 *.elc
+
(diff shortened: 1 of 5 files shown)


xenodium (approved):
Thanks!

Issue #587, opened 2026-05-18
Long region preview's "Expand..." button is sent literally to the agent


alberti42:
When I send a region longer than 5 lines (e.g. via `agent-shell-send-region`), the inserted preview is capped and followed by an `Expand...` button (`agent-shell.el:6113-6135`, `max-preview-lines 5`). If I press `RET` to submit without first clicking the button, the agent receives the truncated preview plus the literal string `Expand...` — the full text never reaches it.

Cause: the full text is captured only inside the button's action closure. `shell-maker-submit` → `comint-send-input` reads the prompt with `buffer-substring-no-properties`, so the closure is never consulted.

Possible fixes:
1. Auto-expand any collapsed `agent-shell-region-id` regions in the input area before submission.
2. Store the full text on a text property (or invisible overlay) that submission can substitute back in, so the visible preview stays compact but the agent receives the full content.
3. At minimum, expose `max-preview-lines` as a defcustom so users can disable the cap.

**Repro:**
1. Open a buffer with > 5 lines of text, mark a region spanning all of them.
2. `M-x agent-shell-send-region`.
3. Press `RET` without clicking `Expand...`.
4. Observe that the agent sees `…\nExpand...` instead of the full region.

## Checklist

- [x] *I agree to communicate with the author myself* (not AI-generated).
- [x] I've read the README's [Filing issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#filing-issues) section.
- [x] I'm running the latest versions (fill in below).
- Emacs: GNU Emacs 31.0.60
- agent-shell: `67babec` (2026-05-16, merge of #583)
- acp.el: `03099fa` (2026-05-03, merge of #18)
- ACP package: OpenCode adapter (`agent-shell-opencode.el`, bundled with agent-shell)
- Agent CLI: `opencode` `0.0.0-integrated-202605150926` (personal fork with `--attach` support; see config below)
- [ ] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [ ] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README).

-- Thanks a lot for your consideration! Andrea

xenodium:
Thanks for filing.

This is kinda working as expected. The idea is not to fill up the prompt with too much text and let the agent expand it since the exact location of the text is included in the file reference. Is the agent not expanding the content itself?

[...]

PR #694, opened 2026-07-03
Send selected regions as fenced code blocks

alberti42:
## Why

When I send a region to an agent, I prefer to send it as a proper fenced
code block than as a `path:start-end` reference plus a bare numbered preview
in loose prose, as the current behavior.

Three reasons:

- The agent gets cleaner, more familiar input: a fenced block with a
  language tag and a `cat -n`-style body. This is the shape coding agents
  emit and consume for file contents.
- A fenced block makes it much easier to type prose above and below the snippet
  in the same prompt without the pasted lines bleeding into my own text.
- An automatic info string characterizing the fenced blocks; this is a nice
  touch that guides the agent to know what type of input it received (C code, elisp,
  LaTeX, etc).

I have already been doing this on my Emacs setup using "advices"; however, this
solution relied on brittle "advices", and I genuinely believe this solution represents
an improvement for all users of agent shell.

This is all text sent to the agent, so there's no change to any API or
on-disk format.

However, if users for some reason want to opt out from the new behavior, there is a
defcustom: by specifying '`unfenced` instead of `'fenced`, content is passed following
the old behavior. 

## What changed

Three commits, reviewable independently:

### 1. `agent-shell-region-context-style` — fenced code block for regions

New `defcustom` controlling how a selected region is formatted as context.
It's consumed at the single chokepoint `agent-shell--get-region-context`, so
it covers both `agent-shell-send-region` and the `region` source of
`agent-shell-send-dwim`.

- `fenced` (new default): a fenced block whose info string carries the
  buffer language and a clickable `path:line-range`, with the existing
  capped, numbered body (Expand button and all) inside the fence.
- `unfenced`: a clickable `path:line-range` followed by the numbered
  preview, without the fence.

`unfenced`:

```
    path/to/file.el:359-361

    359       (foo)
    360       (bar)
    361       (baz)
```

`fenced`:

````
```elisp path/to/file.el:359-361
359   (foo)
360   (bar)
361   (baz)
```
````

The language tag comes from a new `agent-shell-language-alist` (major mode →
fence identifier, e.g. `emacs-lisp-mode` → `elisp`, `c++-mode` → `cpp`) via a
small helper `agent-shell--language-for-mode`, which falls back to stripping
the `-ts-mode`/`-mode` suffix (so `python-ts-mode` → `python`). This replaces
the inline suffix-strip previously used to compute a region's `:language`.

### 2. Number line context as `cat -n`

`agent-shell--get-numbered-region` previously emitted `   N: line`. This
switches it to the `cat -n` shape (right-aligned line number, a tab, then the
line) — the format agents already produce and read for file contents. It
also touches the error/diff context previews built on the same helper (the
trim regexp and the error-line highlight in `agent-shell--format-diagnostic`,
now `-> N<tab>`).

Before: `   360: (bar)`
After:  `360\t(bar)`

### 3. Collapse single-line range in the header

A single-line selection previously showed its header range as `path:N-N`.
It now emits `path:N`: for one line the header already pins the exact line, so
the range form is redundant, and the body carries the raw line unnumbered
(this preserves the existing `;; Same line region? Avoid numbering.` path):

````
```elisp path/to/file.el:359
(foo)
```
````

## Compatibility

Both defaults change the *text sent to the agent* (region context is now
fenced; numbered previews are now tab-separated), but nothing user-facing
breaks and there's no API/format change. `agent-shell-region-context-style`
can be set to `unfenced` to restore the old region format.

If you'd rather these be opt-in, I'm happy to flip the default of
`agent-shell-region-context-style` to `unfenced`, and/or split the `cat -n`
change into its own PR since it has the wider blast radius.

## Tests

`tests/agent-shell-tests.el`:

- `agent-shell--language-for-mode-test` — alist overrides win, else the
  suffix strip; caller-supplied alist respected.
- `agent-shell--get-region-context-style-test` — both `fenced` and
  `unfenced` output.
- `agent-shell--get-numbered-region-test` updated for the `cat -n` shape.

---

If you wish for any change, I am happy to follow your lead.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -357,6 +357,54 @@ Sources are checked in order until one returns non-nil."
                          (function :tag "Custom function")))
   :group 'agent-shell)
 
+(defcustom agent-shell-region-context-format-function nil
+  "Function to format a selected region before it is sent as context.
+
+Affects \\[agent-shell-send-region] and the `region' source of
+\\[agent-shell-send-dwim] (see `agent-shell-context-sources').
+
+When nil (the default), the region is sent as a clickable
+`path:line-range' header followed by a capped, line-numbered preview.
+
+When set to a function, it is called with a single plist argument and
+should return the string to send in place of the default formatting, or
+nil to fall back to it.  The plist has these keys:
+
+ :file        Absolute path of the source file, or nil for a buffer with
+              no file.
+ :language    `major-mode' name with the `-ts-mode'/`-mode' suffix stripped
+              (e.g. \"python\").  This is a raw hint, not a canonical
+              Markdown fence identifier -- for example `c++-mode' yields
+              \"c++\", not \"cpp\"; map it yourself if you need a fence tag.
+ :line-start  Line number of the first selected line (1-based).
+ :line-end    Line number of the last selected line (1-based).
+ :content     The selected text, verbatim.
+ :agent-cwd   The target agent's working directory.  Use it to derive a
+              path relative to the agent (as the default does for :file),
+              or nil if unavailable.
+ :agent       Identifier symbol of the target agent (e.g. `claude-code'),
+              or nil if unavailable.
+ :model       Current model id of the target agent (e.g.
+              \"claude-opus-4-8\"), or nil if unavailable.
+
+The keys above are what the function cannot otherwise reach: the target
+shell (:agent-cwd :agent :model) and the selection, which is already
+deactivated by the time the function runs (:content :line-start
+:line-end).  The source buffer itself is not passed because it is the
+current buffer while the function runs: reach it via `current-buffer'
+(and hence `buffer-name', `default-directory', buffer text, overlays,
+...).
+
+This is an escape hatch for tailoring region context to a particular
+agent, model, or workflow -- for example, sending only the
+`path:line-range' header without the line-numbered body, or wrapping the
+body in a fenced code block with a language tag.  The best shape can
+differ across agents and model versions, which is why :agent and :model
+are provided for the formatter to branch on."
+  :type '(choice (const :tag "Default formatting" nil)
+                 (function :tag "Custom formatter"))
+  :group 'agent-shell)
+
 (cl-defun agent-shell--make-acp-client (&key command
                                              command-params
                                              environment-variables
@@ -7829,6 +7877,7 @@ When PICK-SHELL is non-nil, prompt for which shell buffer to use."
                            (agent-shell--shell-buffer)))
          (text (agent-shell--get-region-context
                 :deactivate t
+                :shell-buffer shell-buffer
                 :agent-cwd (with-current-buffer shell-buffer
                              (agent-shell-cwd)))))
     (if (with-current-buffer shell-buffer (shell-maker-busy))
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by xenodium** [comment]
Thanks for the PR!

Unfortunately, this may be an area where our subjective preferences may be at odds. While there are some familiar traits to the approach, I'm thinking we may be conflating two of them while generating a somewhat non-standard third one.

AFAIK, block headers do not typically cater for additional info beyond the language. Similarly, the body does not cater for line numbers. This one may be the more challenging one of the two since the body content should be valid in the above-mentioned language. Fenced blocks are typically selectable and pastable (ie. copy, paste, and run this python snippet).

I could try to give more arguments here, but frankly recognise this is a subjective area and I happen to have a preference for the existing display. Should I receive many more requests for this format, I would be much more open to the idea, but at the end of the day they are somewhat equivalent thus suspect it's mostly personal preference to go with one way vs the other.

We may be tempted to just add another setting and keep the existing default, but I also have to weigh the added complexity and configuration point for somewhat equivalent offerings. We have lots of options already (typically driven by personal preferences) so I regularly need to factor in whether another defcustom is necessary.

Sorry, probably not the outcome you were after.

> I have already been doing this on my Emacs setup using "advices"; however, this solution relied on brittle "advices"

No doubt advices aren't the first solution to reach for, but hey they sure are the thing that often enable us to have our Emacs cake and eat it too. Sounds like they have been serving you well in this case. It's OK to continue using them.

**Comment by alberti42** [comment]
Hi Álvaro,

Thanks for the very quick feedback. It was useful and gave material to think about.

Allow me to first start with a clarification. I like your current format! And my PR was never about improving on the current style to satisfy my personal taste for how context region should be formatted. My goal was to pick the format that works best for the LLM on the receiving end.

For example, if you query Claude, it has a strong preference for `cat -n` formatted data.

If you ask Codex, it also has a preference for `cat -n` when reviewing code, but also for a format like
 
```
 37 │ def foo():
 38 │ return 1
```

Your feedback gave me something to think about. Why don't we keep the current default but add a single injection point, defaulting to nil (i.e., current behavior), that could be called `agent-shell-region-context-format-function`. When set, it's called with a single `plist` and returns the formatted string, or nil to fall back to the current default:

```elisp
(:file          "src/foo.py"
 :line-start    37
 :line-end      42
 :content       "def foo():\n    return 1"
 :source-buffer #<buffer foo.py>
 :agent         "claude-code"
 :model         "claude-opus-4-8"
 :agent-cwd     "/home/me/project/")  ; for path relativization
```

This would be a minimal change to the current code. It preserves the current default; but it also gives a way to power users to write an optimized formatter for their agent.

Your feedback was important because it made me think there is not a single preference that fits all situations. Different agents may have different preferences when it comes to format; the best format may also depend on the particular workflow. Realistically, it would be impossible for `agent-shell` (essentially for you) to maintain the myriad of possible formats.

And what is the best format today for model Opus 4.8 may be different in Fable 5.0; I mean, the best format may depend on the agent, workflow, and version of the model too.

If we provide such a minimum change `agent-shell-region-context-format-function`, we are empowering power users to pick the best format for their workflow.

Should I replace my commits with a single, much more scoped commit adding `agent-shell-region-context-format-function` ? Or would you prefer to implement it yourself?

Cheers, Andrea

alberti42:
P.S.: I updated the previous message to include a clearer example of the `plist` I am thinking of. Thanks!

xenodium:
Thanks for this Andrea.

> My goal was to pick the format that works best for the LLM on the receiving end.

Aren't we somewhat looking to fix something that ain't broken? Are we currently experiencing issues with the models not parsing the snippets correctly? From there, we could look into ways to fix it, but as far as I can tell this isn't an issue. Models are fairly forgiving in the input format.

> For example, if you query Claude, it has a strong preference for cat -n formatted data.

Are we sure that "emits X" implies "parses X better"? I can think of at least one scenario where I've experienced somewhat similar but opposite asymmetry. Early LLMs would easily digest org markup but would often ignore requests for org markup in their output.

> Different agents may have different preferences when it comes to format

*May* is likely key here. I'd prefer to address this once proven than pre-emptively adding possibly redundant logic.

Do we have evidence that sending in one shape or the other has measurable benefits? If both formats yield equivalent results, then doesn't picking one or the other become more about personal taste?

> Why don't we keep the current default but add a single injection point, defaulting to nil

If we've established formats are equivalent in the eyes of current LLMs, then do we really need another setting? In this particular case exposing a fairly explicit contract with 8 values.

alberti42:
Hi Álvaro,

Thanks for your consideration and patience.

> Aren't we somewhat looking to fix something that ain't broken?

There is nothing fundamentally wrong. It was just me trying to optimize/polish an already beautiful package (agent-shell).

> Do we have evidence that sending in one shape or the other has measurable benefits? If both formats yield equivalent results, then doesn't picking one or the other become more about personal taste?

I did more experiments this afternoon on real-world situations. It happens I was editing some academic texts.

I found out, both with Claude and Codex, that even though they had already read the full document into context so we could discuss it, if I give them an excerpt and ask a question or to make a small edit, they always end up re-reading the text anyway.

I asked the agents and I was told that this is their normal behavior to ensure they are using the authoritative, correct source directly from the file; LLMs are probably trained to go for the direct source rather than trust the input of the user. I can imagine that this avoids issues where the user sends wrong information, making it impossible for the agent to diagnose what the real problem is.

I attach a screenshot below to show what I mean. You clearly see that the agent dispatched the "read" tool even though it had the whole file in context and I even copied the excerpt I wished to edit.

Sending the numbered lines is mostly a waste of tokens, since the agent tends to ignore it and re-read the same content. So, this is a 180° turn from what I was trying to do initially. I am inclined to think that we don't need to feed all numbered lines, even less so with fenced blocks as I originally proposed.

What I am now doing in my setup, I only send the filename and the range of lines I want to discuss (essentially your header, without all the subsequent lines).

Now, regarding the PR and the idea of providing a formatter `agent-shell-region-context-format-function`, the concrete case I see is exactly the one I described. In my case, I would prefer to still provide the header but remove the lines entirely because they just waste tokens.

To be clear:

- I think it is a fine-tuning / fine optimization of agent-shell. I have no bad feelings whatsoever if you decide in the end to turn the proposal down.
- Probably there are ways to instruct agents not to re-read content already provided by the user, as agent-shell does. I am sure one can write a hard gating for the agent that forbids the duplicate reading of the same information. It is just more tedious and requires setting up special agent instructions.

Overall, thanks a lot for the exchange. I am happy with whatever you decide (like it is now, or with the formatter function).

Cheers,
Andrea

Screenshot 2026-07-04 at 23 30 21

alberti42:
Hi Álvaro,

I've reworked the PR down to the minimal version we ended up discussing, so it's concrete enough for you to say yes or no to.

I dropped the three original commits (the fenced blocks and the `cat -n` numbering). There's now a single commit that changes no default behavior at all.

What it adds is `agent-shell-region-context-format-function`, which is nil by default, consumed at the single chokepoint `agent-shell--get-region-context` (so it covers `agent-shell-send-region` and the `region`/`line` sources of `send-dwim`). When it's nil (i.e., the default) nothing changes: you still get the clickable `path:line-range` header plus the capped, numbered preview, byte for byte. When set, it's called with a plist and returns the string to send, or nil to fall back.

On the "8 values" worry from last time: I tried to keep the contract small. The plist only carries what the formatter genuinely cannot reach on its own:

- from the target shell, which the formatter has no handle to: `:agent-cwd` `:agent` `:model`;
- the selection, which is already deactivated by the time the function runs, so it can't re-derive it: `:content` `:line-start` `:line-end`;
- plus `:file` and `:language` for convenience; in principle, they can be derived from `(current-buffer)`, so if it is important for you to minimize the keys in the plist, we could drop these two.

I chose to *not* pass the source buffer, because it's the current buffer (i.e., directly reachable with `(current-buffer)`) while the function runs; so `current-buffer` and hence its name, directory, text, overlays are already available. And `:language` is just the raw major-mode name with the suffix stripped, not a canonical Markdown identifier — it's the same value agent-shell already computed for the region before this PR. If someone wants a real fence tag they map it themselves, which is exactly why there is no reason to provide a mode→language table in the core (I proposed one in my original PR, but dropped it here).

So, this PR is no longer about "format X parses better than format Y." It rather empowers power users to choose the best shape for their agents and model versions.

I think it is overall a nice addition, without adding much burden of code that you will have to maintain in the future; it is more a refactor of currently existing code, plus a `defcustom` for the user to specify `agent-shell-region-context-format-function`. However, if you still don't like this idea, then I suggest you close/reject it. No hard feelings. It was a good exchange either way.

One review note: please look at the diff with `?w=1` (ignore whitespace):

https://github.com/xenodium/agent-shell/pull/694/changes/4fcc107cfa597bc9c20e7cc9a093923faa9f3586?w=1

The existing formatting block is now nested one level deeper (inside an `or`), so it got re-indented. Without `?w=1` the diff makes it look like a large block was rewritten.

Cheers,
Andrea

xenodium:
> It rather empowers power users to choose the best shape for their agents and model versions.

Sorry Andrea. The proposal still looks like we're trying to solve a non-problem. It's not clear what users would get from the models that they can't already get today. Similarly, we've yet to produce evidence of a model not understanding the current format or preferring something else.

alberti42:
> Similarly, we've yet to produce evidence of a model not understanding the current format or preferring something else.

It is just a matter of token economy. All my tests showed me that the agent, stubbornly, always re-read the lines provided in the input. Did you pay attention that the agent always uses its own tools to reread the file because, by default, it does not trust what the user had provided? To be precise, the agent is smart and does not re-read the entire file but only the block of lines provided in the input.

Hence, I was thinking: if the agent is so meticulous that anyway it re-reads the entire block of lines, the best approach is to just provide the relative file name and range of lines.

I am not saying that the agent is confused; just that we are providing the same input twice, the first time in the prompt and the second time by the agent that re-fetches the same content because by default, it is programmed not to trust the user.

I am just curious: do you experience a different behavior where your agent does not use its read tool to re-fetch the same block of lines? I only tested Claude and Codex.

So, in conclusion, mine is merely a fine-tuning to save tokens.

If it does not bother you that the same content is loaded twice in the context, just keep the current code. I tried my best to make minimal changes. But if the current PR shape does not convince you, I totally respect it. Then the best is to close/reject it. I have absolutely no bad feelings.

Thanks again for reading through this long exchange!

xenodium:
> Did you pay attention that the agent always uses its own tools to reread the file

Yup. Looks like we've gone full circle.

From https://github.com/xenodium/agent-shell/issues/587#issuecomment-4621290697

> The idea is not to fill up the prompt with too much text and let the agent expand it since the exact location of the text is included in the file reference

PR #696, opened 2026-07-04
Strip source-buffer properties from grabbed region context

liaowang11:
agent-shell--get-region-context built its single-line file-link preview with `buffer-substring`, and `agent-shell--get-numbered-region` built its numbered preview the same way. `buffer-substring` copies all text properties, so when the source buffer was a `markdown-mode` or `poly-markdown` buffer, `markdown-mode`'s font-lock leaked into the grabbed context: the emphasis markup characters carried `'invisible markdown-markup'` + `'markdown-markup-face'` and the enclosed text carried italic faces. `_hello_world_` then showed up in the compose buffer as `helloworld` (underscores hidden, inner text italic) — the same intraword-underscore symptom the response renderer was already fixed to avoid.

Use `buffer-substring-no-properties` for both preview paths so the context carries plain text and no foreign-mode styling.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -7422,7 +7422,8 @@ Uses AGENT-CWD to shorten file paths where necessary."
                                             (max-preview-lines 5))
                                         (if (= (count-lines char-start char-end) 1)
                                             ;; Same line region? Avoid numbering.
-                                            (buffer-substring-no-properties char-start char-end)
+                                            (agent-shell--buffer-substring-with-faces
+                                             char-start char-end)
                                           (agent-shell--get-numbered-region
                                            :buffer buffer
                                            :from char-start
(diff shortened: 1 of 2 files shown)


xenodium:
Thanks for the PR!

Wouldn't we want to at least keep `face` and `'font-lock-face`? Often gives us syntax highlighting for free.

liaowang11:
> Thanks for the PR!
> 
> Wouldn't we want to at least keep `face` and `'font-lock-face`? Often gives us syntax highlighting for free.

Yes, it's a good idea to keep some faces, I added a custom buffer-substring function to keep both.

xenodium (approved):
Thanks for the PR and follow-up changes.

PR #697, opened 2026-07-04
Fix agent-shell-send-dwim double-inserting context with prefix arg

liaowang11:
With a prefix argument, `agent-shell-send-dwim` called `agent-shell--dwim` to switch to or start a shell and then inserted the DWIM context itself. But `agent-shell--dwim` already carries the context to the chosen shell, so the context landed twice — most visibly when picking an existing session, where the deferred (post `session-selected`) insert and the trailing insert both fired.

Let `agent-shell--dwim` own the send for the prefix branches. To keep this from regressing `C-u` (new shell) in non-viewport mode — the one prefix case that wasn't doubled, because the new-shell branch was the only branch that didn't carry context — make that branch carry context like the others. Extract the shared display-and-insert logic into `agent-shell--display-and-insert-context`.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -972,6 +972,29 @@ With \\[universal-argument] \\[universal-argument] prefix ARG, prompt to pick an
    (t
     (agent-shell--dwim))))
 
+(defun agent-shell--display-and-insert-context (shell-buffer text)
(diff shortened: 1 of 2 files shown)


xenodium (approved):
Thank you for the fix!

PR #698, opened 2026-07-04
fix mouse-copy inside formatted text

jcubic:
This fixes the copy (yank) of the text inside formatted Markdown with a mouse. Not sure if the test makes sense here. Because you can't actually test this. You need a real window and a mouse to reproduce.

I use this code to be able to use the mouse to yank text:

```lisp
(setq x-select-enable-clipboard t)
(setq mouse-drag-copy-region 'non-empty)
(setq x-select-enable-clipboard-manager nil)
;; (setq interprogram-paste-function 'x-cut-buffer-or-selection-value)
(setq interprogram-paste-function 'x-selection-value)
(delete-selection-mode t)

(add-to-list 'yank-excluded-properties 'font)
(add-to-list 'yank-excluded-properties 'font-lock-face)
```

Before the fix, when you copy the text with the mouse and select it (exactly the formatted text or inside), the text is not copied because the selection is empty.

The keyboard copy works fine. Only the mouse selection (yank) is affected.

## Checklist

### DIFF
--- a/agent-shell-markdown.el
+++ b/agent-shell-markdown.el
@@ -385,7 +385,15 @@ body un-fontified."
         (put-text-property (point-min) (point-max)
                            'yank-handler
                            (list (lambda (s)
-                                   (insert (substring-no-properties s))))))
+                                   (insert (substring-no-properties s)))))
+        ;; Mark rendered chars `fontified' so jit-lock never re-runs over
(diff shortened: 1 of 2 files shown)


xenodium (approved):
Thank you for the fix

PR #700, opened 2026-07-05
fix: reverse mouse selection

jcubic:
This is another fix for the mouse selection.

The bug was that the selection only worked from smaller point to bigger; it was not working when you selected the range in reverse order.

The bug was causing an empty string to be returned by the filter and added to the kill-ring when using a mouse. There is also a separate bug in Emacs that the empty string is added to the kill ring.

To reproduce, you need to enable mouse selection. This is my setup:

```elisp
(setq x-select-enable-clipboard t)
(setq kill-read-only-ok t)
(setq mouse-drag-copy-region t)
(setq x-select-enable-clipboard-manager nil)

(if (fboundp 'x-cut-buffer-or-selection-value)
    (setq interprogram-paste-function 'x-cut-buffer-or-selection-value)
  (setq interprogram-paste-function 'x-selection-value))

(delete-selection-mode t)

(add-to-list 'yank-excluded-properties 'font)
(add-to-list 'yank-excluded-properties 'font-lock-face)
```

The same bug is in Emacs filter for the `*Backtrace*` buffer. Reported [here](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=81365)

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -1602,10 +1602,19 @@ See also `agent-shell-confirm-interrupt'."
 
 (defun agent-shell--filter-buffer-substring (start end &optional delete)
   "Return visible text between START and END, stripping hidden markup.
-If DELETE is non-nil, delete the text between START and END."
-  (let ((text "")
-        (pos start))
-    (while (< pos end)
+If DELETE is non-nil, delete the text between START and END.
+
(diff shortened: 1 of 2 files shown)


xenodium (approved):
Thank you for the fix!

Issue #688, opened 2026-07-02
Cursor tool calling not displaying results


aburtsev:
## Checklist

- [x] *I agree to communicate with the author myself* (not AI-generated).
- [x] I've read the README's [Filing issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#filing-issues) section.
- [x] I'm running the latest versions (fill in below).
  - agent-shell version:  0.57.1
  - acp.el version: 0.12.2
  - ACP package (e.g. claude-agent-acp) version: I use built-in cursor ACP `(setq agent-shell-cursor-acp-command '("cursor-agent" "acp"))`
  - Agent CLI (e.g. claude, gemini) version: agent acp
- [ ] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [x] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README).

I'm seeing in `agent-shell-viewport-view-traffic` that tools are called correctly and results are obtained, but they are not being displayed

<img width="939" height="232" alt="Image" src="https://github.com/user-attachments/assets/c1c9c1c4-9071-496d-b888-cadf2268ecf0" />

Here are the ACP traffic details:
<img width="947" height="397" alt="Image" src="https://github.com/user-attachments/assets/74c49036-4b45-49c3-a816-a75a4383accb" />

<img width="936" height="564" alt="Image" src="https://github.com/user-attachments/assets/d66349e0-3da3-44ac-b920-b9440c41a8c0" />

xenodium:
I've migrated to official ACP server in [#386](https://github.com/xenodium/agent-shell/commit/c91b20c29e5512fc7dc340d22c265b9ff377b741). Needs testers. Mind trying there?

aburtsev:
@xenodium 
The issue isn't with the official ACP server but with how Cursor ACP sends the `tool_call_update` data. It sends `rawOutput` instead of `content`. As you can see in the screenshot above. I worked on this issue over the weekend, and I think one approach could be to create a sort of adapter that would allow us to perform normalization on a per-provider basis. With this approach, the `agent-shell` core would remain clean and focused, and we would be able to customize `tool_call` output processing for each provider. Today I plan to create a PR to demonstrate this idea. I already implemented it and tested on my work tasks for the last 2-3 days.

xenodium:
Thanks for this.

[...]

Issue #386, opened 2026-03-09
Update cursor to use official ACP


Lenbok:
Cursor now has official ACP support: https://cursor.com/docs/cli/acp
We should use that rather than the third party acp adapter (which is limited in functionality and seems to have regular auth issues).

## Checklist

- [x] *I agree to communicate with the author myself* (not AI-generated).
- [x] I've read the README's [Filing issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#filing-issues) section.
- [x] I'm running the latest versions (fill in below).
  - agent-shell:
  - acp.el:
  - ACP package (e.g. claude-code-acp):
  - Agent CLI (e.g. claude, gemini):
- [x] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [ ] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README).

Lenbok:
(BTW, I'm taking a look at implementing this, and will give it a run locally)

xenodium:
> (BTW, I'm taking a look at implementing this, and will give it a run locally)

Thanks!

0x6362:
I started on an implementation of this and there's a few gotchas I'm not sure how @xenodium would want to handle and likely merit further discussion -- I admittedly didn't spend a ton of time deep diving into this since I've only made a small, minimal patch that gets things working with authentication and the default Agentic mode. See #393.

Small issues:
- the ACP server didn't seem to be properly updating me on when it made some changes, but I only experimented very briefly and it's likely a PEBKAC error (though I've not seen this same bug with Codex in my more extensive usage). I'll try and develop a minimal repro case and provide logs.
- The official server doesn't expose the modes/models available. I only noticed this issue on the availability after I'd written a patch to support passing the default model ID to the third party Cursor ACP server. I've excluded that code for now as it only creates an error

The big one are the custom Cursor ACP messages such as `_cursor/update_todos`:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "_cursor/update_todos",
  "params": {
    "toolCallId": "tool_$UUID",
    "todos": [
      {
        "id": "1",
        "content": "Fix authenticate method-id to cursor_login for API key case",
        "status": "in_progress"
      },
      {
        "id": "2",
        "content": "Add --api-key command-line support per Cursor docs",
        "status": "pending"
      },
[...]

PR #702, opened 2026-07-06
add cursor toolcall output adapter

aburtsev:

## Checklist

# Render Cursor tool output from ACP `rawOutput`

## Problem

In Cursor ACP sessions, tool_call blocks show status transitions (`pending` → `in_progress` → `completed`) but have **no output body**. Edit tools still work because Cursor populates standard `content` diff blocks for those.

Cursor ACP completes tools with agent-specific data in `rawOutput` while leaving `content` empty:

```json
{
  "sessionUpdate": "tool_call_update",
  "toolCallId": "call_001",
  "status": "completed",
  "rawOutput": { "stdout": "...", "stderr": "", "exitCode": 0 }
}
```

`agent-shell` only read `params.update.content` and rendered it through the existing markdown path. `rawOutput` was ignored, so the UI stayed empty even though the session was otherwise healthy.

This is not an ACP protocol violation. Both `content` and `rawOutput` are optional on [`ToolCallUpdate`](https://agentclientprotocol.com/protocol/v1/schema#param-tool-call-update). Cursor leans on `rawOutput` with agent-specific shapes (`stdout`/`exitCode`, `content`, `totalMatches`, etc.) rather than the `content` text blocks shown in spec examples. Interpreting `rawOutput` is the client's responsibility.

## Solution

Add a small, agent-agnostic hook in core and a Cursor-specific adapter that converts known `rawOutput` shapes into standard ACP content blocks before the existing render/transcript path runs.

## Related issues

- https://github.com/xenodium/agent-shell/issues/386
- https://github.com/xenodium/agent-shell/issues/688

### DIFF
--- a/agent-shell-cursor.el
+++ b/agent-shell-cursor.el
@@ -60,12 +60,12 @@ Only one of API-KEY, AUTH-TOKEN, LOGIN, or NONE should be provided."
   (agent-shell-cursor-make-authentication :none t)
   "Configuration for Cursor authentication.
 
-By default authentication is handled externally: agent-shell sends no
+By default authentication is handled externally: `agent-shell' sends no
 ACP authenticate request and relies on an existing Cursor login (run
 `agent login' once outside Emacs).  This matches how Cursor was used
 before and needs no configuration.
 
-Optionally, configure agent-shell to manage authentication instead.
+Optionally, configure `agent-shell' to manage authentication instead.
 
 For no authentication, handled externally (default):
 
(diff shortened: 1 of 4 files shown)


xenodium:
Thanks for the PR!

Now's a good opportunity for us figure out how to handle custom ACP logic more generically.

More specifically, I'm thinking what we add to agent configs are adapters at a coarser granularity. This should scale across agents with fewer integration hooks and also enable us to establish a repeatable pattern.

1. In terms of granularity, I'm thinking the adapters should operate at request or notification level. For notifications, receive a notification and return either modified notification or nil. This enables generic logic to continue working as is without introducing more edge cases in core. So for the content instance (this PR), we add an adapter to Cursor config :notification-adapter (receives acp-notification, returns notification with the modifications it needs).

The notification is a transient, single-owner object, so the adapter can just override the fields it cares about in place, no copying required.

```emacs-lisp
(cl-defun agent-shell--adapt-notification (&key state acp-notification)
  (if-let* ((adapter (map-nested-elt
                      state '(:agent-config :notification-adapter)))
            ((functionp adapter)))
      (funcall adapter :acp-notification acp-notification)
    acp-notification))
...
(acp-subscribe-to-notifications
 :client (map-elt state :client)
 :on-notification (lambda (acp-notification)
                    (agent-shell--on-notification
                     :state state
                     :acp-notification
                     (agent-shell--adapt-notification
                      :state state
                      :acp-notification acp-notification))))
...


2. This is now out of scope for this PR, but we can do a similar thing for incoming requests, adding a :request-handler. The core logic checks for a config's request handler and offers the request to it. A custom agent handler returning a response takes precedence over the core handlers.

Hope this makes sense!

aburtsev:
@xenodium Thank you for the detailed explanation! It makes sense, and I completely agree with your suggestion. I'll work on this and update the PR accordingly.

aburtsev:
@xenodium 
I have updated the PR according to your recommendations.  
I focused only on adding `notification-adapter` and kept `request-handler` out of scope. I don't currently have a use-case for the `request-handler`, but I can add it in a future PR if you think it would be beneficial right now. Let me know!

xenodium (approved):
Thank you for reworking. Looks great. The changes to agent-shell.el are now minimal.

PR #713, opened 2026-07-10
Align embedded-image rendering with link rendering

alberti42:
## Short description

Embedded images (`alt`) are rendered by *replacing* the markdown markup with the alt text and painting the image on top. Links (`[title](url)`) work the same way. But the image renderer had quietly drifted away from the link renderer in two respects, and each drift caused a user-visible bug. This PR closes both gaps.

Both fixes are small and confined to one function, `agent-shell-markdown--replace-images`; each comes with a regression test that fails before the change and passes after.

### Reproducer

When the agent's reply contained an image on its own line, like:

```
paragraph before

![svg graphics](~/.cache/arXiv/2603.15561v1/figure_1_final.svg)

paragraph after
```

the image rendered fine, but **everything after it vanished** — `paragraph after` (and any further paragraphs) was not shown.

Clues that helped localize it:

- A plain link `[svg graphics](…)` (no leading `!`) rendered everything correctly.
- A bare image path on its own line (no markdown syntax) rendered everything correctly.
- Replaying the full conversation later (`agent-shell-session-restore-verbosity 'full`) showed the missing text.

So it only happened with the `alt` form, and only the *first* time the reply was streamed in — never on a full re-render.

### Why it happened

Agent replies arrive in small chunks and are appended to the buffer as they stream.  To know where each new chunk goes, the shell marks the whole message body with a single, **continuous** "this is body text" tag and locates the body by that continuous run.

Most renderers keep the underlying text (and its tags) and only *paint* styling on top. The image renderer was the exception: it **deleted** the `alt` markup and **re-inserted** the alt text as a brand-new, untagged string. That untagged placeholder punched a hole in the continuous "body" run, splitting it into a piece before the image and a piece after it.

On the next streamed chunk, the code that locates the body got confused by the gap and latched onto the piece *after* the image — which begins with the blank line following the image, already marked "invisible" (that is how trailing blank lines are normally hidden). The visibility check saw an invisible first character, concluded the whole body was collapsed, and every line appended from then on was hidden too. The text was still in the buffer, just invisible — which is exactly why a fresh full re-render displayed it correctly.

This also explains the working cases: links already capture their title *with* its tags, and bare image paths never delete anything.

### The fix

Make the placeholder carry the surrounding text's properties, exactly as the link renderer already does, so the body run stays continuous. The empty-alt case and the remote-image fallback inherit the surrounding tags the same way.

### Reproducer

Copying a rendered image with `agent-shell-copy-as-markdown` produced the bare alt text `svg graphics` instead of restoring the original `svg graphics`.

### Why it happened

`agent-shell-copy-as-markdown` reconstructs the original markdown from an `agent-shell-markdown-source` text property that each markup-stripping renderer is expected to leave behind (bold, italics, headings, links, code blocks, tables all do this). The image renderer never set it, so there was nothing to reconstruct from and copying fell back to the visible placeholder text.

### The fix

Stash the original `alt` markup on `agent-shell-markdown-source` when rendering the image — the same way the link renderer does — for both the inline-image and the remote-fallback paths.

## Testing

- `agent-shell-markdown-image-render-preserves-surrounding-properties` — guards Bug 1 (the property run is not broken by the placeholder).
- `agent-shell-markdown-image-reconstructs-to-source` and `agent-shell-markdown-remote-image-fallback-reconstructs-to-source` — guard Bug 2 (copy round-trips back to the original markup).
- Nothing that already worked changes: the image still renders in place, and a plain (non-markdown) copy still yields the alt text.
- Full test suite passes. The only failure is the pre-existing, timing-based flaky `agent-shell-idle-event-rearms-on-new-trigger-test`, which is unrelated and documented as flaky in batch.

### DIFF
--- a/agent-shell-markdown.el
+++ b/agent-shell-markdown.el
@@ -773,6 +773,14 @@ For example, the buffer \"see ![logo](logo.png)\" becomes
                                                   (match-end 1))))
                  (url (buffer-substring-no-properties
                        (match-beginning 2) (match-end 2)))
+                 ;; Stash the original `![alt](url)' markup so
+                 ;; `agent-shell-copy-as-markdown' round-trips the image back to
+                 ;; source rather than yielding the bare alt placeholder (mirrors
+                 ;; `--replace-links').  Guarded so a re-render doesn't overwrite
+                 ;; an already-captured source.
+                 (source (unless (get-text-property markup-start
+                                                    'agent-shell-markdown-source)
+                           (agent-shell-markdown-reconstruct markup-start markup-end)))
                  (path (agent-shell-markdown--resolve-image-url
                         url image-cache-directory)))
             (cond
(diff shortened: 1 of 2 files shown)


xenodium (approved):
Thank you for the fix!

PR #715, opened 2026-07-11
Fix table column wrapping for CJK text

ktakahashi74:
## Summary

Tables containing Japanese (or other CJK) text never shrink to fit the window: `agent-shell-markdown-table-wrap-columns` has no effect and the table overflows.

Cause: `agent-shell-markdown--table-longest-word` splits cell content on whitespace only. CJK sentences contain no spaces.

Fix:

- In the minimum-width calculation, treat line-breakable characters (category `|`: CJK ideographs, kana, Hangul, etc.) as individually breakable, each contributing its own `char-width`. Emoji sequences (ZWJ, skin-tone modifiers) don't have category `|`, so they stay unbreakable as before.
- Teach `agent-shell-markdown--table-wrap-text` to break after line-breakable characters in addition to whitespace, so an ASCII word embedded in CJK text (e.g. "API") is not split mid-word. A zero-width follower (combining char, variation selector, ZWJ) blocks the break.

Wide characters contribute `char-width` directly instead of being pixel-measured one by one, since tables re-render on every streamed chunk.

## Before / after

before:
before

after:
after

## Testing

- Added three tests to `tests/agent-shell-markdown-tests.el`; all 113 tests pass.
- Verified interactively in my own agent-shell sessions with Japanese tables.

## Checklist

### DIFF
--- a/agent-shell-markdown.el
+++ b/agent-shell-markdown.el
@@ -1510,19 +1510,39 @@ pipes between rows."
     (string-width str)))
 
 (cl-defun agent-shell-markdown--table-longest-word (&key str window)
-  "Return display width of longest word in STR.
-Uses `agent-shell-markdown--table-display-width' so non-ASCII words
-get accurate measurement when WINDOW is given."
+  "Return display width of the longest unbreakable unit in STR.
+
+Runs of non-breakable characters form unbreakable words, measured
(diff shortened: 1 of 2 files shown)


xenodium (approved):
ありがとう ございます!

PR #717, opened 2026-07-12
Make the markdown render context available via a public function

alberti42:
# Make the markdown render context available via a public function

Follow-up to #674 (7ee6ff9).

## Background

In today's agent-shell, external packages can plug into the markdown renderer through the `agent-shell-markdown-render-functions` hook. When the agent's response streams into the buffer, each registered function is handed a small `context` describing the text it's about to work on:

- where the fenced code blocks are and
- where the inline `` `code` `` spans are,

so the function knows which regions to leave alone (e.g., the package `agent-shell-math-renderer` does not want to render math, say, inside a code block).

That works well for streamed output. But the same hook never fires for **static** text, which is inserted all at once rather than streamed. Static text includes the case of submitted user prompts.
 
## The problem

To render such a region, an external package needs to build the *same* `context` for itself. But in today's agent-shell there is no public way to get it. The code that assembles the context lives inline inside `agent-shell-markdown-replace-markup`, and it is a mix of two things:

- Calls to several private (`--`-prefixed) functions, such as `agent-shell-markdown--source-blocks`, `agent-shell-markdown--inline-code-ranges` and `agent-shell-markdown--make-markers`. These are at least callable, even if private and unsupported.
- Plain inline glue that is not exposed as *any* function, not even a private one. It projects the fenced blocks into sorted ranges, feeds those ranges into the inline-code scan (so backticks that sit inside a code block are ignored), and wires the pieces together in a specific order.

So reproducing the context is not simply "call a few private functions."  An external package would have to reach into some of the private functions and copy the inline glue verbatim, since that glue has no callable form to begin with. Copied code that shadows internals it cannot even name is bound to quietly drift out of sync as the package evolves. This would be a fragile workaround at best.

## What this PR does

It refactors the code without changing any of its current behaviors: it extracts that assembly into a single public function, `agent-shell-markdown-context`, which runs on the current (narrowed) region and returns the very alist the render hook already documents:

```elisp
((:source-blocks . …)        ; the fenced code blocks in the region
 (:inline-code-ranges . …))  ; the inline `code` spans in the region
```

Crucially, `agent-shell-markdown-replace-markup` now builds *its* context by calling this same function, instead of assembling it inline. So the streaming path and any external caller share **one** builder and cannot drift apart — which is the whole point of the change.

Two small internal changes come with it:

- The step that projects the fenced-block descriptors into plain sorted `(start . end)` ranges is now a private helper, `agent-shell-markdown--source-block-ranges`. This projection appears only once in today's code, but after the split it is needed in two places — inside the new function and in `replace-markup` — so factoring it out keeps a single copy instead of pasting the same lines twice.
- `agent-shell-markdown--run-render-functions` now receives the already-built context rather than rebuilding it, so the alist is constructed in exactly one place.

For the public `context` alist, I chose to provide the same two keys the `render-functions` hook is already documented to pass, so it stays faithful to that existing contract used in streamed text.

It is worth repeating that this PR introduces no behavior changes for existing users: same rendering, same hook, same context shape — there's simply now a supported way to obtain that context for regions that didn't come through the streaming path.

## Testing

- Byte-compiles cleanly (no new warnings).
- The full markdown test suite passes (116/116).
- Sanity-checked directly: on a buffer mixing an inline `` `x` ``, a ` ```math ` fenced block, and another inline `` `y` ``, `agent-shell-markdown-context` returns the two expected keys, one source block (language `"math"`), and the two inline ranges (`x`, `y`) — correctly ignoring the backticks that live inside the fenced block.

### DIFF
--- a/agent-shell-markdown.el
+++ b/agent-shell-markdown.el
@@ -313,6 +313,7 @@ body un-fontified."
                                 '(agent-shell-markdown-watermark nil))))
     (let ((watermark (agent-shell-markdown--watermark-start))
           (external-results)
+          (context)
           (source-blocks)
           (source-ranges)
           (rendered-ranges)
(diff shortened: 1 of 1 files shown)


xenodium (approved):
Thank you for the change

PR #612, opened 2026-06-01
Add `agent-shell-session-prompt-exclude-choices` defcustom

nhojb:
Adds `agent-shell-session-prompt-exclude-choices` defcustom to support filtering of the prompt choices when starting a new shell.

For example you can exclude the downloads and temp options:

```
(setq agent-shell-session-prompt-exclude-choices '(:downloads-shell :temp-shell))
```

## Checklist

- [x] *I agree to communicate (PR description and comments) with the author myself* (not AI-generated).
- [x] *I've reviewed all code in PR myself and will vouch for its quality*.
- [x] I've read and followed the [Contributing](https://github.com/xenodium/agent-shell/blob/main/CONTRIBUTING.org) guidelines.
- [ ] I've filed a feature request/discussion for a new feature.
- [ ] I'm making visual changes, so I'm including screenshots so you can view and discuss.
- [x] I've added tests where applicable.
- [x] I've updated documentation where necessary.
- [x] I've run `M-x checkdoc` and `M-x byte-compile-file`.

PR #723, opened 2026-07-16
Add agent-shell-extra-shell-choices to control the start-shell prompt

catern:
It controls which extra choices ("New Downloads shell", "New temp shell", "Switch to shell buffer") are offered when prompting to start a shell.

The main motivation is disabling all those other choices, leaving only "New shell" and existing sessions.  And since a prompt for only one choice is not useful, when there are no existing sessions and the other choices are all disabled, the prompt is just skipped in favor of immediately starting the new shell.

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -763,6 +763,23 @@ Available values:
          (set-default sym value))
   :group 'agent-shell)
 
+(defcustom agent-shell-extra-shell-choices '(downloads-shell temp-shell other-shell)
+  "Extra choices offered when prompting to start a shell.
+
+Available choices:
+
+  `downloads-shell': Start a new shell in ~/Downloads.
+  `temp-shell': Start a new shell in a temporary directory.
+  `other-shell': Switch to an existing shell buffer.
+
+When starting a new shell is the only remaining choice (no extra
+choices and no existing sessions to resume), the prompt is skipped
+and a new shell is started."
+  :type '(set (const :tag "New Downloads shell" downloads-shell)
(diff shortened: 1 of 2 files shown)


xenodium:
@nhojb @catern thank you both for the PRs!

While the customisation hasn't received +1's from other users, we now have 2 PRs 🙂 #612 and #723 

We have a couple of approaches for customising options (remove vs keep), so I figured I should offer yet a third approach 😅 Sorry! 

Having said that, I think https://github.com/xenodium/agent-shell/commit/62eddc37e5599001f3f4ca2249533cc32d5f473e should be versatile enough to offer both approaches (uses a function setting), but also enables users to reorder or even rename options. While it's not presented as simply as you both achieved, I consider the feature more of a deeper (advanced) customisation, so I'm hoping it justifies the third alternative requiring a little elisp.

Either way, it should hopefully enable what you were both after.

xenodium:
@catern your PR still offers the additional benefit of auto-selecting new when its the only option. Fancy rebasing? Alternatively, I'm happy to follow-up and generalise it (any single option is automatically chosen).

nhojb:
@xenodium LGTM. Nice to finally have this customization in `main` 😄

xenodium:
> Alternatively, I'm happy to follow-up and generalise it (any single option is automatically chosen).

@catern added in https://github.com/xenodium/agent-shell/commit/6da58df2a3dc83260f22657c9663bb97dae9a025 Gonna go ahead and close the PR, but please shout if I missed functionality it included.

PR #739, opened 2026-07-29
Add opencode model variant customization

nhojb:
Opencode does not currently support setting a persistent model variant via the opencode config json - or at least it does not work when running via acp. So this PR adds:

1. Support for setting the `default-thought-level-id` via `agent-shell-make-agent-config`
2. A new `agent-shell-opencode-default-model-variant` defcustom

Note: if the variant is not supported by the current model, then the user will see an error when loading the session

Screenshot 2026-07-29 at 14 07 13

Screenshot 2026-07-29 at 13 52 40

## Checklist

### DIFF
--- a/README.org
+++ b/README.org
@@ -1051,6 +1051,7 @@ always go to Evil modes if you need to with ~C-z~).
 | agent-shell-opencode-acp-command              | Command and parameters for the OpenCode client.                                 |
 | agent-shell-opencode-authentication           | Configuration for OpenCode authentication.                                      |
 | agent-shell-opencode-default-model-id         | Default OpenCode model ID.                                                      |
+| agent-shell-opencode-default-model-variant    | Default OpenCode model variant.                                                 |
 | agent-shell-opencode-default-session-mode-id  | Default OpenCode session mode ID.                                               |
 | agent-shell-opencode-environment              | Environment variables for the OpenCode client.                                  |
 | agent-shell-path-resolver-function            | Function for resolving remote paths on the local file-system, and vice versa.   |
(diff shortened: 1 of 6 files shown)


xenodium:
Thanks for the PR!

I'm wondering if you'd be keen to steer the PR towards a more generic mechanism? Instead of adding `agent-shell-opencode-default-model-variant`, we add something like:

```emacs-lisp
(defcustom agent-shell-opencode-default-config-options
  '(("model"         . "anthropic/claude-opus-4-5")
    ("thought_level" . "high")
    ("mode"          . "plan"))
  "Default OpenCode config options, applied at session start.
Keys are config-option categories advertised by the agent; see
the \"Available config options\" section in your shell."
  :type '(alist :key-type string :value-type string))
```

I know we haven't done this before in project, but it would enable us to set precedence for moving towards a more scalable mechanism for users to set whatever option is supported by their agent without requiring additional agent-shell changes.

To support this path, I've just submitted https://github.com/xenodium/agent-shell/commit/95c700ad2c89be53b867dbd84d249ceb6b6d7ecf which enables users to know exactly what is and isn't supported by their agents.

image

nhojb:
> I'm wondering if you'd be keen to steer the PR towards a more generic mechanism?

Sounds good to me. I will also mark the old `agent-shell-opencode-default-model-id` as deprecated and fallback if the new config does not exist.

nhojb:
Done. Apologies for the delay.

You can now set opencode options like this:

```
(setq agent-shell-opencode-default-config-options
   '(("model" . "openrouter/moonshotai/kimi-k3")
     ("effort" . "high")))
```

Note: I did not remove `default-model-id` or `default-session-id` from `agent-shell-make-agent-config` even though they can now be set via `default-config-options`. That could be done in a follow-up, but touches most of the other agents.

xenodium:
> Done. Apologies for the delay.

All good. Thanks for the additional changes.

Merged and followed up with https://github.com/xenodium/agent-shell/commit/9ce3c06a26116878119d3baa0a93407f3955c806

PR #740, opened 2026-07-30
Add a `when-active` option to `agent-shell-activity-group-expand-by-default`

nhojb:
I've been using `agent-shell-activity-group-expand-by-default t` for the new activity groups so that I can see progress while the agent is working.

Then I realised that I only care about activity in the currently active group. The others can close, keeping the session clean.

So this PR makes `agent-shell-activity-group-expand-by-default` a choice with the following options:

- `never`: backwards compatible with `nil`
- `always`: backwards compatible with `t`
- `when-active`: The group the agent is currently working in shows its members, and collapses once the agent moves on to a new group or the turn ends

I've been testing `when-active` in various sessions and it seems to work well - keeping me informed of the agent's current progress, but keeping the overall session clean.

Screenshot 2026-07-30 at 16 35 12

## Checklist

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -139,17 +139,53 @@ When non-nil, tool use sections are expanded."
   :type 'boolean
   :group 'agent-shell)
 
-(defcustom agent-shell-activity-group-expand-by-default nil
-  "Whether an activity group header is expanded by default.
+(defcustom agent-shell-activity-group-expand-by-default 'never
+  "When activity group headers should be expanded.
 
 An activity group is a run of consecutive agent actions (tool calls,
-and eventually thoughts) rendered under one collapsible header.  When
-non-nil (the default), the group shows its members.  When nil, it starts
-collapsed, showing only the header with its aggregated status and
-completed/total count.  Individual members still follow
+and eventually thoughts) rendered under one collapsible header.
+
+  `never'       Groups start collapsed, showing only the header with its
+                aggregated status and completed/total count.
+  `always'      Groups show their members.
+  `when-active' The group the agent is currently working in shows its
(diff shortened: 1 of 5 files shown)


agent-shell.el
```diff
@@ -139,18 +139,54 @@ When non-nil, tool use sections are expanded."
   :type 'boolean
   :group 'agent-shell)
 
-(defcustom agent-shell-activity-group-expand-by-default nil
-  "Whether an activity group header is expanded by default.
+(defcustom agent-shell-activity-group-expand-by-default 'never
```
xenodium: nit: Can we defer adding the new values since we're effectively supporting more values even if we're deprecating some?

Also, maybe consider naming the new value either `latest`, `current`, or `active`. I'm leaning towards `latest` myself.

```emacs-lisp
(defcustom agent-shell-activity-group-expand-by-default nil
  "..."
  :type '(choice (const :tag "Never (collapsed)" nil)
                 (const :tag "Always (expanded)" t)
                 (const :tag "While the agent is working in it" latest))
  :group 'agent-shell)
```

xenodium (changes requested):
Thanks for the PR and sorry for the delay reviewing. I've been getting requests around grouping and wanted to wait a bit for grouping itself to settle as a landed feature but also to see where the requests were going. Want to avoid adding unpopular features prematurely if possible (along with unnecessary complexity).

The PR looks good to me. Mainly some nits about the defcustom itself since that becomes part of the maintained surface.

nhojb:
Thanks for the review. I've updated the defcustom.

xenodium (approved):
Thanks for the follow-up changes.

PR #742, opened 2026-07-30
Fix executable-find on remote host

CeleritasCelery:
This change will not impact the behavior of local agent-shell at all, but it will fix a lingering issue where agent-shell tramp will not work if you don't have the same executable on both the remote host and the local host. Enabling the remote flag means it will find the executable on the host you are currently working on instead of only the local host.

See junyi-hou/agent-shell-tramp#5

## Checklist


xenodium (approved):

xenodium:
Thanks!

PR #743, opened 2026-07-31
Preserve "claimed" regions when rendering bold / italic / strikethrough

alberti42:
## The PR in one paragraph

When an agent's message puts **bold** (or *italic*, or ~~strikethrough~~) around something a plugin like `agent-shell-math-renderer` has already rendered — for example a math formula inside `**...**` — that rendered content is currently lost; this PR fixes that, and along the way removes some duplicated code by refactoring it.

## The problem (in simple terms)

agent-shell lets other packages hook in and render their own content inside a message; an example is the LaTeX math renderer [agent-shell-math-renderer](https://github.com/alberti42/agent-shell-math-renderer) that turns `\(E=mc^2\)` into an actual typeset equation image.

To render this safely, such plugins "claim" the little stretch of text they are responsible for and mark it as off-limits, so agent-shell's own formatting steps leave it alone. In short, the contract is: *once a region is claimed, don't disturb it.*

There was a gap in that contract. agent-shell formats bold text by finding a `**...**` span, **deleting the whole thing, and retyping the inner text** without the `**` markers. That works fine for ordinary text. But if the bold happens to *wrap around* a region a plugin already claimed, the delete-and-retype throws that claimed content away:

- the plugin's "bookmark" pointing at the content is left pointing at nothing, and
- any image/overlay the plugin had placed there is dropped.

So a formula written as `**this matters: \(E=mc^2\)**` would simply fail to render — the surrounding text became bold, but the equation vanished. And because the region was still marked "already handled," agent-shell never tried again, so it stayed broken.

This is easy to trigger in practice: agents very often emphasize a sentence that contains an inline formula.

## Why it is worth fixing (beyond the refactor)

This isn't only about one math plugin. It's about **agent-shell keeping its promise** to any package that extends it. The whole point of the "claim a region and it won't be touched" mechanism is that plugins can rely on it. Today that promise quietly breaks in the common case where emphasis surrounds the claimed content.

Fixing it makes the extension mechanism something plugins can actually rely on.

## The fix

Bold, italic, and strikethrough all did the same delete-and-retype dance in three near-identical blocks of code. This PR:

1. **Pulls that shared logic into one small helper**, so the three passes no longer repeat themselves.
2. **Changes the approach** so it only removes the `**` (or `*` / `~~`) markers and **leaves the inner text exactly where it is**. Nothing inside the span is deleted, so anything a plugin anchored there survives untouched.

The visible result is unchanged for normal text — bold still looks bold — but a claimed region inside emphasis now keeps its bookmarks and its rendered image.

## How it was verified

- Reproduced the original failure (a formula inside `**...**` losing its anchor), then confirmed the fix keeps the anchor valid and the content in place.
- All existing markdown tests still pass.
- Added a focused regression test that wraps a claimed/frozen region in bold and checks that its bookmarks, its overlay, and its "off-limits" marker all survive.

## Risk

Low. The behavior for ordinary emphasized text is identical (same face, same copy-as-markdown output); the only difference is that the inner characters are kept rather than deleted and retyped. The change is confined to the three emphasis passes plus the new shared helper.

### DIFF
--- a/agent-shell-markdown.el
+++ b/agent-shell-markdown.el
@@ -514,6 +514,42 @@ this returns `(((:watermark . 1200)))'."
                         nil))
     (nreverse results)))
 
+(defun agent-shell-markdown--emphasize-span (markup-start markup-end
+                                                          inner-start inner-end
+                                                          face)
+  "Strip emphasis delimiters around INNER-START..INNER-END and face the text.
+
+MARKUP-START..MARKUP-END spans the whole construct including its
+delimiters (e.g. `**X**'); INNER-START..INNER-END is the inner text
+\(`X').  Only the leading and trailing delimiters are deleted -- the
+inner text is left in place rather than deleted and re-inserted -- so
+any overlays, markers, or `agent-shell-markdown-frozen' regions an
+earlier pass or an external `agent-shell-markdown-render-functions'
+renderer placed inside survive.  (Deleting and re-inserting collapsed
+external markers to a zero-length span and dropped overlays, so a
+renderer that anchors an async result inside, say, a bold span -- e.g.
+math inside `**...**' -- lost it.)
+
+FACE is layered onto the remaining text with `add-face-text-property'
+\(so it composes with faces from earlier passes), and the original
+markdown is stashed on `agent-shell-markdown-source' (for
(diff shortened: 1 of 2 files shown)


xenodium (approved):
Thanks for the improvement!

PR #746, opened 2026-08-02
Avoid repeated system-sleep load attempts

liaowang11:
## Summary

This PR fixes repeated attempts to load the optional `system-sleep` library
during Agent Shell event dispatch.

## Findings

Profiling a live Agent Shell session on Emacs 30.2.50 found that
`agent-shell--inhibit-sleep` called `(require 'system-sleep nil t)` before
checking `agent-shell-inhibit-system-sleep`. Because `system-sleep` is not
available on Emacs 30, every busy event retried the failed load. The profile
attributed 15.4% of inclusive CPU samples and 62.8% of sampled allocation
bytes to `require` (approximately 325 MB of 517,938,025 sampled bytes).

## Changes

- Check `agent-shell-inhibit-system-sleep` before looking for or loading the
  optional dependency.
- Add `agent-shell--system-sleep-available-p`, which memoizes one unavailable
  library-load attempt while preserving the existing behavior when the API is
  already available.
- Add regression coverage proving that disabled sleep inhibition makes no load
  attempt and that an unavailable backend is attempted only once.
- Preserve the existing busy/blocked/ready status transitions, terminal-event
  release, and single-token behavior.

## Verification

- Five focused system-sleep ERT tests pass.
- `M-x checkdoc` passes for `agent-shell.el`.
- `M-x byte-compile-file` passes for `agent-shell.el`.
- `git diff --check` passes.

This PR addresses the repeated `system-sleep` load finding only. The separate
collapsed-group streaming-update finding is intentionally not included.

## Checklist


xenodium:
Thank you for the improvement.

PR #748, opened 2026-08-03
Preserve properties on escaped Markdown punctuation

Scott-Guest:
## Problem

Rendering backslash-escaped Markdown punctuation replaces the escaped pair with a placeholder. The placeholder drops caller-owned text properties. Consequently, when escaped punctuation terminates a fragment label, the rendered character falls outside the fragment's state, section, and read-only ranges.

## Change

Carry caller-owned properties from the escaped character onto the placeholder using the renderer's existing property filter. Renderer-owned properties remain controlled by the renderer.

## Verification

- Added a regression test that renders `den\.` with UI properties on the period. Before this change the rendered period loses `agent-shell-ui-state`; after this change it retains the state, section, and read-only properties.
- Ran the full ERT suite
## Checklist


xenodium (approved):
Sorry for this. Thank you for fixing!

Issue #751, opened 2026-08-03
agent-shell-ui--group-children can loop forever (100% CPU, unbounded GC) while collapsing a tool-call group


eddof13:
### Summary

A `session/update` notification for a Bash tool call wedged Emacs at 100% CPU indefinitely (fully unresponsive, growing heap). I traced it to a non-terminating loop in `agent-shell-ui--group-children`, reached while auto-collapsing a tool-call group. The loop `push`es on every iteration and only makes progress by advancing to a member block's `:end`, with no guard that `:end` actually moves forward — so a non-advancing block spins forever and conses without bound (heap reached ~1.2 GB before I intervened).

### Environment

- agent-shell 0.66.1 (`Package-Version: 20260803.716`, revision `53ed8216e52e`)
- acp 0.13.1, Emacs 30.2 (macOS 26.6, aarch64)
- Buffer: a `Claude Agent @ web` session

### Symptom

Emacs pinned at 100% CPU, `R` (running) state, unresponsive to `C-g`. `sample`(1) showed 100% of time under one path:

```
command_loop_1 → read_key_sequence → read_char
  → wait_reading_process_output → timer_check
    → Ffuncall → exec_byte_code ...
```

~65% of that was `garbage_collect` (`mark_char_table` / `process_mark_stack`); the rest was text-property scanning (`Fnext/previous_single_property_change`, `lookup_char_property`, `find_interval`, interval-tree balancing) plus continuous `Fcons` — i.e. an allocation-bound loop, not a one-shot large scan.

### Lisp backtrace (via `kill -USR2`)

```
map-elt(((:start . 16411) (:end . 16430)) :end)
agent-shell-ui--group-children(:group-qualified-id "1-activity-1")
agent-shell-ui--group-child-region(:group-qualified-id "1-activity-1")
agent-shell-ui--set-group-collapsed("1-activity-1" t)
agent-shell-ui-update-fragment(... :label-left #("[…][run]" ...))
agent-shell--update-fragment(...)
agent-shell--on-notification(... :buffer #<buffer Claude Agent @ web> ...)
...
acp--route-incoming-message(:message ((method . "session/update")
  (params (update (_meta (claudeCode (toolName . "Bash"))) (toolCall ...)))))
timer-event-handler([t 27248 59110 471883 nil #f(compiled-function () ...) nil nil 0 nil])
```

[...]

PR #752, opened 2026-08-03
Guard group member walk against non-advancing block range #751

hamza-m-masood:
Fixes the non-terminating loop reported in #751.

`agent-shell-ui--group-children` walks group members by jumping to each member's `:end`, and nothing requires that end to move forward:

```elisp
(let ((block (agent-shell-ui--block-range :position (point))))
  (push (list ...) children)
  (setq pos (map-elt block :end)))   ; only progress signal
```

A block range ending at or behind `pos` re-examines the same member forever, consing on every pass. #751 has a `sample`(1) profile of that state: ~65% of the time in `garbage_collect`, heap climbing until Emacs is unresponsive. A `nil` `block` is also unguarded and reaches `goto-char` as `nil` on the next iteration.

This adds a strict forward-progress requirement, so a stalled range ends the walk instead of spinning. That covers both callers, since `--group-child-region` and `--group-insertion-point` share `--group-children`.

I have not identified what produces the non-advancing range. I tried the duplicate-`:qualified-id` theory from #751 — planting an earlier run with a member's id — and it did **not** reproduce: `--block-range` resolved correctly and the walk terminated. So this is a defensive invariant on the loop, not a fix for the underlying range calculation, and the real trigger is still open in #751.

Trade-off worth naming: when a range does stall, enumeration now stops early rather than looping, so a folded group could leave a trailing member visible and `--group-insertion-point` could return a slightly early position. Both seem clearly preferable to the hang, but it is a behaviour change in that degenerate case.

### Test

`agent-shell-ui-group-children-stops-when-block-range-stalls-test` stubs `--block-range` to report a stalled range for members while letting the header resolve normally. Verified it fails without the production change (returns `("ns-t1")`, i.e. the walk accepts the stalled member) and passes with it. It asserts the guard rather than reproducing the hang, so it fails fast instead of stalling CI.

Full suite: 426 tests, 0 unexpected (1 pre-existing skip). `byte-compile-file` diffed against baseline — no new warnings. `checkdoc` clean on both changed files.

## Checklist

---

Disclosure on the first two boxes, deliberately left unchecked: this description and the code were drafted with AI assistance (Claude Code), so I cannot honestly tick them as written. I'm reviewing the diff myself and will follow up in my own words to confirm I vouch for it. Flagging rather than quietly ticking, since CONTRIBUTING.org is explicit about this. Happy to close if you'd rather just take the patch from #751 directly.

### DIFF
--- a/agent-shell-ui.el
+++ b/agent-shell-ui.el
@@ -605,12 +605,19 @@ equal to GROUP-QUALIFIED-ID; the run stops at the first non-member."
             (let ((state (get-text-property (point) 'agent-shell-ui-state)))
               (unless (and state (equal (map-elt state :group-id) group-qualified-id))
                 (throw 'done nil))
-              (let ((block (agent-shell-ui--block-range :position (point))))
+              (let* ((block (agent-shell-ui--block-range :position (point)))
+                     (end (map-elt block :end)))
+                ;; POS advances only to END, with nothing requiring END to move
+                ;; forward.  A range ending at or behind POS re-examines the
+                ;; same member forever (consing on every pass), and a nil BLOCK
+                ;; reaches `goto-char' as nil.  Stop instead.
+                (unless (and end (> end pos))
+                  (throw 'done nil))
                 (push (list (cons :qualified-id (map-elt state :qualified-id))
                             (cons :start (map-elt block :start))
-                            (cons :end (map-elt block :end)))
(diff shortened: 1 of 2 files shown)


xenodium (approved):

xenodium:
Thanks for the PR

Issue #755, opened 2026-08-05
Make file link clicks configurable (open in another window instead of taking over the shell)


Gleek:
Hello Alvaro,

When I click a file link in a rendered markdown response (or hit RET on it), the file opens in the current window and replaces the shell buffer. I lose sight of the conversation I was reading, and have to switch back to get context on why I opened that file in the first place. Most of the time I'd rather have the file appear beside the shell so I can read both.

We seem to be calling find-file directly, which replaces the same window and I was wondering if we could have an option to display the window based on the function. Similar to how we have for agent-shell buffers (agent-shell-display-action).

I noticed there are several find-file call sites (for images, links, `@file` mentions, agent-shell-diff-visit-file, etc). I suppose all of them could follow this variable.
`agent-shell-diff-visit-file`, though might require a different setting because that's an interactive command and people might expect the same window to be replaced

What do you think?

## Checklist

- [x] *I agree to communicate with the author myself* (not AI-generated).
- [x] I've read the README's [Filing issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#filing-issues) section.
- [ ] I'm running the latest versions (fill in below).
  - agent-shell version: 0.63.6 ( I don't see any code in newer version solving this)
  - acp.el version: 0.13.1
  - ACP package (e.g. claude-agent-acp) version: NA
  - Agent CLI (e.g. claude, gemini) version: NA
- [ ] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [ ] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README).

Gleek:
An inbuilt alternate is to use other-window-prefix (C-x 4 4) before pressing return on the link. But IMO, adding native support for this improves the usability further.  WDYT @xenodium ?

xenodium:
Sorry for the delay. I agree. Been meaning to reply earlier. I'm working on related changes at the moment, which will make all of this more manageable. I'll update once I've got these pushed.

xenodium:
Changes are in as of v0.73.1. Use `agent-shell-file-display-action`.

PR #756, opened 2026-08-06
Preserve point during table rendering

Lenbok:
I have been trying to use `embark-act` on items in the agent-shell session buffer (for example, when the agent refers to a file, jump to that text and `C-. RET` to open the file). I've found that if the session contains markdown tables, when I return to the agent shell buffer (e.g. by exiting out of the embark menu), point is no longer where I activated embark from - it's at the end of the last markdown table in the buffer. 

I got an agent to inspect the problem, and it's assessment is that a window resize hook is fired when the embark menu window opens, which causes `agent-shell-markdown-rerender-tables` to call `agent-shell-markdown--render-table` on the tables - this latter function deletes and inserts the table (which moves point), but there's no `save-excursion` protecting the original point. This PR adds the `save-excursion` which fixes the issue for me.

While I first noticed it when using embark, there are probably other ways to trigger this (e.g. if one were to make link clicking open in a new window, as requested in #755, that would likely also trigger a window resize)

## Checklist

### DIFF
--- a/agent-shell-markdown.el
+++ b/agent-shell-markdown.el
@@ -2756,13 +2756,16 @@ is safe to call on display and on resize."
         (let ((inhibit-read-only t)
               (buffer-undo-list t)
               (modified (buffer-modified-p)))
-          (dolist (region (nreverse regions))
-            (agent-shell-markdown--render-table
-             (list (cons :source (nth 2 region))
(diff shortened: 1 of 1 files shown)


xenodium (approved):
Thank you for the fix

PR #761, opened 2026-08-08
Fix Codex ACP auth and command output

mrychlik:
## Summary

- update Codex ACP authentication method IDs to match current codex-acp releases
- render Codex command output from rawOutput.formatted_output with content blocks as fallback
- make idle timer state updates tolerate legacy alist state without :idle-timer
- add regression coverage for Codex auth IDs, command output extraction, and missing idle-timer state

## Root cause

agent-shell was sending obsolete Codex ACP authenticate method IDs such as chatgpt, openai-api-key, and codex-api-key. codex-acp 1.1.14 advertises and accepts chat-gpt and api-key, so startup failed with JSON-RPC -32600 Invalid request on the authenticate request.

Codex tool completion updates also carry terminal output under rawOutput.formatted_output rather than params.update.content, leaving completed shell command blocks empty in Emacs and transcripts.

## Validation

- reproduced the invalid authenticate request against local codex-acp 1.1.14
- verified methodId chat-gpt no longer returns the invalid-request response
- emacs -Q --batch ... -l tests/agent-shell-openai-tests.el -f ert-run-tests-batch-and-exit
- focused ERT tests: agent-shell--tool-call-output-text-test and agent-shell-idle-event-adds-missing-idle-timer-state-test


xenodium:
Thank you for the PR!

Looks like perhaps the PR is based on an older agent-shell checkout/revision? Doesn't https://github.com/xenodium/agent-shell/commit/cd2e6a663eae2d4fe1c8628d090104b5c2148876 offer what you need already?

If keen, we could still use the `rawOutput.formatted_output` change, but please isolate from all the other changes in this PR. I'd recommend a new PR to avoid piling up noise in commits.

PR #762, opened 2026-08-09
Document viewport workflow

KarimAziev:
I was confused about the viewport commands. Sometimes I accidentally invoked the command that opens the viewport, but it wasn't at all clear which command was responsible. I searched for all commands with "viewport" in the name, but none of them looked like the one that opened it. I checked the docs too, but found nothing there either. Finally, I discovered that it was `agent-shell-other-buffer`. 

This PR extends the documentation with that information and warns users not to invoke the viewport edit/view modes directly in an agent-shell buffer.

## Checklist


xenodium (approved):
Thanks for the fix!

PR #763, opened 2026-08-09
Render raw tool output

mrychlik:
## Summary

- render tool call output from rawOutput.formatted_output when present
- keep the existing content block rendering as the fallback
- add focused regression coverage for both paths

## Validation

- emacs -Q --batch -L /home/marek/.emacs.d/straight/repos/acp.el -L /home/marek/.emacs.d/straight/repos/shell-maker -L . -l tests/agent-shell-tests.el -f ert-run-tests-batch-and-exit


xenodium:
Thank you for the improvement. Merged.

PR #765, opened 2026-08-10
Prevent syntax highlighting from delaying other buffers' mode hooks

Scott-Guest:
## Problem

`agent-shell-markdown--highlight-code` manually binds `delay-mode-hooks` while activating a language major mode in a temporary buffer. Because `delay-mode-hooks` was not first made buffer-local, that dynamic binding temporarily makes its default value non-nil. 

As a result, another buffer without its own buffer-local value can see mode-hook delays intended only for the syntax-highlighting buffer. When that buffer calls `run-mode-hooks`, Emacs queues its mode hooks instead of running them. The buffer's mode initialization may then remain incomplete until `run-mode-hooks` is called there again. In particular, this can happen while [`agent-shell-desktop.el`](https://github.com/timfel/agent-shell-desktop.el) is restoring shells and one of their large transcripts is being rendered.

## Changes

- Use Emacs's built-in `delay-mode-hooks` macro, which makes the variable buffer-local before binding it and preserves the intended suppression across the major mode's `kill-all-local-variables` call.
- Add a regression test that verifies hooks remain delayed in the syntax-highlighting buffer while hooks in another buffer run normally.

## Verification

- Confirmed the regression test fails before the fix because the observer buffer sees the dynamically bound default of `delay-mode-hooks` as non-nil and queues its hook instead of running it.
- Confirmed the regression test passes after the fix.
- Ran the full test suite

## Checklist

### DIFF
--- a/agent-shell-markdown.el
+++ b/agent-shell-markdown.el
@@ -3240,10 +3240,10 @@ unchanged."
             ((fboundp mode)))
       (with-temp-buffer
         (insert code)
-        (let ((inhibit-message t)
-              (delay-mode-hooks t))
-          (funcall mode)
-          (font-lock-ensure))
+        (let ((inhibit-message t))
+          (delay-mode-hooks
+            (funcall mode)
+            (font-lock-ensure)))
         (buffer-string))
     code))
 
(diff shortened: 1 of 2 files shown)


xenodium (approved):
Thank you for the fix!

PR #766, opened 2026-08-11
Refresh viewport header even with nil agent-shell-prefer-viewport-interaction

catern:
agent-shell-prompt-compose creates a viewport buffer without requiring agent-shell-prefer-viewport-interaction to be non-nil.  We should update the header of that buffer even so.

## Checklist


xenodium:
Thank you!

PR #768, opened 2026-08-12
Add "agent-shell-queue-list" to see what’s queued

matthewbauer:
Previously, you couldn’t see what you had queued. This pops up a small buffer to display what’s queued. You can’t edit it, however.

Not sure if you want something quite like this, but was going to include it sense it was quite easy to add w/ claude's help.

## Checklist

### DIFF
--- a/agent-shell-prompt-queue.el
+++ b/agent-shell-prompt-queue.el
@@ -31,6 +31,7 @@
 
 (require 'map)
 (require 'ring)
+(require 'text-property-search)
 (require 'agent-shell-faces)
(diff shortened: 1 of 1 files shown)


xenodium:
Thanks for the PR. Finally applied and gave it a try. While I understand the appeal of having a dedicated command to list queued items, the implementation is fairly equivalent to what's available when invoking `M-x agent-shell-prompt-queue-remove`. I think for the time being, it's probably better to exclude the addition until we have a better picture of what we'd like to offer via a dedicated buffer. Coincidentally, this queue management mode could also be a third party package.

PR #772, opened 2026-08-15
Ask about killing the shell only when that is what was asked for

OSadovy:
Sending a composed prompt with C-c C-c asked "Kill shell session too?" whenever the viewport had been opened by anything other than `agent-shell-prompt-compose' -- which is the only path that sets `agent-shell-viewport-dismiss-on-send' buffer-locally, and so the only one that dismisses rather than kills.

`agent-shell-viewport-compose-send-and-kill' already binds `agent-shell-viewport--clean-up' to nil around the kill, with a comment saying exactly why the question should not be asked.  The binding never took effect: the variable is declared at the foot of the file, well after this use, so at compile time the symbol was not yet special and under lexical binding the `let' produced a dead lexical variable.  The dynamic value stayed t and `kill-buffer-hook' asked anyway.  The compiler said so in two warnings:

  Unused lexical variable 'agent-shell-viewport--clean-up'
  Variable 'agent-shell-viewport--clean-up' declared after its first use

Moving the declaration above its first use makes the existing binding dynamic, which is what it was written to be.  Both warnings go, and sending no longer offers to kill the session it just sent to.

## Checklist

### DIFF
--- a/agent-shell-viewport.el
+++ b/agent-shell-viewport.el
@@ -79,6 +79,8 @@
 (declare-function agent-shell-completion-mode "agent-shell-completion")
 (declare-function agent-shell-yank-dwim "agent-shell")
 
+(defvar-local agent-shell-viewport--clean-up t)
+
 (defvar agent-shell-header-style)
 (defvar agent-shell-prefer-viewport-interaction)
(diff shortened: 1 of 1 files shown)


xenodium (approved):
Argh... thanks!

Issue #773, opened 2026-08-15
[feature] steer prompts into the running turn


OSadovy:
Today a prompt sent while a turn is running is queued until the turn ends (#487), so a wrong direction can only be corrected after it has finished being wrong - unlike e.g. claude code cli experience, where one can correct the agent mid-turn.

Both adapters I use already support injecting a prompt into the running turn:

- Method: `_session/steering` (an extension, not in the ACP schema)
- Capability: `InitializeResponse._meta.steering.supported` — top-level `_meta`, a sibling of `agentCapabilities`, not inside it
- Params: `{sessionId, prompt}`, `prompt` being the same content blocks as `session/prompt`
- Outcomes: `injected`, `startedNewTurn`, `failed`, and `promptRequired` (Claude only, opt-in via `_meta.steering.idleBehavior`)

Verified against `@agentclientprotocol/claude-agent-acp` 0.66.0 and
`@agentclientprotocol/codex-acp` 1.2.0.

The standards-track equivalent is `session/inject` (agentclientprotocol/agent-client-protocol#1261, still open, targeting ACP v2), so the extension is the only thing usable today.

Behaviour differs by agent: Claude aborts the running generation and answers in ~2s; Codex accepts the steer but finishes the current message first — on the same prompt, ~2.5 minutes.

Proposal: when the agent advertises steering and a turn is running, steer instead of queueing; queue otherwise. Any non-`injected` outcome falls back to queueing so the prompt is never lost. Allow users to opt-out and queue instead of steering.

I have a working implementation and am happy to open a PR if the direction looks right.

## Checklist

- [x] *I agree to communicate with the author myself* (not AI-generated).
- [x] I've read the README's [Filing issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#filing-issues) section.
- [x] I'm running the latest versions (fill in below).
  - agent-shell version: 0.72.2
  - acp.el version: 0.13.1
  - ACP package (e.g. claude-agent-acp) version: @agentclientprotocol/claude-agent-acp 0.66.0 (also tested @agentclientprotocol/codex-acp 1.2.0)
  - Agent CLI (e.g. claude, gemini) version: claude 2.1.220
- [ ] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [ ] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README).

[...]

PR #777, opened 2026-08-18
Steer prompts into the running turn instead of queueing them

OSadovy:
Implements the steering proposal from #773.

When the agent advertises steering and a turn is running, a prompt sent from the
shell is injected into that turn instead of being queued until it ends. Agents
that don't advertise it queue exactly as before.

The method is `_session/steering` — an ACP extension rather than spec (hence the
leading underscore), advertised in the `initialize` response's top-level `_meta`
as `steering.supported`, a sibling of `agentCapabilities`. Verified against
`@agentclientprotocol/claude-agent-acp` 0.66.0 and `@agentclientprotocol/codex-acp`
1.2.0. The standards-track equivalent, `session/inject`
(agentclientprotocol/agent-client-protocol#1261), is unmerged and targets ACP v2,
so the extension is what's usable today.

**Queueing stays the fallback and nothing is lost.** A steer is attempted only
when the agent advertises it, `agent-shell-steer-when-busy` is on, and a turn is
running that isn't blocked on a permission answer. Every other outcome — the turn
had ended, the agent refused, the request failed — either submits the prompt
normally or puts it back on the queue. A shell waiting on a permission answer
keeps queueing deliberately: what an agent does with a message injected while a
tool sits on that question isn't defined by any implementation I could find.

Entry points:

- Sending from a busy shell steers automatically.
- `C-u M-x agent-shell-prompt-queue` queues that one prompt instead.
- `M-x agent-shell-steer` steers explicitly and errors when it can't.
- `agent-shell-steer-when-busy` (the one new defcustom, since #773 asked for
  opt-out) turns the automatic behaviour off entirely.

Neither adapter echoes a steered prompt back while the turn runs, so it's rendered
client-side, reusing the field/face shape already used for replayed user messages
and closed with shell-maker's end-of-prompt marker — without that, chat mode reads
it as the live prompt and the prompt bar hides it. Also adds a `prompt-steered`
session event.

**On automating cancel-and-resend instead** (your comment in the issue): I don't
think they're equivalent. Interrupting drops the work in flight and leaves a
cancellation in the transcript, while a steer keeps the turn and what it has
already done. They compose, though — `agent-shell-steer` could fall back to
interrupt-and-continue for agents that don't advertise steering, which is what
@liaowang11 suggested. Happy to do that as a follow-up if you'd like it.

Behaviour differs per agent, worth knowing before trying it: Claude interrupts
what it's generating and answers in a couple of seconds; Codex accepts the steer
but finishes the message in flight first.

Testing: five tests covering the request shape, outcome mapping, the decision to
steer, the four-way fallback (including that a failed steer keeps the prompt), and
the rendered prompt. Full suite passes at 568.

## Checklist

### DIFF
--- a/README.org
+++ b/README.org
@@ -672,6 +672,14 @@ In view mode, use =r= to compose a reply, =R= to quote the response in a reply, 
 Do not invoke =agent-shell-viewport-edit-mode= or =agent-shell-viewport-view-mode= directly in an agent shell buffer.  They are major modes installed by the entry points above on the separate viewport buffer, not commands for converting the shell buffer.
 #+end_quote
 
+*** Steering a running turn
+
+A prompt sent while the agent is working is queued and submitted once the turn ends.  Agents that support steering can take it into the turn already running instead, so a wrong direction can be corrected without waiting for it to finish.  Steering is an ACP extension (=_session/steering=) rather than part of the spec, advertised by the agent at initialization: the Claude and Codex packages implement it, other agents keep queueing.
+
+Sending from a busy shell steers automatically when the agent supports it.  Use a prefix argument (=C-u=) to queue that one prompt instead, or set =agent-shell-steer-when-busy= to nil to always queue.  =M-x agent-shell-steer= steers explicitly, and errors when there is no running turn to steer or the agent cannot steer.  A steer that the agent does not accept falls back to queueing, so the prompt is never lost.
+
+What the agent does with the steered prompt is up to the agent: Claude interrupts what it is generating and responds to it, while Codex finishes the message in flight first.
+
 *** Specific Agent Commands
 
 Start a specific agent shell session directly:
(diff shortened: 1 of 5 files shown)


xenodium:
Thanks a lot for the PR. I've been taking it for a spin before merging. Great feature!

My main concern is coupling queueing with steering prematurely as we likely haven't yet discovered all the edge cases or unexpected behavior. This may leave folks a little puzzled if queueing starts behaving differently at times.

For example, I had an instance where steering completely stopped the current task and started working on a new task from steering. For now, I would like to surface these cases as early as possible to avoid misleading users.

I've merged the PR as is. Having said that, my preference is to strip the feature back to core scope. That is, keep steering (via `M-x agent-shell-prompt-steer`) and queuing separate for the time being while we let the feature settle and we learn from its quirks.

I prepared branch https://github.com/xenodium/agent-shell/tree/steer and applied its changes to main as a fast-follow patch to the PR. I understand this may divert from your target scope. I kept going back and forth as to how to best apply these changes and eventually landed on this path. Hope that's ok.

Issue #781, opened 2026-08-20
Feature request: jump to file:line references in agent output


OSadovy:
## Checklist

- [x] *I agree to communicate with the author myself* (not AI-generated).
- [x] I've read the README's [Filing issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#filing-issues) section.
- [x] I'm running the latest versions (fill in below).
  - agent-shell version: 0.73.4
  - acp.el version: 20260803.2034
  - ACP package (e.g. claude-agent-acp) version: claude-agent-acp 0.67.0
  - Agent CLI (e.g. claude, gemini) version: claude 2.1.234
- [ ] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [ ] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README).

---

Agents cite their sources constantly: `docs/audit.md:500`, `src/main.rs:120-140`,
sometimes with a column like `foo.el:12:5`. Right now those are just text in the
buffer, so getting to one means retyping the path somewhere else.

I'd like to press RET on a reference and land in the file on that line.

Almost all of them arrive wrapped in backticks, since that's how agents write
them. A few turn up in plain prose. There are also some inside fenced blocks,
usually grep output — I'm honestly not sure whether those should be links too,
or left alone.

Ideally they'd also be stops for item navigation, so `n` and TAB reach them
without having to put point on one first.

I have a version of this working in my own config and would be happy to send a
PR if you're interested.

xenodium:
Ah interesting. I stopped noticing since I have https://github.com/xenodium/emacs-skills/blob/main/skills/file-links/SKILL.md and they are always clickable. I suppose this can also be added to agents.md or even the agents system prompt. I'm thinking we may get better reliability and possibly performance by influencing the agent. Want to share your changes (even if just a draft) to get a feel for things?

xenodium:
Thanks for https://github.com/xenodium/agent-shell/pull/782/commits/36ec29669acb47ca4256294682c4f9edc28299ea

argh... I went to fix some link issues and ended up adding support for your request too. Sorry. Changes are in https://github.com/xenodium/agent-shell/commit/9595db2b4714381ba1b0e77a14b73914c3f40f3b. Does it work ok for you?

OSadovy:
[...]

PR #782, opened 2026-08-20
Link the file references agents cite in prose

OSadovy:
## Checklist

- [x] I've filed a feature request/discussion for a new feature. (#781)

---

Draft for #781, to get a feel for things as you asked. Rewrote my advice config into a patch against agent-shell.

On the [skill](https://github.com/xenodium/emacs-skills/blob/main/skills/file-links/SKILL.md): That one is elegant! I still think the two compose rather than compete. A link your skill
produces is already handled, and this pass leaves it alone — it skips text
that already carries a link target. This is for the agents that don't emit
links - e.g.  a session where the skill isn't installed. It also
doesn't depend on the agent complying every time, which is the part I can't
verify by reading the output.

Everything needed to *follow* one of these was already there.
`--parse-local-link` reads `path:LINE` and `path:LINE-LINE`, and
`agent-shell-markdown-visit-file` opens the file, lands on the line and pushes
the origin onto xref's marker stack. What was missing was detection —
`--linkify-urls` scans `browse-url-button-regexp`, which no bare relative path
matches, so a citation never reached the code that would have handled it.

One thing worth your eye:

**Inline code.** The new pass takes the fenced blocks alone as its
avoid-ranges rather than the merged set. Nearly every citation arrives inside
backticks, so avoiding inline spans skips nearly everything. Those bodies are
tagged `agent-shell-markdown-frozen`; my reading is that the tag exists to stop
later passes re-parsing markup that's already resolved, and adding properties
re-parses nothing — but you'd know if that's a stretch. A path inside a fenced
block does stay text, on the grounds that it's sample content or command output
rather than a citation.

Asked Opus to measure the cost. Here is what we got:

Costs, on a 35k buffer with `:force t` (streaming renders narrow to the new
chunk, so this is the pessimistic case):

| | without | with |
|---|---|---|
| no references | 2.3 ms | 3.3 ms |
| 20 references | 2.5 ms | 3.8 ms |
| 400 references | 2.4 ms | 8.2 ms |

So ~1 ms per 35k is the regexp scan, and each reference on top costs ~11 µs —
about half resolving the path (`file-exists-p` and `expand-file-name`), about
half applying the link properties, which any link pays anyway.

Only the resolving half is avoidable, and only by caching. Keyed on the path
that would save about three quarters of the checks in my own buffers (656
references, 165 distinct paths); keyed on the whole reference, only a quarter.
I left it out, since the 400-reference row is a reference every 88 characters
and my real sessions run one per ~2300 — call it 0.2 ms a render. Happy to add
it if you disagree.

### DIFF
--- a/agent-shell-markdown.el
+++ b/agent-shell-markdown.el
@@ -425,6 +425,10 @@ body un-fontified."
         ;; After the markup passes, so a URL already consumed as a
         ;; `[title](url)' destination or an image's is not seen again.
         (agent-shell-markdown--linkify-urls :avoid-ranges avoid-ranges)
+        ;; Fenced blocks alone, not AVOID-RANGES: a citation is nearly
+        ;; always inside an inline `code' span, which those would skip.
+        (agent-shell-markdown--linkify-file-references
+         :avoid-ranges source-ranges)
         (agent-shell-markdown--style-dividers :avoid-ranges avoid-ranges)
         (agent-shell-markdown--style-blockquotes :avoid-ranges avoid-ranges)
         (agent-shell-markdown--style-lists :avoid-ranges avoid-ranges)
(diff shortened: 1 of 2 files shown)


OSadovy:
Superseded by 0306939 — thanks @xenodium. Your version handles the streaming case mine got wrong, and excusing the frozen tag only for positively-identified inline spans is cleaner than my blanket approach. Closing.

xenodium:
Thanks for flagging and proposing a fix

Issue #788, opened 2026-08-20
Dired DWIM context can insert prompt garbage into new agent sessions


rpoisel:
## Description

Starting an agent session from a Dired buffer can accidentally insert the current Dired row as prompt context.

This happens because automatic DWIM context may treat Dired's current-line fallback as an obvious file selection, or fall through to raw line context and send listing text such as `.git/`.

## Steps to Reproduce

1. Open a Dired buffer.
2. Move point to an unmarked file or directory row, e.g. `.git/`.
3. Start an agent session with `M-x agent-shell-openai`.
4. Observe that prompt context may include the Dired row or `@.git`.

## Expected Behavior

Starting a session from Dired should not insert file or line context unless the user explicitly selects or marks something.

## Checklist

- [x] I agree to communicate with the author myself (not AI-generated).
- [x] I've read the README's Filing issues section.
- [x] I'm running the latest versions (fill in below).
  - agent-shell version: 26969ef21808218a439d6bb86face1333ae76169
  - acp.el version: 20260820.906
  - ACP package version: `@agentclientprotocol/codex-acp` 1.6.2
  - Agent CLI version: `codex` 0.148.0
- [x] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package. Not applicable.
- [x] For issues, I'm including ACP traffic (as per README). Not applicable.

xenodium:
Thanks for the feature request.

> Starting a session from Dired should not insert file or line context unless the user explicitly selects or marks something.

I personally use this feature quite a bit. It removes a little friction from having to mark a file before sending to the shell. Being emacsers, we can all argue subjectively as to what's the right approach until another emacser comes along with yet another way of doing things. Rather than changing the default, I'd prefer to enable you to configure agent-shell in your preferred way. With that in mind, couldn't you replace `files` from `shell-context-sources` with a custom function using you're preferred method? To make it less involved, I could perhaps offer another option named `explicit-files`. Would that help?

PR #789, opened 2026-08-20
Fix Dired DWIM context fallback

rpoisel:
Dired's current-line fallback is useful for explicit file sending, but it should not count as obvious context when starting or reusing a shell.

Only use region-selected or explicitly marked Dired files for automatic file context, and skip current-line context in Dired so raw listing rows are not inserted into prompts.

Note: I am not really good at writing elisp code. This commit has been created with the assistance of an AI agent (OpenAI Codex). I do understand the code, but there might be better ways to improve the situation. It works for me - I am typically starting agent-shell sessions from dired buffers without selecting entries (files, directories, etc.).

## Checklist

- [x] I've filed a feature request/discussion for ~a new feature~ the observed behavior: #788 
- [] I've updated documentation where necessary.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -8423,7 +8423,18 @@ Buffer filename is OBVIOUS if its an image."
       (list (buffer-file-name))
     (or
      (agent-shell--dired-paths-in-region)
-     (dired-get-marked-files))))
+     (when (and obvious (equal major-mode 'dired-mode))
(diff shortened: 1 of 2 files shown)


rpoisel:
This PR does not fix the referenced issue under all circumstances. So, it can be merrily understood as a first try to improve the situation but not a real fix.

PR #794, opened 2026-08-22
Mark restored prompts with comint-highlight-prompt

OSadovy:

## Checklist

## The bug

After restoring a session, `C-c C-o` on a restored interaction silently
shows the *previous* one in the viewport. No error, nothing in
`*Messages*`.

## Why

`agent-shell.el:3011` renders the prompt prefix for a `user_message_chunk`
replayed during `session/load`. Its own comment says it means to match the
field/face shape comint emits for a live prompt, but it applies
`agent-shell-prompt` alone.

That face *inherits* `comint-highlight-prompt`, so the prompt looks
correct. But `shell-maker--re-search-forward-prompt` matches with
`(memq 'comint-highlight-prompt (ensure-list ...))`, and inheritance
doesn't satisfy `memq`. The restored interaction never enters
`shell-maker--extract-history`, so `agent-shell-interaction-at-point`
returns nil — and `agent-shell-viewport-refresh` guards its entire body on
that value, so it does nothing at all.

Same applies to `session/push`, which goes through the same branch.

## The fix

Append the face rather than substitute it.

`agent-shell-prompt` stays first for precedence and remains customizable,
and downstream packages matching on it with `memq` (there are tsome) keep working.

## Verified

Against `shell-maker--re-search-forward-prompt` on the rendered prompt
text:

```
agent-shell-prompt                             -> not found
(agent-shell-prompt comint-highlight-prompt)   -> found
```

No tests added since I think the change is trivial, but happy to add one. The existing suite
passes.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -3008,7 +3008,7 @@ Clears STATE's `:expanded-activity-group'."
                           (concat (propertize
                                    (map-nested-elt
                                     state '(:agent-config :shell-prompt))
-                                   'font-lock-face 'agent-shell-prompt
+                                   'font-lock-face '(agent-shell-prompt comint-highlight-prompt)
                                    'field 'output)
                                   (propertize content-text
                                               'font-lock-face 'agent-shell-input))


xenodium (approved):
Thank you for the fix

Issue #792, opened 2026-08-21
[Feature] support agent-initiated questions (ACP elicitation)


OSadovy:
## Checklist

- [x] *I agree to communicate with the author myself* (not AI-generated).
- [x] I've read the README's [Filing issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#filing-issues) section.
- [x] I'm running the latest versions (fill in below).
  - agent-shell version: 0.74.2
  - acp.el version: 20260803.2034
  - ACP package (e.g. claude-agent-acp) version: 0.70.0
  - Agent CLI (e.g. claude, gemini) version: 2.1.234
- [ ] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [ ] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README).

## The problem

Agents can ask the user a structured question mid-turn - "which library
should we use?", "which of these approaches?" - and wait for an answer
before continuing. ACP calls this elicitation.

agent-shell doesn't support it. Elicitation is opt-in - a client has to
advertise it before an agent will use it - so this isn't a visible failure
today. It just never happens.

What agents do in its absence varies. Some disable their question tool
outright, so the model can no longer ask at all. Others degrade it to a
yes/no permission prompt. Questions originating from MCP servers are
auto-declined.

That's a behaviour difference - a decision that would have been put to the
user might get settled by the model some other way instead (more guessing?).
And a fair number of published skills and MCP servers are written assuming
the client can ask a question and wait for the answer.

## How this might look in agent-shell

The question appears in the transcript as a normal entry, in place, not as a
popup or a minibuffer prompt. Nothing is modal - the buffer stays usable. A
single request can carry several questions, so one entry may hold more than
one.

Options are listed and numbered, each with a marker showing selection state:

```
    Which serialization library?

    1. ( ) serde
           Mature, derive macros, largest ecosystem
    2. (*) miniserde
           Smaller, fewer features, faster compiles
    3. ( ) Hand-rolled
           No dependency, more code to maintain

    Other: ________________________________

     Submit   Decline
```

Round markers for single-choice, square for multiple-choice, so the two are
distinguishable at a glance without reading a label. Selecting an option in a
single-choice question clears the others; in a multiple-choice question it
toggles.

Selecting works three ways: press the option's number, press `RET` with point
[...]

Issue #759, opened 2026-08-07
Expand/collapse group with TAB


Azkae:
Hello,

I prefer to expand or collapse groups using TAB instead of RET.
So I have this change in my agent-shell repo:

```
diff --git a/agent-shell-ui.el b/agent-shell-ui.el
index 155deb0..0f77b90 100644
--- a/agent-shell-ui.el
+++ b/agent-shell-ui.el
@@ -1307,7 +1307,7 @@ block's beginning instead of the previous block."
   "Create keymap with ACTION."
   (let ((map (make-sparse-keymap)))
     (define-key map [mouse-1] action)
-    (define-key map (kbd "RET") action)
+    (define-key map (kbd "TAB") action)
     (define-key map [remap self-insert-command] 'ignore)
     map))
 
```

I am wondering if we could make this customizable, maybe with a keymap. I can provide a PR if you are interested.
I have a work in progress [here](https://github.com/xenodium/agent-shell/compare/main...Azkae:agent-shell:feature/action-keymap)

xenodium:
Thanks for filing.

I'd been meaning to look more into this, so I've gone down a rabbit hole to simplify agent-shell-ui.el a little. Change is in https://github.com/xenodium/agent-shell/commit/961709016a911d33e37b64ef4e352416ddc40ba2

You can now bind to TAB as follows:

```emacs-lisp
(define-key agent-shell-ui-fragment-map (kbd "TAB") #'agent-shell-ui-toggle-fragment)
```

Beware: Out of the box, TAB is bound to agent-shell-next-item in agent-shell-mode-map, so you'll need to change that binding also and possible agent-shell-previous-item.

Lemme know if this doesn't handle your use-case.

Azkae:
That's perfect, thank you!

PR #797, opened 2026-08-23
Render ACP elicitation forms

OSadovy:
Implements #792.

> The code is mostly AI-written. I've been using it daily since and have tested the behaviours below against a live agent, so I'll vouch for it working.

## What it does

- Advertises `elicitation.form` in `initialize`. `url` is not supported yet.
- Renders `elicitation/create` as an in-buffer form covering every `ElicitationPropertySchema` variant, both spellings of each enum shape (`oneOf`/`enum`, `items.anyOf`/`items.enum`), and the forward-compat unknown type.
- Validates required-presence on submit only; `minLength`, `pattern`, `minimum`, `minItems` and friends are not implemented yet - left for the agent to reject.
- TAB walks the form as one more item kind beside prompts, blocks and permission buttons; RET acts. Named commands reading text properties rather than per-control closures (#759).
- Reports `blocked` while a form is pending, so a queued prompt isn't steered into a turn that's waiting on an answer.
- Cancels pending forms on interrupt, and arms the same idle timer an unanswered permission does.

## Slightly opionated behaviour that isn't obvious

- **The free-text "Other" field folds into the select it belongs to.** Agents bridging an ask-the-user tool send a select plus a companion text field, marked with an un-namespaced `_meta` key. Rendered as two independent fields, one question reads as two and can be answered twice in contradictory ways.
- **On a single-select the two replace each other; on a multi-select they combine into one string.** Those agents read the companion first and return early, so a response carrying both silently loses the ticks - verified against the Claude adapter. AFAIK Zed is not doing it - we can do better here.
- **A checkbox answers only when ticked, required, or carrying a default.** An unticked optional box is indistinguishable on screen from one nobody reached. Same rule Zed's form uses.
- **Integer fields reject a fraction rather than truncating it**, and each retry is prefilled with the text that was rejected.
- **An elicitation with no tool call behind it** - from an MCP server, or scoped to a request rather than a session - keeps its settled form, which is the only record of what was asked.
- **Option previews** open from the option (`?`) or its disclosure glyph, and are offered as `help-echo` so `C-h .` reads them without opening. `?` is bound only on options that actually carry a preview, so it stays free everywhere else - the viewport still opens its help menu with it. That `_meta` key is agent-namespaced, so previews are Claude-only and their absence changes nothing.

## Incidental fixes

Happy to submit those as separate PRs.

- `agent-shell-ui--echo-action-hint` takes a command and keymap, so chrome bound to its own map names its own key.
- A fragment no longer erases the `help-echo` its body's content carried - it set the property to nil whenever debugging was off.

## Not included

`url` mode; constraint validation; digit shortcuts; eldoc.

## Where this differs from #792

I wrote the issue before building any of it, and a few things changed on the way.

- Options aren't numbered and there are no digit shortcuts - the digits are already taken in the viewport map, and without them the numbering is just noise.
- "Other" is a minibuffer read rather than an editable field inline in the buffer. An editable region inside otherwise read-only rendered text is a lot of machinery for one field, and the form re-renders on every interaction, which would fight it.
- TAB stops on every control rather than jumping question to question.
- A cancelled form shows `Cancelled` without preserving whatever had been typed into it.
- No SVG radio/checkbox markers - text only for now.

## One question

`acp-make-initialize-request` hardcodes `clientCapabilities` to the `fs` pair, so there's no way to advertise elicitation through it. I build the request locally instead (`agent-shell--make-initialize-request`), which does mean agent-shell no longer picks up future changes to that constructor. Would you rather acp.el grew a capabilities argument? Happy to send that first and keep using the helper here.

## Checklist

### DIFF
--- a/README.org
+++ b/README.org
@@ -923,6 +923,14 @@ All of the above settings can be applied on a per-project basis using [[https://
 - =C-c C-c= - Interrupt current agent operation
 - =TAB and Shift-TAB= - Navigate interactive elements
 
+  When an agent asks a question, its form renders in the buffer and =TAB=
+  stops on each option, field and button:
+
+  - =RET= - Pick the option, toggle the checkbox, edit the field, or press
+    the button at point.
+  - =C-c C-c= - Interrupt, cancelling any question still waiting for an
+    answer.
+
   To customize =RET= binding behaviour, you can use something like:
 
   #+begin_src emacs-lisp :lexical no
--- a/README.org
+++ b/README.org
@@ -923,14 +923,6 @@ All of the above settings can be applied on a per-project basis using [[https://
 - =C-c C-c= - Interrupt current agent operation
 - =TAB and Shift-TAB= - Navigate interactive elements
 
-  When an agent asks a question, its form renders in the buffer and =TAB=
-  stops on each option, field and button:
-
-  - =RET= - Pick the option, toggle the checkbox, edit the field, or press
-    the button at point.
-  - =C-c C-c= - Interrupt, cancelling any question still waiting for an
-    answer.
-
   To customize =RET= binding behaviour, you can use something like:
 
   #+begin_src emacs-lisp :lexical no
--- a/agent-shell-elicitation.el
+++ b/agent-shell-elicitation.el
@@ -170,8 +170,32 @@ For example:
              (list (cons :value (map-elt option 'const))
                    (cons :title (or (map-elt option 'title)
                                     (map-elt option 'const)))
-                   (cons :description (map-elt option 'description))))
+                   (cons :description (map-elt option 'description))
+                   (cons :preview (agent-shell-elicitation--option-preview option))))
            (append options nil)))
+
+(defconst agent-shell-elicitation--option-meta-key '_claude/askUserQuestionOption
+  "`_meta\=' key an option's preview travels under.
+
+Unlike the free-text companion marker, this one is namespaced to the
+agent that sends it, so only Claude sends a preview today.  An option
+without it renders without a preview to open.")
+
+(defun agent-shell-elicitation--option-preview (option)
+  "Return OPTION's preview text, or nil when there is none.
+
+A preview is the longer form of an answer -- a mockup, a snippet, an
+outline of what picking it would do -- that an agent attaches for the
+client to show on demand.
+
+For example:
+
+  (agent-shell-elicitation--option-preview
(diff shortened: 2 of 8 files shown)


xenodium:
Thank you for the PR and sorry for the delay.

> which does mean agent-shell no longer picks up future changes to that constructor. Would you rather acp.el grew a capabilities argument?

Yes please. acp-make-initialize-request doesnt offer other options likely because no one else asked for new ones.

checkdoc fails agent-shell-elicitation.el:1047 and :1464, "Disambiguate agent-shell--state" (needs \= or backquote in the docstring).  Mind running checkdoc and fixing the warning? Same for other files in case something else is picked up.

I'd like to give this fature more soaking and maybe iterate on UX before we roll it out to all. For example, could TAB navigation land cursor/point inside the `[ ]`? Is it possible to block submission if questions are not answered (without adding much complexity)? No need to address these now in this PR. Mind adding `defvar agent-shell-elicitation--experimental-feature-enabled` so we can use the feature ourselves initially? That'd give me more time to give it a a more thorough spin. When we're ready to roll it out, we'll remove the defvar.

OSadovy:
Rebased on main and addressed the review.

- Fixed the two checkdoc warnings, and swept the other files for
  anything else. Nothing new turned up.
- Added agent-shell-elicitation--experimental-feature-enabled,
  defaulting to nil. With it off the elicitation capability goes
  unadvertised, so nothing raises a form, and an agent that asks anyway
  gets the existing method-not-found answer. Dropped the README section
  since it documented something off by default.
- Also fixed a batch of docstring escapes emitting a stray "=" in this
  branch's own docstrings, the same thing a792324 and 5e2d314 fixed on
  main. Separate commit, happy to drop it if you'd rather it went alone.

acp.el side is xenodium/acp.el#30. Once that lands I'll swap
agent-shell--make-initialize-request back to the helper.

On blocking submission when questions aren't answered: that gate exists
already, but it keys off the schema's `required` array, and the Claude
adapter sends no `required` at all, so in practice it never fires for
me. Easy fix is to also refuse a form where nothing has been answered,
since Decline is the way to say you're not answering.

OSadovy:
Swapped the handshake back to `acp-make-initialize-request` now that xenodium/acp.el#30 is out in acp 0.15.1, which agent-shell already requires, so `agent-shell--make-initialize-request` is gone. Also rebased on main.

PR #801, opened 2026-08-25
Strip copies in the substring filter, not via yank-handler

OSadovy:
## Problem

Rendered output and block-quoted replies carry a `yank-handler` text property
whose lambda strips properties (and the leading `> `) at paste time. The
property rides on the characters, which has surprising reach:

- Once the text lands in any other buffer, copies from *there* still run
  agent-shell's lambda - for the rest of the session, in buffers unrelated to
  agent-shell.
- `insert-for-yank` is not just `C-y`: registers, rectangles, `cua`, xterm
  paste, and any third-party property-preserving insertion all hit the handler.
- The same `M-w` yields different content depending on destination: `C-y`
  gives `hello`, while the OS clipboard and isearch's `C-s C-y` keep the
  stored `> hello`. `yank-excluded-properties` is never consulted for handler
  text, so no setting affects any of this.

## Change

Delete both `yank-handler` puts and finish the job in
`agent-shell--filter-buffer-substring`, which shell and viewport buffers
already install as `filter-buffer-substring-function` - the hook Emacs
provides for deciding what a copy of a buffer yields. Copies now come out
property-free, and one `> ` per line is dropped from text tagged with a
private `agent-shell-block-quote` marker (set by `agent-shell--block-quote`
in place of the handler). The decision happens once, at copy time, in buffers
agent-shell owns; paste stays stock Emacs.

## Behaviour

`M-w`, `C-w`, mouse copy, and registers out of agent-shell buffers paste the
same as before. Markdown-rendered blockquotes still keep their `> ` on copy,
so the source round-trips. Two visible changes, both toward consistency:

- Copies of block-quoted replies are now identical everywhere. Previously the
  OS clipboard and isearch kept the `> ` that yank removed.
- Strings from the standalone `agent-shell-markdown-convert` API, killed from
  non-agent-shell buffers, now paste with stock Emacs semantics. No in-tree
  caller does this. Happy to gate it behind a keyword if you'd rather keep
  the old behaviour there.

## Context

Found while speech-enabling agent-shell with Emacspeak: its speech layer
inserts buffer text via `insert-for-yank` precisely to keep the text
properties it builds voices from, and the handler strips them. Any consumer
using `insert-for-yank` for property-preserving insertion hits the same wall,
and a buffer-local filter cannot break them.

Tests: the yank test is rewritten as filter tests - properties stripped,
`> ` dropped from quoted replies, one level per line for nested quotes,
markdown blockquotes keep theirs. Full suite passes.

## Checklist

- [ ] I've filed a feature request/discussion for a new feature. (N/A - bug-ish cleanup, no new feature)
- [ ] I'm making visual changes, so I'm including screenshots so you can view and discuss. (N/A - no visual changes)
- [x] I've updated documentation where necessary. (docstrings updated)

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -2096,9 +2096,13 @@ If DELETE is non-nil, delete the text between START and END.
 
 A paste elsewhere should give plain characters rather than agent-shell's
 implementation, so every text property is dropped: our faces, keymaps,
-cursor sensors, display overrides and internal markers all go.  The one
-kept is `yank-handler', which rewrites content on insertion rather than
-decorating it (see `agent-shell--block-quote').
+cursor sensors, display overrides and internal markers all go.  One
+\"> \" per line goes too, from text `agent-shell--block-quote' marked,
+so a quoted reply copies as the plain text it shows.
+
+Deciding this at copy time rather than on insertion keeps one copy
+meaning one thing: a `yank-handler' would rewrite for `yank' while the
+system clipboard and isearch kept the raw \"> \".
 
 `agent-shell-markdown-replace-markup' renders by rewriting markup in
 place, so the buffer already holds rendered text and there is nothing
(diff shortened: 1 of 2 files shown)


xenodium:
Rebased to main and merged

xenodium:
Thanks for the PR!

PR #803, opened 2026-08-25
Label user and agent turns in `agent-shell-copy-as-markdown`

alberti42:
## The problem

Copying a transcript region with `agent-shell-copy-as-markdown` puts terminal chrome on the kill ring:

````
Pi> Let's make a new test. Print verbatim

```math
E=mc^2
```
<shell-maker-end-of-prompt>
```math
E=mc^2
```
````

`<shell-maker-end-of-prompt>` is `invisible t` — it is not visible in the buffer and it is not markdown, so it has no business in a copy at all. `Pi> ` is visible, but a bare terminal prompt is not how a markdown transcript marks a turn.

Everything else already round-trips correctly; this is only about the chrome around the turns.

## Why not just strip them

The marker is the *only* boundary between a submitted prompt and its response. Removing it leaves the two running together with nothing in between — worse than the status quo for anyone pasting a transcript into a document, an issue, or another LLM.

So this renders the chrome as markdown rather than stripping it.

## After

````
**Me:**

Let's make a test. Print verbatim

```math
E=mc^2
```

**Pi:**

```math
E=mc^2
```
````

## How it works

No new machinery. `agent-shell-markdown-reconstruct` already substitutes a span's `agent-shell-markdown-source` into the copy — the empty string being the documented "inserted chrome vanishes" case. The chrome simply was never stamped. **`agent-shell-markdown-reconstruct` is unchanged.**

Both spans are inserted by shell-maker, not agent-shell: it calls `shell-maker-insert-end-of-prompt-marker` itself on submit (shell-maker.el:2299) and emits the prompt through its own output filter. So stamping happens at **copy time**, not insertion time, which keeps the change inside agent-shell — **no shell-maker change is required**.

- `agent-shell--label-turn-chrome` stamps the two spans, found by properties they already carry: `shell-maker--marker` for the marker, and a prompt face for the prompt.
- `agent-shell--reconstruct-with-turn-labels` does this on a **scratch copy** of the region (`buffer-substring` carries the text properties reconstruct reads), so the shell buffer is never modified by a copy.

shell-maker's other markers (`<shell-maker-failed-command>`, `<shell-maker-interrupted-command>`) are stashed empty and vanish — they were being copied verbatim too.

## Shared naming with `agent-shell-chat-mode` (second commit)

`agent-shell-chat-mode` already labels these same two anchors on screen — it boxes each user turn `Me` and each response with the agent's name. But it draws them with **overlays**, and reconstruct reads **text properties**, so a copy cannot see them: with chat-mode on, the buffer reads as a clean chat while a copy still yields `Pi> ` and the raw marker.

Rather than inventing a second set of names, who is speaking is named once and used by both. Two private chat-mode helpers move to `agent-shell.el` (the copy path cannot require chat-mode, since `agent-shell.el` requires *it*):

```
agent-shell-chat--agent-name    -> agent-shell--agent-label
agent-shell-chat--prompt-face-p -> agent-shell--prompt-face-p
```

chat-mode's on-screen output is unchanged, but its user label — previously the hardcoded string `"Me"` — becomes customizable on the way, which seems worth having on its own: it is the one name a user is most likely to want in their own language.

## New options

| Option | Default | Effect |
|---|---|---|
| `agent-shell-user-label` | `"Me"` | Your name — chat-mode's box and copies |
| `agent-shell-agent-label` | `"%s"` | Agent's name; `%s` is its mode-line name, omit it for a fixed name such as `"Agent"` |

Two options, not three: a copy renders either name as `**Name:**`, which did not seem worth making configurable. Emptying a name leaves that side unlabelled while still dropping its chrome, so "no labels at all" needs no option of its own.

Substitution in `agent-shell-agent-label` is literal, so a `%` elsewhere in the string needs no escaping. The blank lines around a label are added in code, not configured, so no setting can produce lopsided spacing.

## Fixed along the way

Taking the prompt predicate from chat-mode also fixes prompt detection: `comint-highlight-prompt` alone misses restored and echoed prompts, which carry `agent-shell-prompt`. Those would have gone unlabelled.

## Possible extension (not in this PR)

`agent-shell-agent-label` could accept a function as well as a string (`:type '(choice string function)`), letting a user compose a label from richer per-turn metadata — which model answered, when.

That is deliberately left out, because agent-shell does not currently record it. The model is session-level (`:model-id`), so switching models mid-session would retroactively relabel every earlier turn, and there is no per-turn timestamp at all — only `:last-activity-time`, which a copy would render as "now". Recording per-turn model and time is a much larger change, plausibly belonging in shell-maker.

Adding the function option later would not break anything: the option would simply accept a function as well as a string, so any label already configured as a string keeps working.

## Testing

Select across a prompt and its response in a shell buffer and run `M-x agent-shell-copy-as-markdown`. Checked manually against a live shell, and in batch for: label substitution and spacing, the shell buffer being unmodified by a copy, the `%s`/fixed/`"%s (agent)"` label shapes, an empty name dropping that side's label and chrome, `<shell-maker-failed-command>` vanishing, and restored-prompt faces being labelled. Both files byte-compile clean.

### DIFF
--- a/agent-shell-chat-mode.el
+++ b/agent-shell-chat-mode.el
@@ -47,8 +47,11 @@
 
 (declare-function agent-shell-subscribe-to "agent-shell")
 (declare-function agent-shell-unsubscribe "agent-shell")
+(declare-function agent-shell--agent-label "agent-shell")
+(declare-function agent-shell--prompt-face-p "agent-shell")
 
 (defvar agent-shell--state)
+(defvar agent-shell-user-label)
 ;; Soft reference: `agent-shell-prompt-bar-mode' may be unbound when the
 ;; prompt bar is not loaded.  Read it with `bound-and-true-p'.
 (defvar agent-shell-prompt-bar-mode)
@@ -116,27 +119,6 @@ FACE carries the box (see `agent-shell-chat-me-label').
 For example, (agent-shell-chat--label \"Me\" \\='agent-shell-chat-me-label)
 returns \" Me \" in that face."
   (propertize (format " %s " text) 'face face))
-
-(defun agent-shell-chat--agent-name ()
-  "Return the attached agent's display name for the response label.
-
-For example, with a mode-line name of \"Claude\" returns \"Claude\";
-with none available, returns \"Agent\"."
-  (or (map-nested-elt agent-shell--state '(:agent-config :mode-line-name))
-      "Agent"))
-
-(defun agent-shell-chat--prompt-face-p (value)
-  "Return non-nil when a `font-lock-face' VALUE marks a shell prompt.
-A live prompt carries `comint-highlight-prompt'; a restored or echoed
-prompt carries `agent-shell-prompt' (which inherits it).  Either may be
-repeated.
-
-For example, \\='comint-highlight-prompt, \\='agent-shell-prompt, and
-\\='(comint-highlight-prompt comint-highlight-prompt) all return non-nil,
-while \\='default returns nil."
-  (let ((faces (if (listp value) value (list value))))
-    (or (memq 'comint-highlight-prompt faces)
-        (memq 'agent-shell-prompt faces))))
 
 (defun agent-shell-chat--extends-bg-p (face)
   "Return non-nil when FACE paints an `:extend' background past end of line.
@@ -212,7 +194,7 @@ whitespace separates them."
       (while (< pos (point-max))
         (let ((run-end (or (next-single-property-change pos 'font-lock-face)
                            (point-max))))
-          (when (agent-shell-chat--prompt-face-p
+          (when (agent-shell--prompt-face-p
                  (get-text-property pos 'font-lock-face))
             (push (cons pos run-end) runs))
           (setq pos run-end)))
@@ -308,7 +290,7 @@ above, putting the first line of a multi-line input out of reach of
                                       (skip-chars-forward " \t\n")
                                       (point))))
                (me-label (agent-shell-chat--label
-                          "Me" 'agent-shell-chat-me-label))
+                          agent-shell-user-label 'agent-shell-chat-me-label))
                ;; Face the padding and marker `default' so they do not inherit
                ;; the covered text's face: a display string's unfaced chars
                ;; take the face of the text they replace, and after a code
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by xenodium** [comment]
Thanks for the PR.

In the current approach, we run into hierarchical shortcomings. For example, if the agent generates any headings then Me/Agent turns get pushed down the markdown hierarchy. It also mixes fragment UI (▶▼) with markdown.

I had a go at it in https://github.com/xenodium/agent-shell/commit/707d402a036dd90ebca49611285f465b10850b5a. I've done some limited testing. Please take it for a spin.

You should see something like:

```markdown
# Me

question

# Claude

## Explored the repo

### Read file-1

contents 1

### Read file-2

contents 2
```

alberti42:
Thanks for reworking my PR. Your approach of handling the heading is the right one; so, I agree with your objections.

I took your commit for a spin, and I suspect a few problems:

- A copied code block or table can lose its Markdown formatting, depending on what else is highlighting the buffer.
- With chat mode off there are no turn labels, the prompt is still copied verbatim, and a question and its answer run together.
- A response containing a deep heading can end up with a heading that ties the turn heading above it.

I have candidate fixes for these. If you can already work out a few of these issues with a proper solution, I am happy to test your fixes. Otherwise, I will open a PR in the next days, maybe early next week — I still want to test everything properly first.

xenodium:
Thank you. I don't have fixes for these just yet.

> I have candidate fixes for these.

Yes please. Send a PR if keen.

Heads-up... While it would be nice to fix all of these, I have to make a judgement call on whether or not the fixes are worth it based on added complexity or amount of code needed. For example, if `With chat mode off there are no turn labels` requires a significant amount of code or complexity, I'm OK with not supporting this configuration. agent-shell ships with chat mode turned on, so it should work for all users out of the box. While I'm happy to try to accommodate for all sorts of configurations, us Emacs users are notoriously diverse in the number of things we want to configure, often in different ends of the spectrum. Trying to support everyone's favourite configuration adds another degree of complexity to my everyday maintenance, thus why I try to scale things down whenever possible. Hope that makes sense.

alberti42:
> Heads-up... While it would be nice to fix all of these, I have to make a judgement call on whether or not the fixes are worth it based on added complexity or amount of code needed.

Yes, we see it the same way. For this reason, I did not submit the first thing produced by vibe-coding. I need to check them individually and convince myself whether it makes sense in relation to the added complexity and severity of the bug/missing feature.

xenodium:
Thank you

Issue #786, opened 2026-08-20
chat mode: `<up>` skips the first line of a multi-line prompt draft


OSadovy:
## Checklist

- [x] *I agree to communicate with the author myself* (not AI-generated).
- [x] I've read the README's [Filing issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#filing-issues) section.
- [ ] I'm running the latest versions (fill in below).
  - agent-shell version: 0.73.4
  - acp.el version: 20260803.2034
  - ACP package (e.g. claude-agent-acp) version: claude-agent-acp 0.70.0
  - Agent CLI (e.g. claude, gemini) version: claude 2.1.234
- [ ] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [ ] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README).

Reproduced on 0.73.4; the code path below is unchanged in `main` (0.74.2).

## Steps to reproduce

1. Start an agent shell, send a prompt, wait for the response.
2. At the prompt, type:
   ```
   first line

   second line
   ```
3. `C-a` on `second line`, then `<up>` `<up>`.

**Expected:** the blank line, then `first line`.

**Actual:** the blank line, then the end of the agent's last response line. `first line` is not reachable with `<up>` at any goal column.

## Cause

`agent-shell-chat-mode.el:297`

```elisp
(start (if keep-term (1+ raw-start) raw-start))
```

There are two newlines between a response and the prompt. `keep-term` gives back one — the response's terminator — but the second one terminates the blank line, and stays covered. That blank line is then wholly inside the `Me` overlay's `display ""` run, so it has no screen presence.

`adjust_point_for_property` will not leave point inside such a run, and for an *empty* display string moving backward it goes to `beg - 1`: the character before the whole run, which is the end of the response line two lines up.

Same reasoning as the comment at line 283, one newline further along.

`agent-shell-chat-me-keeps-response-terminator-test` inserts the real `"\n\n"` layout but only asserts the *first* terminator is uncovered, so it passes either way.

## Possible fix direction

Starting the `Me` overlay at the prompt's `line-beginning-position` would keep the blank line a real line.

That alone would let surplus blank lines through — across 8 shells here the overlay covers 1, 3 and 6 newlines — so collapsing the surplus may want a separate run above the kept blank line rather than extending this one.

[...]

PR #804, opened 2026-08-25
Stop the live prompt flickering on every keystroke

alberti42:
## The problem

Typing into the shell flickers. Each character paints the input flush left for a frame, then jumps right under the ` ❯ ` marker — so a whole sentence visibly shifts sideways as it is typed. This causes a conspicuous flickering with each keystroke typed by the user.

## Cause

The marker is drawn as a `line-prefix` on the overlay covering the live comint prompt:

```elisp
(list (cons 'before-string (if label-nl "" before))
      (cons 'display "")
      (cons 'line-prefix (or marker ""))
      (cons 'wrap-prefix (or marker "")))
```

A `line-prefix` belongs to the whole line, and the live prompt shares its line with the input being typed. Emacs therefore cannot use its cheap single-line redisplay path: every `self-insert` re-lays the line out, and the unindented state is visible in between.

Measured — with counters on `agent-shell-chat--relabel`, `--label-prompts`, `--decorate-prompt-region` and friends, 25 keystrokes produced **zero** calls to any of them. No chat-mode code runs while typing, so this is redisplay, not a deferred relabel. Clearing `line-prefix` alone (leaving `wrap-prefix` in place) stops the flicker.

## Fix

Draw the marker as the covered prompt's `display`, which was already replacing that text with `""`:

```elisp
(cons 'display (or marker ""))
(cons 'line-prefix "")
```

`display` is a property of the covered text rather than of the line, so it costs nothing per keystroke. `wrap-prefix` is untouched — it applies only to wrapped continuation rows and is not implicated.

## This does not reintroduce #786

The comment at that call site explains why `line-prefix` was chosen: a string shown at the prompt position keeps `previous-line` from settling on the input's first line. That is true of an **inserted** string — a `before-string` holds point and cursor on the row above, which is #786.

A `display` string is different in kind: it stands on the positions of the text it covers, which are the prompt's own. Vertical motion therefore behaves exactly as it did when the prompt text was visible.

Verified by hand on a multi-line draft: from the last line, the up arrow walks onto the blank line, then onto the draft's **first** line, and only then up into the transcript above.

## Tests

The full suite passes — 384 tests across 12 files, 0 unexpected.

A new test, `agent-shell-chat-live-prompt-marker-not-line-prefix-test`, pins the mechanism; it fails against the previous implementation.

Two existing tests changed, both asserting the old mechanism rather than behaviour:

- `agent-shell-chat-mode-tests--marker-string` read the marker off `line-prefix`; it now reads `display`.
- `agent-shell-chat-label-is-before-string-test` asserted an empty `display` on every `Me` overlay, as a proxy for "the label is not drawn as a `display`".  It now asserts that directly, since the live prompt's `display` legitimately carries the marker. The label itself must still never be a `display` — that is what keeps the cursor off it — and the test still checks so.

`agent-shell-chat-nothing-shown-at-prompt-test` passes unchanged; only its docstring is updated, since it described the marker as travelling via `line-prefix`.

### DIFF
--- a/agent-shell-chat-mode.el
+++ b/agent-shell-chat-mode.el
@@ -365,8 +365,8 @@ above, putting the first line of a multi-line input out of reach of
                               (1- pos)))
                ;; The live prompt's marker, shown before the input whether or
                ;; not text has been typed yet.  Keying this off `blank' would
-               ;; drop it the instant the user starts typing.  Carried as a
-               ;; `line-prefix', which occupies no buffer position.
+               ;; drop it the instant the user starts typing.  Carried as the
+               ;; covered prompt's `display', standing on its buffer positions.
                (marker (when (and live labeled)
                          (propertize (concat agent-shell-chat--body-indent
                                              agent-shell-chat--prompt)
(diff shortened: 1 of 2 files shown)


xenodium:
Thank you for the PR. I can't reproduce the issue on HEAD. Are you able to reproduce on the latest?

2026-08-29-09:29:13-Emacs

alberti42:
Still reproduces for me on b7731a3. I've rebased the PR onto that commit: in HEAD, the marker is still drawn as a line-prefix `((cons 'line-prefix (if label-nl (or marker input-indent) "")))`, so nothing that landed since the original PR would have changed it.

One difference since I first reported: I now only see it when deleting characters, not when inserting. Insertions and deletions take different redisplay paths, so which one shows it varies. But the fundamental flickering problem is still there and the mechanism is the same.

A video showing the flickering when backspacing is attached.

https://github.com/user-attachments/assets/70f5fb25-e4b8-4df2-818f-5f9bb1fdcc1b

A guess at why you can't reproduce: I'm on Emacs 32.0.50 (master) with the NS port on macOS. I did not check whether line-prefix disabling redisplay's single-line optimization differs between versions.

Independent of reproduction, the PR now includes a test that asserts the mechanism directly and fails without the fix, so it's verifiable either way.

On #786: I understand why line-prefix was used there. The problem was the before-string at the prompt position. A before-string adds text that is not really in the buffer, so the cursor cannot be placed on it, and the up arrow skipped the first line of the draft.

This PR does not bring that problem back, because it does not add any text. It puts the marker in the display property that already covers the prompt. A display property replaces text that is really there, so those positions still exist and the cursor can still be placed on them. Your own comment in the code this replaced said the same thing: "a `display' string is backed by the covered buffer positions, so vertical motion (`next-line'/`previous-line') lands the cursor on the label".

I checked by hand: with a multi-line draft, the up arrow still moves onto its first line, and then up into the transcript.

xenodium:
Thank you for looking into it and rebasing. Merged.

PR #809, opened 2026-08-31
Handle ACP session/list pagination

KarimAziev:
This PR fixes the issue for some providers, like Codex and Qwen, where not all sessions are listed because pagination was not handled. Other providers, like Claude, currently return all sessions on the first page, so their behavior remains unchanged.

I also added an optional page limit for session listing requests. This limits the number of pages/requests followed during pagination, not the number of sessions, and it defaults to `nil`, meaning all pages are fetched. The option therefore does not impose a default cap, a finite limit is explicitly opt-in and may silently truncate the result.

I'm aware of the "Avoid premature customization" guidance in CONTRIBUTING, but it still seems better to have this option because without it agent-shell may follow a provider-determined number of pages, while ACP does not expose a portable page-size/session-count parameter. However, if you think this customization is redundant, no problem-just say the word and I will remove it.

## Checklist

Only after opening the PR did I realize that I should have created an issue first. Sorry about that. I'm happy to treat this PR as the discussion and adjust or reduce its scope based on your feedback.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -7371,6 +7371,58 @@ SESSION-TITLE is an optional display title for the resumed session."
    :on-failure (agent-shell--make-error-handler
                 :state (agent-shell--state) :shell-buffer shell-buffer)))
 
+(cl-defun agent-shell--make-session-list-request (&key cwd cursor)
+  "Return an ACP session/list request for CWD and optional CURSOR.
+
+CURSOR is an opaque token returned as `nextCursor' by a previous
+session/list response.
+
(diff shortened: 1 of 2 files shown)


xenodium:
Thank you for the PR. Merged.

Couple of follow-ups in https://github.com/xenodium/agent-shell/commit/bda0601429c5492000d2c8bcc77a12dd31e80d72 and https://github.com/xenodium/agent-shell/commit/cc24345e9db2a205f7b8821015f89be3de7dd1e4.

Please give them a try and let me know if you run into issues.

PR #438, opened 2026-03-20
Fix for structured input from toolCall.rawInput.plan

timfel:
I am getting this JSON sometimes with codex-acp:

```
{"jsonrpc":"2.0","method":"session/update","params":{"sessionId":"019d02d5-885b-7b40-9c8f-a75845dd96f2","update":{"sessionUpdate":"tool_call","toolCallId":"call_7ADCCGOkY7FL8ram6C1Wbabn","title":"update_plan","status":"completed","rawInput":{"explanation":"Entered planning mode. Starting with task definition before any implementation work.","plan":[{"step":"Clarify the target task and constraints","status":"in_progress"},{"step":"Inspect the relevant code and tests","status":"pending"},{"step":"Draft an implementation and verification plan","status":"pending"}]}}}}
```

or more structured from your traffic log

```
:direction incoming
:kind      notification
:object
           jsonrpc 2.0
           method  session/update
           params
                   sessionId 019d02d5-885b-7b40-9c8f-a75845dd96f2
                   update
                             sessionUpdate tool_call
                             toolCallId    call_7ADCCGOkY7FL8ram6C1Wbabn
                             title         update_plan
                             status        completed
                             rawInput
                                           explanation Entered planning mode. Starting with task definition before any implementation work.
                                           plan
                                                       step   Clarify the target task and constraints
                                                       status in_progress

                                                       step   Inspect the relevant code and tests
                                                       status pending

                                                       step   Draft an implementation and verification plan
                                                       status pending
```

This ends up causing this error:

```
Debugger entered--Lisp error: (wrong-type-argument stringp [((step . "Clarify the target task and constraints") (status . "in_progress")) ((step . "Inspect the relevant code and tests") (status . "pending")) ((step . "Draft an implementation and verification plan") (status . "pending"))])
  string-empty-p([((step . "Clarify the target task and constraints") (status . "in_progress")) ((step . "Inspect the relevant code and tests") (status . "pending")) ((step . "Draft an implementation and verification plan") (status . "pending"))])
  (not (string-empty-p str))
  (and str (not (string-empty-p str)) str)
  agent-shell-ui--string-or-nil([((step . "Clarify the target task and constraints") (status . "in_progress")) ((step . "Inspect the relevant code and tests") (status . "pending")) ((step . "Draft an implementation and verification plan") (status . "pending"))])
```

AFAICT the issue is that agent-shell here assumes that rawInput is a plain string, the spec says it can be an object
[...]

PR #812, opened 2026-09-02
Display Pi terminal output in agent-shell.

timfel:
I am using pi-acp, and it sends terminal output in a custom meta field. This is not against the ACP spec, just a bit unfortunate because now the client needs to do something with it. I am also preparing a PR to pi-acp. However, the maintainer of pi-acp is not super active (judging by the open PRs for other ACP features), so I would like to fix this on the agent-shell side for now if that's ok. The notification-adapter seems like the natural hook for this.

## Checklist

### DIFF
--- a/tests/agent-shell-pi-tests.el
+++ b/tests/agent-shell-pi-tests.el
@@ -1,0 +1,91 @@
+;;; agent-shell-pi-tests.el --- Tests for agent-shell-pi -*- lexical-binding: t; -*-
+
+(require 'ert)
+(require 'agent-shell-pi)
+(require 'map)
+
+;;; Code:
+
+(ert-deftest agent-shell-pi--notification-adapter-terminal-output-test ()
+  "Test `agent-shell-pi--notification-adapter' accumulates terminal output.
+Pi sends terminal output as deltas in `_meta.terminal_output'.  Its current
+final update also includes `_meta.terminal_exit'; a status-only final update
+is accepted as well so it cannot replace the accumulated output with empty
+content."
+  (with-temp-buffer
+    (setq-local agent-shell-pi--terminal-output-snapshots nil)
+    (let ((notification
+           '((method . "session/update")
+             (params
+              (update
+               (sessionUpdate . "tool_call_update")
+               (toolCallId . "call-1")
+               (status . "in_progress")
+               (_meta
+                (terminal_output
+                 (terminal_id . "call-1")
+                 (data . "hello\n"))))))))
+      (agent-shell-pi--notification-adapter :acp-notification notification)
+      (should (equal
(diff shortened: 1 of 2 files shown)


xenodium:
Thanks for the PR.

Mind adding a test and also a pass against CONTRIBUTING.org?

xenodium:
ps. Would be good to document even if in the PR what the schema that we're working against looks like.

A sequence of simple messages would be great. In helping anyone maintaining understand thins like what does pi send after terminal_exit? If it emits a separate tool_call_update carrying the final status without any _meta, the adapter passes it through with empty content, and since we re-render the tool call fragment body from the update's content each time, the terminal output would get blanked out at the very end (and logged as an empty transcript entry). Would that be the case here?

timfel:
I've expanded the docs and had some tests generated which I reviewed. I also asked the agent to review the docs change I did, and the code now makes sure that if a final empty update came, we wouldn't blank out the content (it isn't currently happening in my testing, but good for resilience)

tests/agent-shell-pi-tests.el
```diff
@@ -0,0 +1,91 @@
...
+                 (signal . nil))))))))
+      (agent-shell-pi--notification-adapter :acp-notification notification)
+      (should (equal
+               (map-nested-elt notification '(params update content))
+               '(((type . "content")
+                  (content (type . "text")
+                           (text . "```\nhello\nworld\n```"))))))))
+    (should-not agent-shell-pi--terminal-output-snapshots))
+
+  (with-temp-buffer
```
xenodium, line 56: Guessing you meant to wrap in ert-deftest? Or maybe as part of previous test?

xenodium (changes requested):

xenodium:
Thank you for the changes. Just a minor fix in the test. Mind taking a look?

tests/agent-shell-pi-tests.el
timfel: Oops 🤦

xenodium:
Thanks for addressing the comment. Merged.

> I am also preparing a PR to pi-acp. However, the maintainer of pi-acp is not super active (judging by the open PRs for other ACP features)

Prolly cause it takes time to maintain these projects (reading PRs, answering questions, fixing bugs, adding features, etc)? ;) How's [this](https://github.com/xenodium/agent-shell/pull/438#issuecomment-4129945374) coming along?

timfel:
> How's [this](https://github.com/xenodium/agent-shell/pull/438#issuecomment-4129945374) coming along?

Liberapay has pulled the money out of my account for the last 3 months, so I do hope it's arriving ;)

xenodium:
Thank you sir! 🫡

PR #813, opened 2026-09-03
Add prefix-arg page jumps to viewport paging

liaowang11:
> **Draft note:** description below is a factual first pass to be rewritten in
> my own words before marking this ready for review.

Feature request: #814

## What this does

`C-u N` on `agent-shell-viewport-next-page` / `agent-shell-viewport-previous-page`
(`f` / `b` in the viewport) moves N interactions instead of one.

Paging with no prefix argument takes the same code path as before. The
boundary checks and the compose-snapshot restore for an already-newest page
both still run before the first step.

## Behavior at the ends of history

- Moving forward past the newest interaction restores a parked compose
  snapshot, so `C-u N f` ends where pressing `f` N times would end rather
  than stopping one page short of the draft.
- With no snapshot to restore, an N larger than the remaining history stops
  on the last interaction reached instead of erroring. Pressing `f` that many
  times would have raised `No next page` on the last press; erroring after
  having already moved seemed less useful than stopping.
- Moving backward is unchanged apart from the count: from the first page it
  still raises `No previous page`.

## Implementation

- `agent-shell-viewport--move-pages` loops the existing
  `shell-maker-next-command-and-response` step, returning the last
  interaction reached and whether history ran out.
- `agent-shell-viewport--restore-compose-snapshot` extracts the snapshot
  restore so the pre-step and post-step paths share it.

## Testing

Six tests added under `tests/agent-shell-tests.el`, covering the N jump in
both directions, the prefix argument arriving via `call-interactively`, the
no-prefix single step, and each of the two end-of-history outcomes.

Full suite on this branch: 666 tests, 0 unexpected, 1 skipped
(`agent-shell-restart-preserves-default-directory`, which skips itself).
On `main` at the same commit base: 660 tests, 0 unexpected, 1 skipped.

`checkdoc` reports nothing on `agent-shell-viewport.el`, and
`byte-compile-file` produces the same two warnings as `main`
(`replace-buffer-contents` obsolete in `agent-shell.el`, and
`agent-shell-backward-up-item` not known to be defined), both pre-existing.

## Checklist

### DIFF
--- a/agent-shell-viewport.el
+++ b/agent-shell-viewport.el
@@ -946,22 +946,54 @@ QUOTED-TEXT is inserted as a block quote as part of the reply."
   (insert "continue")
   (agent-shell-viewport-compose-send))
 
-(defun agent-shell-viewport-previous-page ()
-  "Show previous interaction (request / response)."
-  (declare (modes agent-shell-viewport-view-mode))
-  (interactive)
-  (agent-shell-viewport-next-page :backwards t :start-at-top t))
+(defun agent-shell-viewport--restore-compose-snapshot ()
+  "Open the compose page on the parked draft, consuming the snapshot."
+  (let ((snapshot agent-shell-viewport--compose-snapshot))
+    (agent-shell-viewport-edit-mode)
+    (agent-shell-viewport--initialize)
+    (insert (map-elt snapshot :content))
+    (goto-char (map-elt snapshot :location))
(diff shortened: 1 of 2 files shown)


xenodium:
Thank you for the improvement!

xenodium:
FYI https://github.com/xenodium/agent-shell/commit/3bd32a16f4eededd54e190d94318f536ed4fac7c

PR #816, opened 2026-09-04
Support inline form elicitation

daym:
## Checklist

- [ ] I've filed a feature request/discussion for a new feature. Existing discussion https://github.com/xenodium/agent-shell/issues/792
- [X] I'm making visual changes, so I'm including screenshots so you can view and discuss.

Requires https://github.com/xenodium/acp.el/pull/28 mainly because we need an outgoing request’s assigned JSON-RPC ID before transmission, so a request-scoped elicitation can be tied to--and cancelled with--its actual owning request.

There's also an alternative implementation of agent-shell elicitation in #797 but:

- ACP protocol explicitly says [the agent may wait, or continue working, or chain elicitations](https://agentclientprotocol.com/rfds/elicitation#can-agents-use-elicitation-for-information-required-before-responding)--the standardized form is a separate "non-blocking" request. But #797 would make it a turn modal. This PR does not.

### DIFF
--- a/README.org
+++ b/README.org
@@ -923,6 +923,16 @@ All of the above settings can be applied on a per-project basis using [[https://
 - =C-c C-c= - Interrupt current agent operation
 - =TAB and Shift-TAB= - Navigate interactive elements
 
+  When an ACP agent requests structured input, agent-shell displays the
+  authoritative shell and renders the form inline there.  =TAB= and =Shift-TAB=
+  include its fields, options, and actions in normal interactive-element
+  navigation.  Press =RET= to edit a value, toggle an option, or activate
+  =Submit=, =Decline=, or =Cancel=.
+
+  Form elicitation must not be used for passwords, API keys, access tokens,
+  payment credentials, or other secrets.  agent-shell currently supports
+  form elicitation only; it does not advertise URL-mode elicitation.
+
   To customize =RET= binding behaviour, you can use something like:
 
   #+begin_src emacs-lisp :lexical no
--- a/agent-shell-chat-mode.el
+++ b/agent-shell-chat-mode.el
@@ -161,9 +161,9 @@ so a label following it needs a full pad rather than a single newline."
     (or (= beg (point-min))
         (eq (char-before beg) ?\n))))
 
-(cl-defun agent-shell-chat--ensure-overlay (&key tag beg end props rear-advance
-                                                 (anchor-beg beg)
-                                                 (anchor-end end))
+(cl-defun agent-shell-chat--ensure-overlay
+    (&key tag beg end props front-advance rear-advance
+          (anchor-beg beg) (anchor-end end))
   "Ensure a TAG overlay spans BEG..END carrying PROPS.
 
 TAG is a symbol naming what the overlay is for (`me', `agent' and so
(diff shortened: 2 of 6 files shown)


daym:
image

daym:
Existing discussion https://github.com/xenodium/agent-shell/issues/792

xenodium:
Thank you for the PR. Not sure if you've seen the existing #797? While I'm appreciative of the extents that #816 goes (ie. extending acp.el to accommodate for scopling/lifetime), #797 has a much smaller scope (less code for me to maintain). I'm keen to land #797 and iterate on that as needed to improve the feature. I understand this is unlikely the outcome you expected. Sorry.

edit: typos

daym:
I'm fine with maintaining a fork. 

I cannot use # 797 at all. I have seen #797 , it only supports question dialogs *in turns* and it's modal, in clear violation of the protocol. 

I'm using elicitations to ask for confirmation when resuming a chat, so there are no turns yet at that time. Furthermore, my harnesses are containerized, so they can't just open a wayland dialog on their own, for example.

xenodium:
> I'm using elicitations to ask for confirmation when resuming a chat,

Interesting. Can you share more about your workflow? Examples?

daym:
> > I'm using elicitations to ask for confirmation when resuming a chat,
> 
> Interesting. Can you share more about your workflow? Examples?

I write a general coding agent harness https://github.com/friendly-machines-com/loki  in python using only the python stdlib (0 other Python dependencies). It supports (modern only) terminal emulators and can run as a TUI program, console program (headless), and ACP. When using loki-acp, it works fine, but eventually the user would resume an existing chat.

If you want to see the non-acp version first (easier to set up):

1. Run `./loki.py` (no need to install it)
2. Type `/model` (if you didnt set `LOKI_`* env vars beforehand) and select a model to use in the dialog (using a number)
3. Chat for a bit
4. Type `/quit` (loki exits, saves chat log under `$PWD/.loki`)
5. Run `./loki.py resume`
6. Select `1` in the chat log selection dialog (that selects that chat log)
7. *It asks you to confirm that the security-sensitive settings in the chat log are the ones you had when you saved it* (those are for example the URL)
8. Only if yes, it resumes.

An attack scenario is that otherwise an attacker could edit chat logs (which are inside random project directories) and edit them to swap out the chat endpoint (the llm model endpoint) for theirs and wait for the credentials to arrive the next time the user does loki resume.

If you actually try to do this, the easiest way is to set up some env vars like `LOKI_API_KEY=xyz` and `LOKI_API_BASE=someurl` (possibly localhost) and `LOKI_MODEL=abc`. But the regular operation is it gets the token env var names to use from https://models.dev and expects the user to have set them instead (for example `OPENCODE_API_KEY`, `ZHIPU_API_KEY`). The list of models on `/model` gets filtered to show only those where the necessary key env vars are set. I had previous versions which didn't filter--but that would print a list of about 2000 models each time :)

That means the user's wrapper scripts that set those env vars up and start loki should at least use `exec`. loki itself makes sure to overwrite the env var slots in its Linux kernel process record after having taken them. Nevertheless, for avoiding other model shenanigans and timing-/proc-attacks,  I recommend using `smolvm` or at least a Linux container.

There is a separate `loki.py auth` to set up oauth login with openai subscription, and that one is even more hairy if stolen. The latter is why I am overly paranoid.

I actively run this in emacs using agent-shell patched with this #816 , and with a org-latex preview so I get answers in beautifully typeset mathematics :)

image

You can infer this was resumed (`saved connection` at the top) and that's one of the things where the missing elicitation tripped me up :)

There are other use cases for this elicitation without being inside turns--for example it can happen that the access token's lifetime is over and needs to be refreshed and the refresh fails in a way where the refresh was sent from the user's machine but maybe (or not) arrived at the model provider. Now what? Did it arrive or not? This is a distributed systems problem and needs the user to be annoyed and then active -> needs elicitation. 

Also, openai and z.ai both have "banked reset [cards]". From time to time they hand those out to users. The user can then choose when (or if) they want to redeem the reset and then they get the weekly or monthly or whatever limit reset. So I have a command `/account resets` for that, and again that needs a form of what reset cards are available, some way to choose the reset card to redeem (if any), and probably a confirmation that that's really what you wanted to select. Those again need elicitation.

Also, Loki supports a lot of different providers and  uses a json gotten from https://models.dev as the selection list, including where credentials are sent. That needs pinning of the (host, credential name) set, and elicitation if it wasn't seen before or whether it was changed. Often, that is in the very beginning before any turns have started--but it also can be in the middle of a conversation.

PR #821, opened 2026-09-11
Check dependencies against the Package-Requires floors

alberti42:
`agent-shell--start` checks its dependencies at runtime, but against floors that have drifted from the ones the header declares: it compares shell-maker against 0.91.2 and acp against 0.14.3, while `Package-Requires` says 0.97.3 and 0.15.1. This results in the obvious bug: A dependency whose the header label as unsupported would start a shell anyway, and the feature it lacks surfaces later as an unrelated failure/bug.

Package managers that resolve versions enforce the header on install, so this assert is belt-and-braces there. Package managers that resolve only dependency names — straight.el, for one — never look at versions at all, which leaves this check as the only gate, and a stale gate lets precisely the combination the header forbids through.

This names both floors in defconsts next to the declared versions and reports them in the error message, so bumping the header moves the check with it.

Tests: 680 ert tests, 0 unexpected, same 4 skips as the baseline (Emacs 29.1 path, deps cloned at HEAD as CI does).

**Side story: How this affected me specifically**

Concretely, with agent-shell 0.75.3 and shell-maker 0.97.2 every session restore dies silently: `shell-maker--should-auto-scroll-p` signals `args-out-of-range` when asked under the narrowing `agent-shell--update-fragment` installs for `:above-last-prompt`, so `agent-shell--finalize-session-init` never runs and the replayed history stays buffered in the state's `:pending-restore`. acp.el contains response-handler errors and only logs them when `acp-logging` is on, so nothing is reported at all — the shell just sits at a bare prompt with "Starting agent" never reaching "Ready". shell-maker fixed the signal in 0.97.3 (fe38358), which is exactly why the header requires it; the runtime assert compared against 0.91.2 and waved it through.

### DIFF
--- a/agent-shell.el
+++ b/agent-shell.el
@@ -9,6 +9,15 @@
 
 (defconst agent-shell--version "0.75.3")
 
+;; Minimum dependency versions, as declared in the `Package-Requires'
+;; header above.  Package managers that resolve versions enforce the
+;; header on install; those that only resolve dependency names (straight.el,
+;; for one) leave `agent-shell--start' as the sole check, so keep these two
+;; in sync with it.
+(defconst agent-shell--shell-maker-minimum-version "0.97.3")
+
+(defconst agent-shell--acp-minimum-version "0.15.1")
+
 ;; This package is free software; you can redistribute it and/or modify
 ;; it under the terms of the GNU General Public License as published by
 ;; the Free Software Foundation; either version 3, or (at your option)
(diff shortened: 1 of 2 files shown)


alberti42:
Pushed a second commit adding an ert test that pairs the two floors with the header: it reads `Package-Requires` out of agent-shell.el with `lm-header-multiline` and compares each declared version against its defconst, so a header bump that leaves the runtime check behind fails the suite rather than shipping. Verified it fails on drift (setting the shell-maker floor back to 0.91.2 makes it red). Suite is 681 tests, 0 unexpected, same 4 skips as the baseline.

xenodium (approved):
Thanks. Had been meaning to drop the runtime checks. We can give this a try.

xenodium:
FYI https://github.com/xenodium/agent-shell/commit/7377ba827c315fe1aff214f9ba8096068b03a68b

PR #824, opened 2026-09-13
Quote file mentions whose path holds whitespace

dustinfarris:
`agent-shell--parse-file-mentions` accepts two forms, `@path` and `@"a quoted path"`, but `agent-shell--get-files-context` only writes the bare one. A bare mention ends at the first space, so sending a file like `My Design.png` from dired reached the agent as the mention `@My` followed by loose text, and the file was never embedded. Images showed their preview in the prompt, which made it look attached.

This adds `agent-shell--file-mention`, which writes the quoted form when the path contains whitespace, and uses it for the three places `agent-shell--get-files-context` builds a label: the image preview, the project-relative path, and the absolute path.

## Before and after

| Sent file | Before | After |
|---|---|---|
| `src/main.el` | `@src/main.el` | `@src/main.el` |
| `My Design.png` | `@My Design.png` (parsed as `@My`) | `@"My Design.png"` |

Paths without whitespace are unchanged, so existing prompts and tests keep their exact text.

## Why here and not in the parser

The parser already handles the quoted form; it is the writer that never used it. Quoting at the write site keeps a single mention grammar and means anything else that emits mentions (`agent-shell-send-file`, drag and drop) gets the fix by going through `agent-shell--get-files-context`.

## Checklist


xenodium:
Thank you for the fix

xenodium:
FYI https://github.com/xenodium/agent-shell/commit/a46412f2de44ad5d54f0f7be3ba0b76e007945e8

PR #825, opened 2026-09-13
attach files dropped on a shell or viewport buffer

### LINKED ISSUE #784: [Feature] Drag and drop an image into the agent shell buffer
## Checklist

  - agent-shell version: 0.73.3
  - acp.el version: 0.13.1
  - ACP package (e.g. claude-agent-acp) version: claude-agent-acp 0.66.0
  - Agent CLI (e.g. claude, gemini) version: claude 2.1.233
- [ ] For issues, I'm including [ACP traffic](https://github.com/xenodium/agent-shell?tab=readme-ov-file#how-do-i-viewget-agent-client-protocol-traffic) (as per README).

## Request

Dragging an image file onto an `agent-shell` buffer should attach it to the prompt, the way `agent-shell-send-clipboard-image` and `agent-shell-yank-dwim` already do for the clipboard.

## Current behaviour

`agent-shell-mode` doesn't handle drops, so a drop falls through to Emacs' default `dnd-protocol-alist` handling and the image is opened in its own buffer, taking over the window. Nothing reaches the prompt.

## Suggested behaviour

On a `file://` drop in an `agent-shell` buffer, insert the file as prompt context — the same thing `agent-shell-yank-dwim` already does once it has a path:

```elisp
(agent-shell-insert
 :text (agent-shell--get-files-context :files (list path))
 :shell-buffer (agent-shell--shell-buffer))
```

A buffer-local `dnd-protocol-alist` entry set up in `agent-shell-mode` would cover it.

Two things worth deciding:

- Non-image files: attaching a dropped `.log`/`.csv`/source file as context seems just as useful, but falling back to `dnd-open-file` for anything non-image is a defensible scope too.
- Drops that carry image *data* rather than a file URL (dragging an image straight out of a browser) could reuse the existing save-to-`.agent-shell/screenshots` path from `agent-shell--save-clipboard-image` before inserting.

## Why

Clipboard paste covers freshly-taken screenshots well. Drag-and-drop covers the other half: images that already exist as files — a design mockup, a diagram, a screenshot sitting in Finder or Dired — where today the route is copying the path out by hand.

dustinfarris:
Closes #784.

A file dropped on an agent-shell buffer fell through to Emacs's default drop handling and opened in the window. This installs buffer-local `dnd-protocol-alist` entries in the shell buffer and both viewport buffers that attach the dropped files to the prompt instead, the same way `agent-shell-yank-dwim` attaches a clipboard image. Images get their preview; any other file is attached as an `@path` link, relative to the shell's cwd as `agent-shell-send-region` does.

https://github.com/user-attachments/assets/6e808fdb-3c7e-4b0b-b7a7-c45b45d6bec4

## Behavior

- **Local files only.** Only the three `file:` forms Emacs itself routes to `dnd-open-local-file` are taken over. A `file://HOST/path` URL still reaches `dnd-open-file`.
- **Transient drops are copied.** Dragging a screenshot thumbnail or an image out of a browser drops a file promise: macOS writes it under the temp directory for the duration of the drag and deletes it when the drag ends, so by submit time the path pointed at nothing. Such files are copied into `.agent-shell/screenshots` first, named like the sibling screenshot writers name theirs. Files dropped from anywhere else are attached in place, including files of a project that itself lives under the temp directory (`agent-shell-new-temp-shell`).
- **A drop is one insertion.** The handler is a `dnd-multiple-handler` on Emacs 30+, so several files dropped together become one insertion and one busy check. Every file is checked before any is copied; a drop that includes an unreadable file or a directory attaches nothing rather than part of the selection. Emacs 29 still calls the handler once per URL, which `ensure-list` absorbs.
- **Busy shells queue.** A drop on a shell mid-turn is queued the way a sent region is, rather than erroring after the copy was made.
- **Opt out.** When `agent-shell-dnd-enabled` is set to `nil`, Emacs's default drop handling is kept — matching `agent-shell-file-completion-enabled`.

## Related

Independent of #824, but they belong together: the default macOS screenshot name (`Screenshot 2026-09-13 at 8.15.12 AM.png`) contains spaces, and without that fix an `@` mention of a dropped screenshot ends at the first one, so the preview renders and the file is never sent.

## Checklist


xenodium (approved):
Thank you for the PR!

xenodium:
FYI https://github.com/xenodium/agent-shell/commit/b85352e0fc887857f99d1fe8d82d56d5f49d2211

xenodium:
ps. I had no idea you could draw the temporary thumbnail from the macOS screenshot util. Nice one! I always right clicked on it to copy it to clipboard.

dustinfarris:
> ps. I had no idea you could draw the temporary thumbnail from the macOS screenshot util. Nice one! I always right clicked on it to copy it to clipboard.

ah that would explain `agent-shell-send-clipboard-image` 😛 

to be fair, I discovered `agent-shell-send-screenshot` _after_ recording the demo — realizing my video probably wasn't the best flex

Issue #148, opened 2025-12-04
[feature-request] Folding visibility defaults


landakram:
I would find it useful to have "Thought process" sections unfolded by default and tool use sections folded by default, and to be able to configure these.

There are 3 things that would be useful to customize:

- Thought process folded/unfolded by default
- Tool use folded/unfolded by default
- Filter for tool use fold state. It's rarely helpful for me to search tool call results so I wouldn't unfold those by default, but I would definitely unfold edit tool calls by default. 

---

## Checklist

- [x] I've read the README's [Issues](https://github.com/xenodium/agent-shell?tab=readme-ov-file#issues) section.
- [x] For requesting new agent support, I'm including a link to the ACP-capable agent or related ACP package.
- [x] For issues, I'm including ACP traffic (as per README).

xenodium:
Thanks for filing.

> Thought process folded/unfolded by default
> Tool use folded/unfolded by default

These two could be handled by offering configurable defaults. Will take a look.

> It's rarely helpful for me to search tool call results so I wouldn't unfold those by default, but I would definitely unfold edit tool calls by default.

I'm thinking I should look into how org mode handles this sort of thing. When a node is collapsed, it should remain searchable and automatically expanded for matches. This could be a separate feature request (automatically expand nodes when searching).

ps. While this feature request and https://github.com/xenodium/agent-shell/issues/147 would be great additions, please consider [sponsoring](https://github.com/sponsors/xenodium) this project to make the work sustainable. Bonus points for getting [your employer](https://github.com/openai) to sponsor, who are as relevant as it gets ;) I'm now an indie dev, but when I worked in big tech, I was able to get my employer to send coin to the Emacs projects I used.

xenodium:
>  When a node is collapsed, it should remain searchable and automatically expanded for matches. This could be a separate feature request (automatically expand nodes when searching).

This part is now handled by https://github.com/xenodium/agent-shell/commit/478ce5ca6cbcb6ed0d8708b8422c5cd604ab22b4

xenodium:
https://github.com/xenodium/agent-shell/commit/e24266705b7ff1d6d57ec03a4700b764635523d3 introduces `agent-shell-thought-process-expand-by-default` and `agent-shell-tool-use-expand-by-default` options.

xenodium:
Hi @landakram. Happy New Year. I've made some changes since. Anything else needed here? Ok to close this issue?

xenodium:
[...]

PR #832, opened 2026-09-18
Respect search-invisible (whether isearch enters collapsed sections)

mrcnski:
Using agent-shell has a lot of advantages over something like claude CLI in ghostel, because of agent-shell's tighter integration with other Emacs features. Unfortunately, searching doesn't work very well because you get hundreds of spurious results from intermediate model output in collapsed sections. 

This PR makes that behavior configurable by having agent-shell respect the existing `search-invisible` option. See issue #148 which made matching folded content intentional, but forgot to support that option.

You can now, for example, have something like:

```elisp
  (add-hook 'agent-shell-mode-hook (lambda () (setq-local search-invisible nil)))
```

To flip it for a single search, you can type `M-s i` (`isearch-toggle-invisible`) while searching.

## Checklist


xenodium (approved):
Thanks for the improvement

xenodium:
FYI https://github.com/xenodium/agent-shell/commit/82d965fd1b353f52da9978b6c5e7b95c967cb5ed

mrcnski:
Thanks, good follow-up commit!

